PR #8, opened 2019-10-31
Mouse Button Bindings

InfinityGhost:
This adds bindings to the mouse buttons for all platforms.
## Currently coded for:
- [x] Windows
- [x] Linux
- [x] Mac OS X

## Currently confirmed functioning for:
- [ ] Windows
- [x] Linux
- [ ] Mac OS X


InfinityGhost:
Not going to be testing for Windows functionality since it's already broken anyway, though this code should work in a test without devices.

PR #53, opened 2019-12-10
Add Area Rotation

InfinityGhost:
# Changes
- [x] Centered area
- [x] Area rotation
  - [x] Update AbsoluteMode to support it
  - [x] Fix all AreaSetter methods
  - [x] Verify rotation angle accuracy
- [x] Clean up old variables

### DIFF
--- a/TabletDriverLib/Component/Area.cs
+++ b/TabletDriverLib/Component/Area.cs
@@ -23,19 +23,16 @@ namespace TabletDriverLib.Component
         public Point Position { set; get; } = new Point();
         public float Rotation { set; get; } = 0;
 
-        public IList<float> RotationMatrix
+        public float[] GetRotationMatrix()
         {
-            get 
+            var angle = Rotation * (Math.PI / 180);
+            return new float[4]
             {
-                float angle = Rotation * (float)(Math.PI / 180f);
-                return new float[]
-                {
-                    (float)Math.Cos(angle),
-                    (float)-Math.Sin(angle),
-                    (float)Math.Sin(angle),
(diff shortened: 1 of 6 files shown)


TabletDriverLib/Output/AbsoluteMode.cs
```diff
@@ -50,40 +23,60 @@ public override TabletProperties TabletProperties
...
+            pos.Y -= DisplayArea.Position.Y - (DisplayArea.Height / 2);
+
+            // Rotation
+            if (TabletArea.Rotation != 0f)
+            {
+                var tempCopy = new Point(pos.X, pos.Y);
+                var rotateMatrix = TabletArea.GetRotationMatrix();
+                pos.X = (tempCopy.X * rotateMatrix[0]) + (tempCopy.Y * rotateMatrix[1]);
+                pos.Y = (tempCopy.X * rotateMatrix[2]) + (tempCopy.Y * rotateMatrix[3]);
+            }
```
InfinityGhost: This seems to function less accurately the higher the rotation angle is. The positioning may also be incorrect here.

TabletDriverLib/Component/Area.cs
```diff
@@ -25,7 +25,7 @@ public Area(float width, float height, Point position, float rotation)
 
         public float[] GetRotationMatrix()
         {
-            var angle = Rotation * (Math.PI / 180);
+            var angle = -Rotation * (Math.PI / 180);
```
InfinityGhost: Not sure why this works, but it does.

PR #54, opened 2019-12-10
Pen button and Express Key functionality

MoeLava:
# Changes
- [x] Read pen button reports
  - [x] Generic TabletReport
- [x] Read aux button reports
  - [x] Gaomon AuxReport
  - [x] Wacom AuxReport
- [x] Bind button reports
- [x] Update tablet configurations
- [x] GUI
  - [x] Configuration Manager
  - [x] Bindings
- [x] Update README.md

# Testing
Tests are performed on a Gaomon S620


OpenTabletDriverGUI/Models/Settings.cs
```diff
@@ -188,6 +188,26 @@ public MouseButton TipButton
...
+
+        private MouseButton[] _padButtons;
+
+        [XmlArray("PadButtons")]
+        [XmlArrayItem("key")]
+        public MouseButton[] PadButtons
+        {
+            set => this.RaiseAndSetIfChanged(ref _padButtons, value);
+            get => _padButtons;
+        }
```
InfinityGhost: Name this `AuxButtons` and I'm fine with this implementation, assuming this is the buttons on the tablet body.

MoeLava: Resolved, I settled to calling it Express Keys in the user interface, and `AuxButtons` internally for length

TabletDriverLib/Tablet/ITabletReport.cs
```diff
@@ -7,5 +7,9 @@ public interface ITabletReport : IDeviceReport
         uint Lift { get; }
         Point Position { get; }
         uint Pressure { get; }
+
+        bool[] PenButtons { get; }
+        bool[] AuxButtons { get; }
+        bool IsAuxReport { get; }
```
InfinityGhost: We should separate this into a separate `IAuxReport` interface because not all tablet reports will include the aux buttons, but all SHOULD include PenButtons. On Wacom tablets for example, the aux reports are sent through a separate device.

TabletDriverLib/Output/AbsoluteMode.cs
```diff
@@ -43,11 +44,11 @@ public override TabletProperties TabletProperties
...
         public bool Clipping { set; get; }
-        public bool BindingsEnabled { set; get; }
+        public bool TipEnabled { set; get; }
         public float TipActivationPressure { set; get; }
-        public Dictionary<int, MouseButton> MouseButtonBindings { set; get; } = new Dictionary<int, MouseButton>()
+        public Dictionary<(BindingType type, int index), MouseButton> Bindings { set; get; } = new Dictionary<(BindingType type, int index), MouseButton>()
         {
-            { 0, MouseButton.None }
+            { (BindingType.Tip, 0), MouseButton.None }
         };
```
InfinityGhost: Turn the `BindingType, int` tuple into a separate class, maybe a `Binding<T>` class?

MoeLava: I think we can turn it into a `List<Binding>` instead

InfinityGhost: Lets see how it goes. Move those two properties into a `Binding` class and put it into a list.

MoeLava: On second thought, accessing the properties could be much harder with a `List`, I think I'll stick with the dictionary and a `Binding` struct

InfinityGhost:
Pull request is behind, going to update the branch so I can add changes.

TabletDriverLib/Interop/Cursor/MouseButton.cs
```diff
@@ -9,4 +9,11 @@ public enum MouseButton
         Backward = 4,
         Forward = 5
     }
+
+    public enum BindingType
+    {
+        Tip = 0,
+        Pen = 1,
+        Aux = 2
+    }
```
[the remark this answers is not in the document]
InfinityGhost: Scratch that, removing this enum entirely.

InfinityGhost:
Latest changes may have broke Gaomon, but it SHOULD work. We need to implement Wacom's method for Aux reports, which is through a separate HID device.

TabletDriverLib/Vendors/Gaomon/GaomonReportParser.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class GaomonReportParser : IReportParser<IDeviceReport>
+    {
+        public IDeviceReport Parse(byte[] data)
+        {
+            var isAuxReport = ((data[1] & (1 << 5)) != 0) & ((data[1] & (1 << 6)) != 0);
+            if (isAuxReport)
+                return new GaomonAuxReport(data);
+            else
+                return new TabletReport(data);
```
InfinityGhost: This should work to switch between the two report types, but testing is needed.

MoeLava: That works, however the class extends `IReportParser<IDeviceReport>`, whereas the assembly attempts to cast it to `IDeviceReportParser` returning null and using the generic `TabletReportParser` instead. This has been fixed.

OpenTabletDriverGUI/Views/MainWindow.xaml
```diff
@@ -106,32 +106,63 @@
...
+                        </Border>
+                        <Border Classes="r">
+                            <Grid RowDefinitions="Auto,*" Margin="5">
+                                <TextBlock Grid.Row="0" Text="Pen Buttons"/>
+                                <ItemsControl Grid.Row="1" Items="{Binding Settings.PenButtons}">
+                                    <ItemsControl.ItemTemplate>
+                                        <DataTemplate>
+                                            <ComboBox Grid.Row="1" Margin="5"
+                                                Items="{Binding #MouseButtonsBox.Items}"
+                                                SelectedItem="{Binding, Mode=TwoWay}"/>
```
InfinityGhost: Not exactly sure why, but these bindings aren't two-way like they should be. The property is quite literally set to be two way, but it just doesn't work. We need to find a fix

OpenTabletDriverGUI/Views/MainWindow.xaml
```diff
@@ -106,32 +106,63 @@
...
+                        </Border>
+                        <Border Classes="r">
+                            <Grid RowDefinitions="Auto,*" Margin="5">
+                                <TextBlock Grid.Row="0" Text="Express Keys"/>
+                                <ItemsControl Grid.Row="1" Items="{Binding Settings.AuxButtons}">
+                                    <ItemsControl.ItemTemplate>
+                                        <DataTemplate>
+                                            <ComboBox Grid.Row="1" Margin="5"
+                                                Items="{Binding #MouseButtonsBox.Items}"
+                                                SelectedItem="{Binding, Mode=TwoWay}"/>
```
InfinityGhost: Same issue as L143-145

OpenTabletDriverGUI/Views/MainWindow.xaml
InfinityGhost: Turns out you can't have a Two-Way binding on a property without a path, so this will have to be hardcoded in. I'll implement a hardcoded method for now, but I'd really like to make this dynamic.

OpenTabletDriverGUI/Views/MainWindow.xaml
```diff
@@ -106,32 +106,67 @@
...
+                                            <Setter Property="Items" Value="{Binding #MouseButtonsBox.Items}"/>
+                                        </Style>
+                                    </ItemsControl.Styles>
+                                    <ComboBox Margin="5" SelectedItem="{Binding Settings.AuxButtons[0], Mode=TwoWay}"/>
+                                    <ComboBox Margin="5" SelectedItem="{Binding Settings.AuxButtons[1], Mode=TwoWay}"/>
+                                    <ComboBox Margin="5" SelectedItem="{Binding Settings.AuxButtons[2], Mode=TwoWay}"/>
+                                    <ComboBox Margin="5" SelectedItem="{Binding Settings.AuxButtons[3], Mode=TwoWay}"/>
+                                </ItemsControl>
+                            </Grid>
+                        </Border>
```
InfinityGhost: I think we could make this a bit more visually appealing, right now everything is way oversized.

InfinityGhost: Made it a single row.

PR #116, opened 2020-04-06
Use evdev for cursor and keyboard handling

InfinityGhost:
# Changes

- Evdev
  - [x] Add EvdevCursorHandler
  - [x] Add EvdevKeyboardHandler
- Clean up deprecated X equivalents
  - [x] Remove XCursorHandler
  - [x] Remove XKeyboardHandler
# Testing
- [x] Xorg
- [x] Wayland Compositors


InfinityGhost:
We might want to come back to this later because of the lack of proper testing on Wayland, but as of right now it's not completely supported anyway. We'll work on supporting Wayland better with the addition of a proper Wayland display query method that doesn't depend on XOrg.

PR #127, opened 2020-05-02
Fix XP-Pen G430S_B configuration

Vlothx:
Fixes the ActiveReportID, cursor was not moving without this fix.


TabletDriverLib/Configurations/XP-Pen/G430S_B.json
```diff
@@ -1,18 +1,18 @@
...
+  "InputReportLength": 12,
+  "ReportParser": null,
+  "CustomInputReportLength": 0,
+  "CustomReportParser": null,
+  "Width": 101.6,
+  "Height": 76.2,
+  "MaxX": 10160.0,
+  "MaxY": 7620.0,
+  "MaxPressure": 8191,
+  "ActiveReportID": 12,
```
InfinityGhost, line 14: Would be a good idea to verify if this is working among multiple users, also across both supported platforms

usernamemadewheniplayedminecraft: On Arch Linux kernel v5.6.10-arch1-1 and Windows 10 LTSC v1809, I used the modified .json file from this PR. After using the modified .json file the tablet went from using full area to the area specified on OTD. Both drag and hover feel normal.

InfinityGhost: If we can get a test on Windows we should be good to call this working

InfinityGhost: Was just notified that it was tested on Windows by 2 people already and confirmed working. Going to merge.

PR #138, opened 2020-05-17
IAbsoluteMode improvements

skyborgff:
Added the display size as a variable.
This is needed for windows ink

### DIFF
--- a/OpenTabletDriver/Windows/MainWindowViewModel.cs
+++ b/OpenTabletDriver/Windows/MainWindowViewModel.cs
@@ -439,16 +439,9 @@ namespace OpenTabletDriver.Windows
                 };
                 Log.Write("Settings", $"Tablet area: {absoluteMode.Input}");
 
-                if (SelectedDisplay == null)
-                {
-                    absoluteMode.SelectedDisplay = VirtualScreen.Displays.Last();
-                }
-                else
-                {
-                    absoluteMode.SelectedDisplay = SelectedDisplay;
-                }
+                absoluteMode.VirtualScreen = VirtualScreen;
 
-                Log.Write("Settings", $"Screen area: {absoluteMode.SelectedDisplay}");
+                Log.Write("Settings", $"Screen area: {absoluteMode.VirtualScreen}");
 
                 absoluteMode.AreaClipping = Settings.EnableClipping;   
                 Log.Write("Settings", $"Clipping: {(absoluteMode.AreaClipping ? "Enabled" : "Disabled")}");
--- a/OpenTabletDriver/Windows/MainWindowViewModel.cs
+++ b/OpenTabletDriver/Windows/MainWindowViewModel.cs
@@ -441,32 +441,14 @@ namespace OpenTabletDriver.Windows
 
                 if (SelectedDisplay == null)
                 {
-                    absoluteMode.Screen = new Area
-                    {
-                        Width = VirtualScreen.Displays.Last().Width,
-                        Height = VirtualScreen.Displays.Last().Height,
-                        Position = new Point
-                        {
-                            X = VirtualScreen.Displays.Last().Position.X,
-                            Y = VirtualScreen.Displays.Last().Position.Y
-                        }
-                    };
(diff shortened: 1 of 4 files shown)


InfinityGhost:
Lets make this depend on #139 so we can use the `IDisplay` interface instead which can provide the entire maximum screen without unnecessary calculations

OpenTabletDriver/Windows/MainWindowViewModel.cs
```diff
@@ -439,6 +439,17 @@ public void ApplySettings(Settings settings)
...
+                if (SelectedDisplay == null)
+                {
+                    absoluteMode.SelectedDisplay = VirtualScreen.Displays.Last();
+                }
+                else
+                {
+                    absoluteMode.SelectedDisplay = SelectedDisplay;
+                }
+
+                Log.Write("Settings", $"Screen area: {absoluteMode.SelectedDisplay}");
```
InfinityGhost, line 451: We should pass through the entire virtual screen instead. I can't see why we would need just the selected display to be passed through, especially because it has to be selected manually at startup every time. It really should only be used for a maximum display size reference if anything at all.

OpenTabletDriver/Windows/MainWindowViewModel.cs
```diff
@@ -439,6 +439,17 @@ public void ApplySettings(Settings settings)
                 };
                 Log.Write("Settings", $"Tablet area: {absoluteMode.Input}");
 
+                if (SelectedDisplay == null)
+                {
+                    absoluteMode.SelectedDisplay = VirtualScreen.Displays.Last();
+                }
+                else
+                {
+                    absoluteMode.SelectedDisplay = SelectedDisplay;
```
InfinityGhost, line 448: The selected display property is only intended for setting the display area.

TabletDriverPlugin/IAbsoluteMode.cs
```diff
@@ -1,9 +1,12 @@
+using TabletDriverPlugin.Platform.Display;
+
 namespace TabletDriverPlugin
 {
     public interface IAbsoluteMode : IOutputMode
     {
         Area Input { set; get; }
         Area Output { set; get; }
+        IDisplay SelectedDisplay { set; get; }
```
InfinityGhost, line 9: Should rename this to match prior comments about this being focused on the entire display.

OpenTabletDriver/Windows/MainWindowViewModel.cs
skyborgff: If i pass the entire virtual screen i also need to pass the current selected screen index, because virtual screen doesnt say which screen is selected.

skyborgff: > It really should only be used for a maximum display size reference if anything at all.

Thats exactly what i needed. I need the maximum size of the current selected screen

InfinityGhost: > > It really should only be used for a maximum display size reference if anything at all.
> 
> Thats exactly what i needed. I need the maximum size of the current selected screen

Why does it need the currently selected screen? What if a user wants to have windows ink on multiple monitors?

skyborgff: > Why does it need the currently selected screen? What if a user wants to have windows ink on multiple monitors?
https://github.com/InfinityGhost/OpenTabletDriverPlugins/blob/34ed0a2d70f85071d6fdc145ec667a0e39e4107b/OpenTabletDriverPlugins/WindowsInk.cs#L383-L384
Its needed to map from the pixel position into the  HID required position.
If the user wants to use it in more than 1 screen he will use the virtual monitor. It should still map correctly, but i cant be sure as i only have 1 monitor.

InfinityGhost: Using the area's offset and the virtual screen should be enough in order to figure out the display position. I feel like by default it *should* be using the area position for that anyway.

skyborgff: this should be resolved now

PR #140, opened 2020-05-18
Kill oem drivers

skyborgff:
This will kill XP-Pen drivers on start. I dont know the other drivers processes names.
This should eventually become a setting.
Needs a RestarDrivers() function as companion, to restar the killed drivers on OTD close

- [x] KillDrivers()
- [ ] Setting for toggle on/off both kill and restart
- [ ] RestartDrivers()
- [ ] Add major oem drivers


InfinityGhost:
#130 will make this code irrelevant. After some thought, this really should be left up to the user to do.

PR #141, opened 2020-05-19
Fix tablet detection regression

InfinityGhost:
# Changes
- Fixes tablet detection regression from added support to `XP-Pen G640v2`

### DIFF
--- a/TabletDriverLib/Driver.cs
+++ b/TabletDriverLib/Driver.cs
@@ -33,12 +33,15 @@ namespace TabletDriverLib
             Log.Write("Detect", $"Searching for tablet '{tablet.TabletName}'");
             try
             {
-                var matching = Devices.Where(d => d.ProductID == tablet.ProductID && d.VendorID == tablet.VendorID);
-                var tabletDevice = matching.FirstOrDefault(d => d.GetMaxInputReportLength() == tablet.InputReportLength & d.GetMaxOutputReportLength() != 0);
+                var vendorMatch = Devices.Where(d => d.VendorID == tablet.VendorID);
+                var productMatch = vendorMatch.Where(d => d.ProductID == tablet.ProductID);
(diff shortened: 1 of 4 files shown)


TabletDriverLib/Driver.cs
```diff
@@ -33,12 +33,15 @@ public bool Open(TabletProperties tablet)
             Log.Write("Detect", $"Searching for tablet '{tablet.TabletName}'");
             try
             {
-                var matching = Devices.Where(d => d.ProductID == tablet.ProductID && d.VendorID == tablet.VendorID);
-                var tabletDevice = matching.FirstOrDefault(d => d.GetMaxInputReportLength() == tablet.InputReportLength & d.GetMaxOutputReportLength() != 0);
+                var vendorMatch = Devices.Where(d => d.VendorID == tablet.VendorID);
+                var productMatch = vendorMatch.Where(d => d.ProductID == tablet.ProductID);
+                var inputReportMatch = productMatch.Where(d => d.GetMaxInputReportLength() == tablet.InputReportLength);
+                var tabletDevice = inputReportMatch.Count() > 1 ? inputReportMatch.FirstOrDefault(d => d.GetMaxOutputReportLength() != 0) : inputReportMatch.FirstOrDefault();
```
InfinityGhost, line 39: I really think for futureproofing we should make it so that `OutputReportLength` would be an optional tablet identifier.
EX:
```cs
var tabletDevice = tablet.OutputReportLength > 0 ? inputReportMatch.FirstOrDefault(d => d.GetMaxOutputReportLength() == tablet.OutputReportLength) : inputReportMatch.FirstOrDefault();
```

skyborgff: Agree, and if configs get changed, we could add support for number of aux buttons

skyborgff:
After #143 gets merged this can as well

PR #144, opened 2020-05-19
Show display and tablet ratios

robens:
It will update the ratio as you change the size of each area. In case the variables could be placed somewhere else please tell, is my first time going through this code.

otd-ratio

### DIFF
--- a/OpenTabletDriver/Settings.cs
+++ b/OpenTabletDriver/Settings.cs
@@ -81,7 +81,7 @@ namespace OpenTabletDriver
 
         public float DisplayRatio
         {
-            set => this.RaiseAndSetIfChanged(ref _dRatio, value);
+            private set => this.RaiseAndSetIfChanged(ref _dRatio, value);
             get => (float)Math.Round(_dRatio * 1000f) / 1000f;
         }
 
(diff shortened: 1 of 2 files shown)


OpenTabletDriver/Settings.cs
```diff
@@ -73,10 +74,17 @@ public float DisplayHeight
...
                 if (LockAspectRatio)
                     TabletWidth = DisplayWidth / DisplayHeight * TabletHeight;
+                DisplayRatio = DisplayWidth / DisplayHeight;
             }
             get => _dH;
         }
 
+        public float DisplayRatio
+        {
+            set => this.RaiseAndSetIfChanged(ref _dRatio, value);
```
InfinityGhost, line 84: Make this setter private, its not intended to be set outside of this class.

OpenTabletDriver/Settings.cs
```diff
@@ -126,10 +135,17 @@ public float TabletHeight
...
                     _sizeChanging = false;
                 }
+                TabletRatio = TabletWidth / TabletHeight;
             }
             get => _tH;
         }
 
+        public float TabletRatio
+        {
+            set => this.RaiseAndSetIfChanged(ref _tRatio, value);
```
InfinityGhost, line 145: Make this setter private like DisplayRatio.

PR #150, opened 2020-06-17
Add support Vikoo 708 for macos

kroware:
Update HIDSharpCore to seize the device on macos to prevent system generate mouse events for us. 
Add InitString for macos. Same effect as hawku`s TabletDriver`s config
Add HIDReportID filtering, as report size is not enough for finding correct interface for Vikoo 708.
Fix mouse and key support on MacOS


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "HIDSharpCore"]
+	path = HIDSharpCore
+	url = https://github.com/kroware/HIDSharpCore
```
InfinityGhost, line 3: We're not going to use submodules, I'd much rather the changes to that library be pushed to my root HidSharpCore and this subsequently updated before merge.

OpenTabletDriver.Daemon/TabletDebuggerServer.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace OpenTabletDriver.Daemon
 {
-    internal class DeviceDebuggerServer : IDisposable
+    internal class DeviceDebuggerServer<T> : IDisposable where T : IDeviceReport
```
InfinityGhost, line 10: In some cases, devices will push aux reports through the tablet interface with an indicator saying that it is an aux report (i.e. [XP-Pen](https://github.com/InfinityGhost/OpenTabletDriver/blob/master/TabletDriverLib/Vendors/XP_Pen/XP_PenReportParser.cs#L10)). The endpoint automatically detects whether it is a tablet report or an aux report when need be. The only reason two separate device debugger servers exist is to avoid crossing the pipe threads and resulting in broken reports being sent. This change will only break those tablets.

OpenTabletDriver.UX.Mac/Program.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using Eto.Forms;
+
+namespace OpenTabletDriver.UX.Mac64
```
InfinityGhost, line 4: Namespace must match folder path for cleanliness.

OpenTabletDriver.UX/Debugging/PipeReader.cs
```diff
@@ -53,7 +53,7 @@ private async void Main()
                         }
                     }
                 }
-                catch (Exception)
+                catch (Exception e)
```
InfinityGhost, line 56: Make github actions happy lol

TabletDriverLib/Interop/USB/IUSBUtility.cs
```diff
@@ -0,0 +1,7 @@
+namespace TabletDriverLib.Interop.USB
+{
+    public interface IUSBUtility
+    {
+        public bool InitStrings(string hidDvicePath, byte[] array);
+    }
```
InfinityGhost, line 6: I would much rather have a function like this in the HidSharpCore library. All 3 (now) supported platforms have methods to query device strings, and the only one I haven't been able to get working is on Linux.
Adding proper support for a tablet on only 1 platform is something I'm not going to allow just for the sake of making this project consistent across all platforms to provide the same experience all around.

InfinityGhost: https://github.com/InfinityGhost/HIDSharpCore/pull/1
Make a PR targeting the `getstring` branch from the current state with these changes

TabletDriverPlugin/Tablet/TabletProperties.cs
```diff
@@ -199,8 +200,30 @@ public byte[] FeatureInitReport
...
+        /// <summary>
+        /// TO filter interface contain such report hid report id.
+        /// </summary>
+        /// <value></value>
+        [JsonProperty("HidReportID")]
+        public int HidReportID
+        {
+            set => this.RaiseAndSetIfChanged(ref _hidReportID, value);
+            get => _hidReportID;
+        }
```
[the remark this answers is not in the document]
kroware: http://who-t.blogspot.com/2018/12/understanding-hid-report-descriptors.html?m=1
It is used to identify different report type on same usb interface.
For example ：
```
Report ID (01)
Usage Page (Button)
Usage Minimum (1)
Usage Maximum (5)
Report Count (5)
Report Size (1)
Input (Data,Var,Abs)
Report Size (3)
Report Count (1)
Input (Cnst,Arr,Abs)

Report ID (02)
Usage Page (Generic Desktop)
Usage (X)
Usage (Y)
Report Count (2)
Report Size (16)
Input (Data,Var,Rel)
```

OpenTabletDriver.Daemon/TabletDebuggerServer.cs
kroware: https://github.com/InfinityGhost/OpenTabletDriver/blob/0f19d70ed2e4bdd3aafb1541c868be848db13c35/OpenTabletDriver.UX/Debugging/PipeReader.cs#L30
The setting is not passed to serializer.

InfinityGhost: https://github.com/InfinityGhost/OpenTabletDriver/blob/fa030458abc6016934ef25d72fd8b9589800b35d/OpenTabletDriver.Daemon/TabletDebuggerServer.cs#L28

You're still filtering out reports that aren't of type `T`, if anything we should at least change it to `IDeviceReport` on the daemon side if we don't remove this.

https://github.com/InfinityGhost/OpenTabletDriver/blob/fa030458abc6016934ef25d72fd8b9589800b35d/OpenTabletDriver.Daemon/DriverDaemon.cs#L42

TabletDriverPlugin/Tablet/TabletProperties.cs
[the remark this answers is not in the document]
InfinityGhost: I'm actually really glad you replied with this, you actually cleared that up the purpose of report IDs altogether for me.

TabletDriverLib/Tablet/TabletReport.cs
```diff
@@ -14,7 +14,13 @@ internal TabletReport(byte[] report)
...
             var y = BitConverter.ToUInt16(report, 4);
             Position = new Point(x, y);
-            Pressure = BitConverter.ToUInt16(report, 6);
+
+            //For unknown reason, sometimes, the tablet send incomplete report
+            //TODO: Reset the device after observing incomplete report
+            if (report.Length < 8)
+                Pressure = 0;
+            else
+                Pressure = BitConverter.ToUInt16(report, 6);
```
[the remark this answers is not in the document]
InfinityGhost: Done

PR #164, opened 2020-07-06
FIX: Opening URLs with spaces would return an error

flleeppyy:
This fixes the issue where if the user had 2 words in their name, that are spaced out (ex: C:\Users\Jack Smith\...), it would attempt to open a non existent directory (C:\Users\Jack)

image

### DIFF
--- a/NativeLib/Tools.cs
+++ b/NativeLib/Tools.cs
@@ -9,7 +9,7 @@ namespace NativeLib
         {
             if (PlatformInfo.IsWindows)
             {
-                var startInfo = new ProcessStartInfo("cmd", $"/c start {url.Replace("&", "^&")}")
+                var startInfo = new ProcessStartInfo("cmd", $"/c start \"\" \"{url.Replace("&", "^&")}\"")
                 {
                     CreateNoWindow = true
                 };


NativeLib/Tools.cs
```diff
@@ -9,7 +9,7 @@ public static void OpenUrl(string url)
         {
             if (PlatformInfo.IsWindows)
             {
-                var startInfo = new ProcessStartInfo("cmd", $"/c start {url.Replace("&", "^&")}")
+                var startInfo = new ProcessStartInfo("cmd", $"/c start \"\" \"{url.Replace("&", "^&")}\"")
```
InfinityGhost, line 12: What's the need for the extra `\"\"`?

flleeppyy: https://stackoverflow.com/questions/6376113/how-do-i-use-spaces-in-the-command-prompt
`start` needs a title for the command, so we just throw an empty string, then the command.
`start "" "C:\Users\Triple Word User\..."`

InfinityGhost: That's quite odd but okay.

InfinityGhost:
Uhh this was pointed to the wrong branch I just realized after merge lol

InfinityGhost:
May I request you re-create this PR aimed at the `master` branch, I'd like to give you credit for this catch lol

flleeppyy:
Alright then.

PR #178, opened 2020-07-11
Add Wayland display support

tadeokondrak:
This allows talking to Wayland compositors directly, rather than relying
on XWayland data which often doesn't reflect the exact setup on Wayland,
and of course doesn't work without XWayland running.

It tries to use xdg_output for precise data, but falls back to wl_output
information if that protocol is missing.

### DIFF
--- a/TabletDriverLib/Interop/Display/WaylandOutput.cs
+++ b/TabletDriverLib/Interop/Display/WaylandOutput.cs
@@ -1,0 +1,83 @@
+using TabletDriverPlugin;
+using TabletDriverPlugin.Platform.Display;
+using WaylandNET.Client.Protocol;
+
+namespace TabletDriverLib.Interop.Display
+{
+    public class WaylandOutput : IDisplay, WlOutput.IListener, ZxdgOutputV1.IListener
+    {
+        public WlOutput WlOutput { private set; get; }
+        public ZxdgOutputV1 XdgOutput { set; get; }
+
+        public int Index { private set; get; }
+        public float Width { private set; get; }
+        public float Height { private set; get; }
+        public Point Position { private set; get; }
+
+        private string _name, _description;
+
+        public WaylandOutput(WlOutput wlOutput, int index)
+        {
+            WlOutput = wlOutput;
+            Index = index;
+        }
+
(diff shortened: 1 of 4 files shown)


TabletDriverLib/Interop/Platform.cs
```diff
@@ -45,7 +46,16 @@ public static class Platform
...
-                return new XScreen();
+            {
+                try
+                {
+                    return new WaylandDisplay();
+                }
+                catch (SocketException)
+                {
+                    return new XScreen();
+                }
```
InfinityGhost, line 57: Can't you detect X11 or Wayland via environment variables instead of this?

TabletDriverLib/Interop/Display/WaylandOutput.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        public WlOutput WlOutput { private set; get; }
+        public ZxdgOutputV1 XdgOutput { set; get; }
+
+        public int Index { private set; get; }
+        public float Width { private set; get; }
+        public float Height { private set; get; }
+        public Point Position { private set; get; }
+
+        private string _name, _description;
```
InfinityGhost, line 17: Personally would have made these 
```cs
public string Name { private set; get; }
public string Description { private set; get; }
```

TabletDriverLib/Interop/Platform.cs
tadeokondrak: I'm not really sure. I don't think any compositors *don't* set WAYLAND_DISPLAY, but all programs fall back to `wayland-0` if it isn't set. From what I've seen, most programs try Wayland first then fall back to X11 if it errors out. If you want to check WAYLAND_DISPLAY instead, it'll probably still work for everyone.

InfinityGhost: Checking `WAYLAND_DISPLAY` seems like a more sane method of handling this, along with checking if `DISPLAY` has a value would be smart for Xorg.

InfinityGhost: Throwing an exception if both don't exist might be worthwhile as well

PR #202, opened 2020-07-18
Fix filters getting reconstructed often

X9VoiD:
Switch to using List instead of IEnumerable for iterating over the filters to avoid reconstruction in foreach()

### DIFF
--- a/TabletDriverPlugin/Output/AbsoluteOutputMode.cs
+++ b/TabletDriverPlugin/Output/AbsoluteOutputMode.cs
@@ -13,15 +13,16 @@ namespace TabletDriverPlugin.Output
         private float[] _rotationMatrix;
         private float _halfDisplayWidth, _halfDisplayHeight, _halfTabletWidth, _halfTabletHeight;
         private float _minX, _maxX, _minY, _maxY;
-        
-        private IEnumerable<IFilter> _filters, _preFilters, _postFilters;
+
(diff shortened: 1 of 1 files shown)


TabletDriverPlugin/Output/AbsoluteOutputMode.cs
```diff
@@ -13,15 +13,16 @@ public abstract class AbsoluteOutputMode : BindingHandler, IOutputMode
         private float[] _rotationMatrix;
         private float _halfDisplayWidth, _halfDisplayHeight, _halfTabletWidth, _halfTabletHeight;
         private float _minX, _maxX, _minY, _maxY;
-        
-        private IEnumerable<IFilter> _filters, _preFilters, _postFilters;
+
+        private IEnumerable<IFilter> _filters;
+        private List<IFilter> _preFilters, _postFilters;
```
InfinityGhost, line 18: This could have *probably* been left alone as it wouldn't reconstruct the class just because the inherited type is IEnumerable<T> as List<T> implements it, but its a minor nitpick.

X9VoiD: For some reason, even with .ToList(), IEnumerable + Linq reconstructs the classes, so if I put the .ToList() on the foreach() the classes are still reconstructed.

PR #208, opened 2020-07-21
Finally fix filter settings not getting applied

X9VoiD:
There were two problems, the incomplete name when querying the PluginSetting dictionary, and the filter being reconstructed (yeah, it was actually just half-fixed)

Now the filters are only constructed once, instead of three times.

### DIFF
--- a/TabletDriverPlugin/Output/AbsoluteOutputMode.cs
+++ b/TabletDriverPlugin/Output/AbsoluteOutputMode.cs
@@ -14,15 +14,12 @@ namespace TabletDriverPlugin.Output
         private float _halfDisplayWidth, _halfDisplayHeight, _halfTabletWidth, _halfTabletHeight;
         private float _minX, _maxX, _minY, _maxY;
 
-        private IEnumerable<IFilter> _filters;
-        private List<IFilter> _preFilters, _postFilters;
+        private List<IFilter> _filters;
         public IEnumerable<IFilter> Filters
         {
             set
             {
-                _filters = value;
-                _preFilters = value.Where(f => f.FilterStage == FilterStage.PreTranspose).ToList();
-                _postFilters = value.Where(f => f.FilterStage == FilterStage.PostTranspose).ToList();
+                _filters = value.ToList();
             }
             get => _filters;
         }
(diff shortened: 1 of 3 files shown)


TabletDriverPlugin/Output/AbsoluteOutputMode.cs
```diff
@@ -14,15 +14,17 @@ public abstract class AbsoluteOutputMode : BindingHandler, IOutputMode
...
             {
-                _filters = value;
-                _preFilters = value.Where(f => f.FilterStage == FilterStage.PreTranspose).ToList();
-                _postFilters = value.Where(f => f.FilterStage == FilterStage.PostTranspose).ToList();
+                _filters = value.ToList();
+                foreach (IFilter filter in _filters)
+                    if (filter.FilterStage == FilterStage.PreTranspose)
+                        _preFilters.Add(filter);
+                    else if (filter.FilterStage == FilterStage.PostTranspose)
+                        _postFilters.Add(filter);
```
InfinityGhost, line 27: This change still seems entirely unnecessary. By doing this you need to clear the list every time the filters change or settings apply.

X9VoiD: I forgot that with this change, _pre and _post will leak the previous _filters, then Add() the new ones.

X9VoiD: Clearing and rebuilding pre and post every filter and settings change is lighter than doing it on the fly during Transpose()

is what I've come up with.

TabletDriverPlugin/Output/AbsoluteOutputMode.cs
```diff
@@ -14,15 +14,19 @@ public abstract class AbsoluteOutputMode : BindingHandler, IOutputMode
...
-                _preFilters = value.Where(f => f.FilterStage == FilterStage.PreTranspose).ToList();
-                _postFilters = value.Where(f => f.FilterStage == FilterStage.PostTranspose).ToList();
+                _filters = value.ToList();
+                _preFilters.Clear();
+                _postFilters.Clear();
+                foreach (IFilter filter in _filters)
+                    if (filter.FilterStage == FilterStage.PreTranspose)
+                        _preFilters.Add(filter);
+                    else if (filter.FilterStage == FilterStage.PostTranspose)
+                        _postFilters.Add(filter);
```
InfinityGhost, line 29: Apply this change to RelativeOutputMode and I'll merge

X9VoiD: Applied

PR #209, opened 2020-07-24
Fix plugin dependency not loading in shared context

X9VoiD:
Might introduce problems since plugins aren't isolated anymore. I tried creating a custom AssemblyLoadContext for isolation and dependency resolution, but it became too isolated that interfaces didn't work anymore (casting does not work).

### DIFF
--- a/TabletDriverLib/PluginManager.cs
+++ b/TabletDriverLib/PluginManager.cs
@@ -4,6 +4,7 @@ using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Loader;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


InfinityGhost:
I don't see any reason this would introduce problems compared to before. I would like to test if this solves the assembly loading issues on the UX Project as well

X9VoiD:
I've got a proper isolation of plugins now. The problem was I followed Microsoft's tutorial and just took it for granted. But I'll reserve those changes to a separate branch and see if this PR as it is enough.

InfinityGhost:
This fixes a lot of assembly loading issues that I didn't expect to. Upcoming PR fixing all of the GUI long-path references

PR #212, opened 2020-07-25
Fix occurences of plugin double reference in UX

X9VoiD:


OpenTabletDriver.UX/Controls/PluginManager.cs
```diff
@@ -27,26 +27,23 @@ public PluginManager()
...
             {
-                _plugins = value;
-                _pluginList.Items.Clear();
-                foreach (var plugin in Plugins)
-                    _pluginList.Items.Add(string.IsNullOrWhiteSpace(plugin.Name) ? plugin.Path : plugin.Name);
+                var pluginRef = new PluginReference(type);
+                if (type != typeof(T) && !Plugins.Contains(pluginRef))
+                {
+                    Plugins.Add(pluginRef);
+                }
```
InfinityGhost, line 36: Why did you change the `Plugins` set/get when all you really needed to do was check if it was already loaded?

X9VoiD: A bug occurs that if I did the check without changing set/get, the Plugins get cleared, as I have seen on debugger.
Essentially, the check on set/get have to happen after checking that there is no duplicate, not before (which is triggered by L23)

X9VoiD: And when L34 `Plugins = new List<PluginReference>(pluginRefs);` is kept there the plugins still contained the duplicate references.

PR #244, opened 2020-08-03
Hide window in alt+tab when minimized (Windows)

### LINKED ISSUE #242: OpenTablet Driver is visible when ALT-TABBING when it is hiden in the system tray
## Description

OpenTablet Driver is visible when ALT-TABBING when it is hiden in the system tray 

## System Information:

Option            | Value
------------------|------------------
Operating System  |  Windows 10
Software Version  |  v0.3.2
Tablet            |  CTL-490

EXtremeExploit:
closes #242

### DIFF
--- a/OpenTabletDriver.UX/TrayIcon.cs
+++ b/OpenTabletDriver.UX/TrayIcon.cs
@@ -4,7 +4,7 @@ namespace OpenTabletDriver.UX
 {
     public class TrayIcon
     {
-        public TrayIcon(Window window)
+        public TrayIcon(MainForm window)
         {
             var showWindow = new ButtonMenuItem
             {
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.UX/TrayIcon.cs
```diff
@@ -4,15 +4,16 @@ namespace OpenTabletDriver.UX
 {
     public class TrayIcon
     {
-        public TrayIcon(Window window)
+        public TrayIcon(MainForm window)
```
InfinityGhost, line 7: This shouldn't matter as `MainForm -> Form -> Window`

EXtremeExploit: The one with the `Show()` method is MainForm, not `Window`, i could do `((MainForm)window).Show();` but looks horrible, also the class that we are passing is MainWindow, not Window so it would be nice to have the actual class

InfinityGhost: Understandable, keep it as is then.

PR #245, opened 2020-08-04
[WIP] Initialize XP Pen tablet outputs

X9VoiD:


X9VoiD:
> `0x02 0xB0 0x04 0x00 0x00 0x00 0x00 0x00`    

works but devices requiring
> `0x02 0xB0 0x02 0x00 0x00 0x00 0x00 0x00`    

seems to need a different way of sending the init before the HidDevice can be accessed. (like G640)

X9VoiD:
Confirmed devices to work without XP Pen driver
- G640
- G640S

InfinityGhost:
CodeFactor failed due to the site being down, overriding check requirement to merge

PR #251, opened 2020-08-10
show when the system tray icon is left clicked, fix show option and hide from task manager applications when hidden

EXtremeExploit:


EXtremeExploit:
This PR makes the next changes:
* Shows the GUI whenever the system tray icon is left clicked
* Fixes the GUI not displayed when clicked the show option from the system tray
* Hides the process in applications from task manager when the GUI is minimized

InfinityGhost:
If we could figure out the correct way to make it hide from the alt-tab menu we'd be golden

EXtremeExploit:
What hides from alt tab is the property Visible of the window, setting it to false, hides it. HOWEVER when it is hidden any change to the WindowState won't actually happen, and since the Visible property is not directly accessible we have to call Window.Show() which simply makes Visible true, so we have to have this first to make everything work nicely

PR #261, opened 2020-08-24
Fix Reset to defaults message box showing on first start

Airkek:


InfinityGhost:
Good catch lol

OpenTabletDriver.UX/MainForm.cs
```diff
@@ -629,9 +629,9 @@ public Settings Settings
             get => App.Settings;
         }
 
-        private async Task ResetSettings()
+        private async Task ResetSettings(bool force=true)
```
InfinityGhost, line 632: Just for code readability/consistency sake, space out `bool force=true` to `bool force = true`

PR #269, opened 2020-09-02
Get MouseBinding's mouse handler from DriverState

X9VoiD:
# Changes
- [x] Add `IVirtualPointer Pointer` to IOutputMode
- [x] Implement `Pointer` through IVirtualTablet and IVirtualMouse
- [x] Make MouseBinding use `DriverState.OutputMode.Pointer` for processing MouseUp() and MouseDown()

This fixes the BindingHandler issues when plugins implement their own IOutputMode and IVirtualPointer (including IVirtualTablet and IVirtualMouse)

# Issues
Fixes TouchEmu Plugin

### DIFF
--- a/TabletDriverLib/Binding/MouseBinding.cs
+++ b/TabletDriverLib/Binding/MouseBinding.cs
@@ -16,7 +16,7 @@ namespace TabletDriverLib.Binding
         {
             get 
             {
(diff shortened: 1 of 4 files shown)


InfinityGhost (approved):
Reasonable fix, I figured that this would have caused problems at some point

PR #277, opened 2020-09-07
Add Huion HS610 configuration

### LINKED ISSUE #247: Add huion HS610 support
## Description

VID is 0x256C, PID is 0x006D

Two ~~endpoints~~ interfaces: 0 (vendor-defined device) and 1 (pen)

Pen messages appear to be 14 bytes long (or 13? idk, zeros start at 13'th byte), first 12 of them are actual useful data.

Here is a little snippet of how I was reading data from it back when I played around with it in Feb: (can't believe it took me so long to stumble upon this project)
```
const char* DecodeState(uint16_t State)
{
    switch (State)
    {
    case 10: return "HoverStop";     // 0x 00 0A

    case 49162: return "Hover";      // 0x C0 0A
    case 49674: return "Hover + A";  // 0x C2 0A
    case 50186: return "Hover + B";  // 0x C4 0A

    case 49418: return "Touch";      // 0x C1 0A
    case 49930: return "Touch + A";  // 0x C3 0A
    case 50442: return "Touch + B";  // 0x C5 0A

    default: return "Unknown";
    }
}

struct PenInput
{
    PenInput(){}
    PenInput(UCHAR* buffer){ReadFromBuffer(buffer);}
    uint16_t State=0;
    uint16_t Xpos=0;
    uint16_t Ypos=0;
    uint16_t Pressure=0;
    int8_t Xtilt=0;
    int8_t Ytilt=0;

    void ReadFromBuffer(UCHAR* buffer)
    {
        State = *(uint16_t*)(buffer);
        Xpos = *(uint16_t*)(buffer + 2);
        Ypos = *(uint16_t*)(buffer + 4);
        Pressure = *(uint16_t*)(buffer + 6);
        Xtilt = *(int8_t*)(buffer + 8);
        Ytilt = *(int8_t*)(buffer + 9);
    }

    void Print()
    {
        printf("State=%05u (%-9s), Xpos=%05u, Ypos=%05u, Pressure=%05u, Xtilt=%03d, Ytilt=%03d\n", State, DecodeState(State), Xpos, Ypos, Pressure, Xtilt, Ytilt);
    }
};
```

Both X and Y pos axis go from 0 to 0x7FFF (32767), Pressure goes up to 8191, tilt axis reports are between -60 and 60 signed.

Report rate on spec is supposed to be 266 Hz, but with libusb I only could poll at 220 Hz, maybe I was doing something wrong.

AbstractQbit:
Not sure why I was getting 0-32767 ranges mentioned in #247 with libusb, in OTD tablet debugger x is 0-50800, y is 0-31750, so those were set.

Everything apart from ranges, dimensions, name and device string was copied from hs611.

Now OTD seems to work, so this closes #247.

### DIFF
--- a/TabletDriverLib/Configurations/Huion/HS610.json
+++ b/TabletDriverLib/Configurations/Huion/HS610.json
@@ -43,7 +43,6 @@
     "201": "HUION_T194_190307"
   },
   "InitializationStrings": [
-    201,
     200
   ]
 }
--- a/TabletDriverLib/Configurations/Huion/HS610.json
+++ b/TabletDriverLib/Configurations/Huion/HS610.json
@@ -1,0 +1,49 @@
+{
+  "Name": "Huion HS610",
+  "DigitizerIdentifier": {
+    "VendorID": 9580,
+    "ProductID": 109,
+    "InputReportLength": 12,
+    "OutputReportLength": 0,
+    "ReportParser": null,
+    "FeatureInitReport": null,
+    "OutputInitReport": null
+  },
+  "AlternateDigitizerIdentifier": {
+    "VendorID": 0,
+    "ProductID": 0,
+    "InputReportLength": 0,
+    "OutputReportLength": 0,
+    "ReportParser": null,
+    "FeatureInitReport": null,
+    "OutputInitReport": null
+  },
+  "AuxilaryDeviceIdentifier": {
+    "VendorID": 0,
+    "ProductID": 0,
+    "InputReportLength": 0,
+    "OutputReportLength": 0,
+    "ReportParser": null,
+    "FeatureInitReport": null,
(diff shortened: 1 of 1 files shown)


X9VoiD:
Do not use index 201 for DeviceString in HS lineups since they can vary from device to device of the same model as discovered previously

Use index 2 instead.

X9VoiD:
Weird, just found that it uses the old format somehow on index 2, and is using the new format on index 201, keep it, but remove 201 from initialization to remove redundant operation.

InfinityGhost:
Is this ready for merge?

AbstractQbit:
Seems to work ok, so yes. DeviceString can be changed to regex in corresponding branch, along with S620

PR #278, opened 2020-09-07
Use regular expressions for matching device strings

InfinityGhost:


TabletDriverLib/Driver.cs
```diff
@@ -232,7 +233,7 @@ private bool DeviceMatchesAllStrings(HidDevice device, TabletProperties tablet)
                 {
                     // Iterate through each device string, if one doesn't match then its the wrong configuration.
                     var str = device.GetDeviceString(matchQuery.Key);
-                    if (str != matchQuery.Value)
+                    if (!Regex.IsMatch(matchQuery.Value, str))
```
AbstractQbit, line 236: static IsMatch has arguments (string input, string pattern), looks like now they are placed the other way around

InfinityGhost: whoops lol nice catch

PR #279, opened 2020-09-08
Replace bulky math in AbsoluteOutputMode with a single affine transformation using precomputed matrix

AbstractQbit:


TabletDriverPlugin/Output/AbsoluteOutputMode.cs
```diff
@@ -85,6 +86,36 @@ internal void UpdateCache()
             _maxY = Output?.Position.Y + Output?.Height - _halfDisplayHeight ?? 0;
         }
 
+        internal Matrix3x2 CalculateTransformation()
```
InfinityGhost, line 89: I like this, however I'd like to see it changed to take in parameters.
```cs
CalculateTransform(Area input, Area output, TabletProperties tablet)
```

InfinityGhost (changes requested):

InfinityGhost (approved):

PR #293, opened 2020-09-14
Comply to OTD rename changes, and other fixes

X9VoiD:
# Changes
- [x] Update references to custom parsers
- [x] Update G640S config according to report descriptor
- [x] Update H430P, HS64, and HS611 detection
- [x] Use regex for H640P, Inspiroy Q11K, and New 1060 Plus
- [x] Fix G430S_B's OutputInitReport


X9VoiD:
H610 Pro, and osu!tablet are the only remaining huion tablets in `Configuration` with weird device identifiers

PR #296, opened 2020-09-15
Update Windows input injection

### LINKED ISSUE #265: [Windows] Active monitor resolution uses maximum for display
## Description
as the title state.
it seem to depend on the monitor position in the windows display settings
otd not broken
here, i'm using a different monitor with a small resolution (my android tablet in reality) in that position, OTD work properly and the cursor doesn't get stuck

otd broken
here same thing, except OTD is broken and cursor get stuck below the a straight line (illustrated in red) at the same height as my smaller res monitor, this will probably be the same thing if i move my monitor in the setting on the left of display 2,  moving that smaller res monitor on top and under of that monitor is untested (as the app i use like to break if i tweak display position in the settings too much)

also, OTD only use the native resolution of the monitor, not the one set in windows settings

which mean having a 1080 screen and a 4K screen both set to 1080p on windows will result in OTD breaking and the cursor to get stuck in the middle of the selected monitor (right side or left side depending on the selected monitor in OTD)

## System Information:

Option            | Value
------------------|------------------
Operating System  |  Windows
Software Version  |  0.3.2 and higher
Tablet            |  All tablet

X9VoiD:
# Changes
* Allocate `INPUT[]` only once
* Tell Windows to not coalesce the sent inputs
* Use `SendInput()` instead of `mouse_event()` for consistency and efficiency sake (reuse of `INPUT[]`)
* Use the user-defined screen resolution for the display configuration

# Issues
* Fixes #265
* Fixes Windows DPI scaling issues

### DIFF
--- a/OpenTabletDriver.Native/Windows/Display/DisplayInfo.cs
+++ b/OpenTabletDriver.Native/Windows/Display/DisplayInfo.cs
@@ -4,22 +4,22 @@ namespace OpenTabletDriver.Native.Windows
 {
     public class DisplayInfo
     {
-        public DisplayInfo(Rect monitorArea, Rect workingArea, uint flags)
+        public DisplayInfo(Rect monitorArea, uint flags, float scaling = 1)
         {
             MonitorArea = monitorArea;
-            WorkingArea = workingArea;
+            Scaling = scaling;
             IsPrimary = ((MONITORINFOF)flags).HasFlag(MONITORINFOF.PRIMARY);
         }
 
         public Rect MonitorArea { private set; get; }
-        public Rect WorkingArea { private set; get; }
 
-        public int Width => WorkingArea.right - WorkingArea.left;
-        public int Height => WorkingArea.bottom - WorkingArea.top;
-        public int Top => WorkingArea.top;
-        public int Left => WorkingArea.left;
-        public int Bottom => WorkingArea.bottom;
-        public int Right => WorkingArea.right;
+        public int Width => MonitorArea.right - MonitorArea.left;
+        public int Height => MonitorArea.bottom - MonitorArea.top;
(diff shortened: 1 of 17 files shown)


OpenTabletDriver.Native/Windows/Display/DisplayInfo.cs
```diff
@@ -14,12 +14,12 @@ public DisplayInfo(Rect monitorArea, Rect workingArea, uint flags)
...
-        public int Top => MonitorArea.top;
-        public int Left => MonitorArea.left;
-        public int Bottom => MonitorArea.bottom;
-        public int Right => MonitorArea.right;
+        public int Width => WorkingArea.right - WorkingArea.left;
+        public int Height => WorkingArea.bottom - WorkingArea.top;
+        public int Top => WorkingArea.top;
+        public int Left => WorkingArea.left;
+        public int Bottom => WorkingArea.bottom;
+        public int Right => WorkingArea.right;
```
InfinityGhost, line 22: Whats up with this? This just doesn't seem right. IIRC in the windows shell the working area is the area between the taskbar and the rest of the screen space.

X9VoiD: It is tested and it's working (apparently). Cursor can go anywhere.

X9VoiD: Yeah it's wrong, will find other ways to find user-defined resolution

X9VoiD: Pushed a new commit and the resolution is properly detected now.

InfinityGhost (approved):

PR #302, opened 2020-09-17
Add Interpolation capabilities to OTD

X9VoiD:
# Changes
* Added interpolation


OpenTabletDriver.Plugin/Output/IOutputMode.cs
```diff
@@ -6,7 +6,7 @@ namespace OpenTabletDriver.Plugin.Output
 {
     public interface IOutputMode
     {
-        void Read(IDeviceReport report);
+        void Read(ITabletReport report);
```
InfinityGhost, line 9: Do NOT do this, this only creates problems.

InfinityGhost: We rely on all report types being passed to the output mode, and in the read function itself will cast to ITabletReport if needed.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Output/IOutputMode.cs
X9VoiD: So we cast it back to IDeviceReport after modifying the ITabletReport in the interpolator?

X9VoiD: Reverted IOutputMode's Read back to IDeviceReport. Tested working on G640S normal and aux report

X9VoiD:
Helps with #303

X9VoiD:
Will rework this again, there are issues found after fiddling with it.

## Issues
* Pen Arrive/Left spams for a while when pen arrives then settles down
* InterpolationEngine rarely gets stuck on Pen Arrive/Left events
* InterpolatorArgs is unclear with its intents
    * TimeDelta have different meaning depending on where it's called (will probably just leave this to the plugin)
    * Position/Pressure on Interpolate is the same on NewReport (Interpolate's Position/Pressure args should provide current interpolation's position)

InfinityGhost:
I really think this PR should take a different direction, more towards giving up the interfaces to implement their own interpolator instead of writing one directly to just be used. I think filters should be the ones implementing the interpolator interface.

X9VoiD:
The `InterpolationEngine` currently only provides these additional functions

* Create fake reports every specified Hz, to be modified by implementers of abstract class `Interpolator` ([InterpolationEngine.cs#L44](https://github.com/InfinityGhost/OpenTabletDriver/blob/a1edc2b3808fb63ffec56013943fefebef71feaa/OpenTabletDriver.Plugin/Tablet/Interpolator/InterpolationEngine.cs#L44))
* Dynamically switch interpolation on and off when pen arrives/leaves (fixes problems with cursor getting stuck due to interpolator still sending fake reports)
    * This function needs a calculated report rate to work properly, hence the `DriverState.ReportRate`

So rest assured the interpolation actually only happens on the active `Interpolator` plugin

X9VoiD:
New changes pushed

### Changes
* Lose `DriverState.ReportRate` for simplicity
* Make it clear that `NewReport()` args are read-only
* Make it clear that `Interpolate()` args are write-only
* Fixed rare cases where OTD gets stucked in PenLeft event
* Fixed PenArrived/PenLeft spamming

There are essentially no issues anymore with the PR (except slightly higher cpu usage due to busy wait). Both hawku smoothing's and devocub antichatter's interpolation can be implemented with the abstract class Interpolation as a dll plugin

InfinityGhost (changes requested):

InfinityGhost:
Resolve merge conflicts after #322 and #288

X9VoiD:
Merge conflicts resolved and formatting issues fixed

OpenTabletDriver.Plugin/Micro/MicroStopwatch.cs
```diff
@@ -0,0 +1,32 @@
...
+            1000000D / Frequency;
+
+        public MicroStopwatch()
+        {
+            if (!IsHighResolution)
+            {
+                throw new Exception(
+                    "On this system the high-resolution " +
+                    "performance counter is not available"
+                );
```
InfinityGhost, line 20: I'm sorry lmfao
```cs
throw new PlatformNotSupportedException(
    "On this system the high-resolution " +
    "performance counter is not available"
);
```

OpenTabletDriver.Plugin/Micro/MicroTimer.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                return true;
+            }
+
+            return _threadTimer.Join(timeoutInMilliSec);
+        }
+
+        void NotificationTimer(ref long timerIntervalInMicroSec,
+                               ref long ignoreEventIfLateBy,
+                               ref bool stopTimer)
```
InfinityGhost, line 112:
```cs
void NotificationTimer(
    ref long timerIntervalInMicroSec,
    ref long ignoreEventIfLateBy,
    ref bool stopTimer
)
{
    // stuff
```

OpenTabletDriver.Plugin/Micro/MicroTimer.cs
```diff
@@ -0,0 +1,158 @@
...
+{
+    /// <summary>
+    /// MicroTimer class
+    /// </summary>
+    public class MicroTimer
+    {
+        public delegate void MicroTimerElapsedEventHandler(
+                             object sender,
+                             MicroTimerEventArgs timerEventArgs);
+        public event MicroTimerElapsedEventHandler MicroTimerElapsed;
```
InfinityGhost, line 13: This will suffice. 
```cs
public event EventHandler<MicroTimerEventArgs> MicroTimerElapsed;
```

OpenTabletDriver.Plugin/Micro/MicroTimerEventArgs.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+
+namespace OpenTabletDriver.Plugin.Micro
```
InfinityGhost, line 3: Reason for the naming of the namespace? I think `OpenTabletDriver.Plugin.Timers` or something similar makes more sense.

OpenTabletDriver.Plugin/Tablet/ITabletReport.cs
```diff
@@ -5,8 +5,8 @@ namespace OpenTabletDriver.Plugin.Tablet
     public interface ITabletReport : IDeviceReport
     {
         uint ReportID { get; }
-        Vector2 Position { get; }
-        uint Pressure { get; }
+        Vector2 Position { get; set; }
+        uint Pressure { get; set; }
```
InfinityGhost, line 9: What's the need for this change? I personally think a new report should be created for each new report instead of modifying an existing one.

OpenTabletDriver/Driver.cs
```diff
@@ -269,9 +272,7 @@ private void OnReport(object sender, IDeviceReport report)
         {
             if (EnableInput && OutputMode?.TabletProperties != null)
             {
-                OutputMode?.Read(report);
-                if (OutputMode is IBindingHandler<IBinding> handler)
-                    handler.HandleBinding(report);
+                InterpolationEngine.HandleReport(report);
```
InfinityGhost, line 275: Not sure how I feel about this change, maybe that function in `InterpolationEngine` should be in `Driver`?

InfinityGhost (changes requested):
Okay that's a full review now.

OpenTabletDriver.Plugin/Micro/MicroStopwatch.cs
X9VoiD: lmao

OpenTabletDriver.Plugin/Micro/MicroTimerEventArgs.cs
X9VoiD: The code from that namespace is from [MicroLib](https://www.codeproject.com/Articles/98346/Microsecond-and-Millisecond-NET-Timer) so the name came from there

OpenTabletDriver.Plugin/Tablet/ITabletReport.cs
X9VoiD: At the time i didn't know we could make ITabletReports by making a dummy like you did before. Will make the change

InfinityGhost: I would suggest calling it something like `InterpolatedTabletReport`

OpenTabletDriver/Driver.cs
X9VoiD: We can move InterpolationEngine.HandleReport to Driver but that HandleReport needs locking, StateLock, so there won't be race conditions when HandleReport and Interpolate fires at the same time. I wanted to keep references to StateLock to only inside the InterpolationEngine itself

OpenTabletDriver.Plugin/Micro/MicroTimerEventArgs.cs
X9VoiD: Will rename to `OpenTabletDriver.Plugin.Timers`

OpenTabletDriver/Driver.cs
InfinityGhost: Maybe the functions in `InterpolationEngine` should be in `Driver` then? I'm not too keen on the use of the DriverState class inside of the driver itself when the driver itself is already available.

X9VoiD: Oh uhh, my bad, I don't think the HandleReport could be moved. I forgot that the reason the `InterpolationEngine` is there is so we could use PenInRange's `internal set` and properly invoke PenArrive event. The PenInRange's `set` is not available from driver. And PenArrive, PenInRange is in DriverState so plugins could use it, or future output modes that can make use of "pen left" events. If you could think of a way to move stuff around though then I'll comply, double HandleReport looks bad I agree

OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
```diff
@@ -72,11 +75,15 @@ protected virtual void InterpolateHook()
         {
             lock (this.stateLock)
             {
-                if (((DateTime.UtcNow - lastTime).TotalMilliseconds < 150) && this.inRange)
+                if (((DateTime.UtcNow - this.lastTime).TotalMilliseconds < this.reportMsAvg * 1.5) && this.inRange)
```
InfinityGhost, line 78: We could maybe make this non-linear so that it would work across virtually any report rate

X9VoiD: How non-linear tho?

When reportMsAvg is 7.5ms, interpolation will stop when last report is 11.25ms
When reportMsAvg is 3.75ms, interpolation will stop when last report is 5.625ms

So in a sense this is linear yes but it'll work for most report rates (but not when reaching 500hz+ i guess). If you're thinking of another formula tho lmk, will plug that here instead

InfinityGhost: Exactly where this can become flimsy, we do need support for as many report rates as possible to avoid any future problems. Will think on some possible formulas to go off of, but I'm thinking either quadratic or root

X9VoiD: How about making it still linear but instead, we will scale according to these:

| Delta | Timeout |
| :--: | :--: |
| 1ms | 3ms |
| 7.5ms | 11.25ms |

So basically, when average delta is 1ms, timeout is 3ms, then that will scale linearly to 7.5ms : 11.25ms. We will clamp timeout to 3ms when report rate is lower than 1ms

InfinityGhost: Reasonable, lets do that then.

X9VoiD: Change pushed.

For this functionality, I used Vector2 functions to basically form a line. X is the base reportMs and Y is the limit.

InfinityGhost:
Revert the MacOS accelerated code and I'll go ahead and merge. We'll just add it in a later PR prior to release.

X9VoiD:
Reverted :bread:

OpenTabletDriver.Native/Linux/Timers/Enum.cs
```diff
@@ -0,0 +1,32 @@
...
+        None,
+        Thread,
+        ThreadID
+    }
+
+    public enum TimerFlag
+    {
+        Default,
+        AbsoluteTime
+    }
```
InfinityGhost, line 31: Separate into 3 files

OpenTabletDriver.Native/Linux/Timers/Structs/SigEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [FieldOffset(12)]
+        public SigEv notify;
+
+        [FieldOffset(16)]
+        public int tid;
+
+        [FieldOffset(16)]
+        public SigEvThread thread;
+    }
```
InfinityGhost, line 42: Move into separate files

OpenTabletDriver.Native/Linux/Timers/Structs/TimerSpec.cs
```diff
@@ -0,0 +1,18 @@
...
+        public long sec;
+        public long nsec;
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    public struct TimerSpec
+    {
+        public TimeSpec interval;
+        public TimeSpec value;
+    }
```
InfinityGhost, line 17: Move into separate files

OpenTabletDriver.Native/Linux/Timers/Structs/SigEvent.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace OpenTabletDriver.Native.Linux.Timers.Structs
+{
+    public delegate void TimerCallback(SigVal a);
```
InfinityGhost, line 6: Move to `OpenTabletDriver.Native/Linux/Timers/Timers.cs`

OpenTabletDriver.Native/OSX/OSX.cs
```diff
@@ -14,8 +14,10 @@ public static class OSX
     {
         private const string Quartz = "/System/Library/Frameworks/Quartz.framework/Versions/Current/Quartz";
         private const string Foundation = "/System/Library/Frameworks/Foundation.framework/Foundation";
-        private const string IOKit = "/System/Library/Frameworks/IOKit.framework/IOKit";
-        
+        private const string LibDispatch = "/usr/lib/system/libdispatch.dylib";
+
+        public delegate void TimerCallback();
+
```
InfinityGhost, line 20: Unused changes?

OpenTabletDriver.Plugin/IDriver.cs
```diff
@@ -7,8 +7,8 @@ namespace OpenTabletDriver.Plugin
 {
     public interface IDriver
     {
-        event EventHandler<bool> Reading;
-        event EventHandler<IDeviceReport> ReportRecieved;
+        event Action<bool> Reading;
+        event Action<IDeviceReport> ReportRecieved;
```
InfinityGhost, line 11: Invalid change, only `EventHander<T>` is supported in StreamJsonRpc

OpenTabletDriver.Plugin/ITimer.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace OpenTabletDriver.Plugin.Timers
+{
+    public interface ITimer : IDisposable
+    {
+        public void Start();
+        public void Stop();
+        public bool Stop(int milliseconds);
+        public bool Enabled { get; }
+        public float Interval { get; set; }
+        public event Action Elapsed;
```
InfinityGhost, line 12: `public` is unnecessary on interfaces.

OpenTabletDriver/DeviceReader.cs
```diff
@@ -26,20 +26,20 @@ public DeviceReader(HidDevice device, IReportParser<T> reportParser)
         public virtual HidDevice Device { protected set; get; }
         public virtual HidStream ReportStream { protected set; get; }
         public IReportParser<T> Parser { private set; get; }
-        public virtual event EventHandler<T> Report;
+        public virtual event Action<T> Report;
```
InfinityGhost, line 29: Keep `EventHandler<T>`, the `sender` parameter can prove to be useful in the future.

OpenTabletDriver/Interop/Timer/LinuxTimer.cs
```diff
@@ -0,0 +1,135 @@
...
+        public bool Enabled { private set; get; }
+
+        private double _interval;
+        public float Interval {
+            set
+            {
+                _interval = value;
+            }
+            get => (float)_interval;
+        }
```
InfinityGhost, line 131: Make this change
```diff
diff --git a/OpenTabletDriver/Interop/Timer/LinuxTimer.cs b/OpenTabletDriver/Interop/Timer/LinuxTimer.cs
index 6790d54..764c026 100644
--- a/OpenTabletDriver/Interop/Timer/LinuxTimer.cs
+++ b/OpenTabletDriver/Interop/Timer/LinuxTimer.cs
@@ -46,7 +46,7 @@ namespace OpenTabletDriver.Interop.Timer
                     Log.Write("LinuxTimer", $"Failed creating timer: {(ERRNO)Marshal.GetLastWin32Error()}", LogLevel.Error);
                 }
 
-                var interval = Interval * 1000 * 1000;
+                double interval = Interval * 1000 * 1000;
 
                 timeSpec = new TimerSpec
                 {
@@ -121,14 +121,7 @@ namespace OpenTabletDriver.Interop.Timer
 
         public bool Enabled { private set; get; }
 
-        private double _interval;
-        public float Interval {
-            set
-            {
-                _interval = value;
-            }
-            get => (float)_interval;
-        }
+        public float Interval { set; get; }
 
         public event Action Elapsed;
     }
```

OpenTabletDriver/Interop/Timer/WindowsTimer.cs
```diff
@@ -0,0 +1,69 @@
...
+            callbackDelegate = Callback;
+            callbackHandle = GCHandle.Alloc(callbackDelegate);
+        }
+
+        public void Start()
+        {
+            lock (stateLock)
+            {
+                var caps = new TimeCaps();
+                timeGetDevCaps(ref caps, (uint)Marshal.SizeOf(caps));
```
InfinityGhost, line 29: Why not `sizeof(TimeCaps)`?

OpenTabletDriver/Interop/Timer/WindowsTimer.cs
```diff
@@ -0,0 +1,69 @@
...
+                timeEndPeriod((uint)Interval);
+                Enabled = false;
+            }
+        }
+
+        public bool Stop(int milliseconds)
+        {
+            Stop();
+            return true; // waiting not implemented
+        }
```
InfinityGhost, line 50: Why is there a function like this? If its not implemented here why is it implemented anywhere else?

OpenTabletDriver.Plugin/Tablet/Interpolator/Limiter.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Numerics;
+
+namespace OpenTabletDriver.Plugin.Tablet.Interpolator
+{
+    internal static class Limiter
+    {
+        private static readonly Vector2 minimum = new Vector2(1, 3);
+        private static readonly Vector2 maximum = new Vector2(7.5f, 7.5f * 1.5f);
```
InfinityGhost, line 9: Why are these constantly defined? Just wondering the logic behind it.

OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    this.inRange = false;
+                }
+            }
+        }
+
+        public virtual void Dispose()
+        {
+            Enabled = false;
+        }
```
InfinityGhost, line 92: `ITimer` implements `IDisposable`, dispose here.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Tablet/Interpolator/Limiter.cs
X9VoiD: [Link to previous discussion](https://github.com/InfinityGhost/OpenTabletDriver/pull/302#discussion_r506708728)

X9VoiD:
Requested changes pushed :bread:

InfinityGhost (approved):

PR #307, opened 2020-09-19
Small patches

X9VoiD:
# Changes
* Handle bindings only once
* Initialize CTL-472

The bindings were already handled by [Driver.cs#L272](https://github.com/InfinityGhost/OpenTabletDriver/blob/master/OpenTabletDriver/Driver.cs#L272)

### DIFF
--- a/OpenTabletDriver.UX/DaemonWatchdog.cs
+++ b/OpenTabletDriver.UX/DaemonWatchdog.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Timers;
 
 namespace OpenTabletDriver.UX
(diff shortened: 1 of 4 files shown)


OpenTabletDriver.UX/DaemonWatchdog.cs
```diff
@@ -28,6 +29,12 @@ public class DaemonWatchdog: IDisposable
 
         public void Start()
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                daemonProcess.StartInfo.FileName = "OpenTabletDriver.Daemon.exe";
+                daemonProcess.StartInfo.Arguments = null;
+            }
+
```
InfinityGhost, line 37: We rely on the .NET runtime to launch the dll instead, I didn't intend on compiling the daemon as an executable itself in the releases.

InfinityGhost (changes requested):

InfinityGhost (approved):

PR #313, opened 2020-09-22
Feature initialize all IntuosV1

X9VoiD:
and also make the feature init consistent to AlternateDigitizerIdentifier


InfinityGhost:
Will test on my CTL-480, this *might* be a bad move considering most Wacom tablets already work by default.

InfinityGhost:
Appears to work as intended, so will be merging.

PR #314, opened 2020-09-22
Fix H430P ProductID

X9VoiD:


InfinityGhost:
We need to review tablet configuration changes more thoroughly in the future.

PR #318, opened 2020-09-26
Fix ProductID and ReportID of New 1060 Plus

X9VoiD:


InfinityGhost:
Lets make sure that other users of the New 1060 Plus don't have this same situation.

OpenTabletDriver/Configurations/Huion/New 1060 Plus.json
```diff
@@ -2,7 +2,7 @@
   "Name": "Huion New 1060 Plus",
   "DigitizerIdentifier": {
     "VendorID": 9580,
-    "ProductID": 109,
+    "ProductID": 110,
```
InfinityGhost, line 5: How about we make this change for the alternate digitizer?

InfinityGhost (changes requested):

InfinityGhost:
#350 might be related to this

X9VoiD:
The diagnostics from #350 confirmed that the product id is correct, but the input report length is different.
We could make an alternate identifier, one for InputReportLength 8 and one for InputReportLength 12

X9VoiD:
Oh... uhh. We can't make an alternate identifier for MaxPressure lol. We should really rework the TabletProperties now

InfinityGhost:
Update this to conform to latest tablet configuration changes

X9VoiD:
PR updated to new config format

InfinityGhost:
#350, what should we do about it?

X9VoiD:
Converted the PR to draft since we can't test the configurations on Windows (device stream seems to not open), and not verified yet.

X9VoiD:
Superseded by #562

PR #356, opened 2020-10-10
Rework TabletProperties -> TabletConfiguration

### LINKED ISSUE #358: Enumerate over all matching tablets
## Description
When more than one matching tablet endpoint is detected, OTD only considers to open the first matching tablet. This leads to issues when the first match can't be opened.

Instead, we should enumerate over all matching tablet endpoints until opening a stream succeeds.

Will fix issues with H420/420, HS64, and other tablets with similar endpoint characteristics.

InfinityGhost:
# Changes
- Configurations take multiple different identifiers

# Issues
- Closes #358


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -27,7 +27,11 @@ public DriverDaemon()
             Log.Output += (sender, message) => LogMessages.Add(message);
             Log.Output += (sender, message) => Console.WriteLine(Log.GetStringFormat(message));
             Log.Output += (sender, message) => Message?.Invoke(sender, message);
-            Driver.Reading += async (sender, isReading) => TabletChanged?.Invoke(this, isReading ? await GetTablet() : null);
+            Driver.Reading += async (sender, isReading) =>
+            {
+                TabletChanged?.Invoke(this, isReading ? await GetTablet() : null);
+            };
```
InfinityGhost, line 33: Unnecessary change

OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
```diff
@@ -33,8 +33,8 @@ public IEnumerable<IFilter> Filters
             get => _filters;
         }
         
-        private TabletProperties _tabletProperties;
-        public override TabletProperties TabletProperties
+        private DigitizerIdentifier _tabletProperties;
+        public override DigitizerIdentifier Tablet
```
InfinityGhost, line 37: Will rename this to Digitizer for uniformity and to avoid confusion

PR #360, opened 2020-10-12
Add Device String Reader window to UX

X9VoiD:


InfinityGhost:
Is this ready for merging?

X9VoiD:
Yes, this is ready for merge

InfinityGhost:
Appears to break after the first string is called.

InfinityGhost:
Now I can't reproduce that deadlock, I think it might be a good idea to add a timeout to avoid that.

X9VoiD:
Timeout added

InfinityGhost (approved):
Looks good.

PR #361, opened 2020-10-13
Update G640 v2 MaxX and MaxY

ohmaytt:
This is my first time opening a pull request, so please understand if I have done something weird or wrong.

Rev B version of XP-Pen G640 has MaxX value of 16000 and MaxY value of 1000, rather than 32000 and 2000,
Just tested with my Rev B, cursor movement was limited from top left to center with 32000/20000 value.
Worked well with 16000/10000 value.


InfinityGhost:
This is the maximum X & Y I get on mine, update it to match that instead. This little bit of a discrepancy can mean incorrect proportions.
image

PR #366, opened 2020-10-13
Provide access to VirtualScreen for plugins

X9VoiD:
This helps with VMulti-based plugins which needs to normalize values to VirtualScreen size

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -42,7 +42,7 @@ namespace OpenTabletDriver
 
         public DeviceIdentifier AuxiliaryIdentifier { private set; get; }
 
-        public IVirtualScreen VirtualScreen { get; } = Interop.Platform.VirtualScreen;
+        public IVirtualScreen VirtualScreen => Interop.Platform.VirtualScreen;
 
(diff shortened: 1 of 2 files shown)


skyborgff:
Great, this is better than having to create a "fake" AbsoluteOutputMode just to get that info

OpenTabletDriver/Driver.cs
```diff
@@ -41,6 +42,8 @@ private set
 
         public DeviceIdentifier AuxiliaryIdentifier { private set; get; }
 
+        public IVirtualScreen VirtualScreen { get; } = Interop.Platform.VirtualScreen;
```
InfinityGhost, line 45: The current state of this code will only set the virtual screen one time at construction. While this is sufficient now, we might support display size updates in the future.
```cs
public IVirtualScreen VirtualScreen => Interop.Platform.VirtualScreen;
```

InfinityGhost (changes requested):

OpenTabletDriver/Driver.cs
X9VoiD: Change pushed

InfinityGhost (approved):

PR #373, opened 2020-10-14
Revert RelativeMode sensitivity calculation

X9VoiD:
A change during the TabletConfiguration rewrite broke RelativeMode for all platforms. This PR fixes that regression.

### DIFF
--- a/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
@@ -32,19 +32,17 @@ namespace OpenTabletDriver.Plugin.Output
         public IVirtualPointer Pointer => VirtualMouse;
 
         private Vector2 _sensitivity;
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -38,6 +38,7 @@ public Vector2 Sensitivity
             set
             {
                 _sensitivity = value;
+                _reportScaleMultiplier = _sensitivity;
```
InfinityGhost, line 41: This is redundant, we could just use the value to set the scale multiplier or scale by the sensitivity itself.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
X9VoiD: Change pushed

PR #378, opened 2020-10-14
Fixed crash when trying to load plugins

skyborgff:
Example:
If the dlls from this file are in the plugins folder it would cause a crash
https://github.com/X9VoiD/VoiDPlugins/releases/tag/vmulti

### DIFF
--- a/OpenTabletDriver/PluginManager.cs
+++ b/OpenTabletDriver/PluginManager.cs
@@ -24,7 +24,8 @@ namespace OpenTabletDriver
             if (file.Extension == ".dll")
             {
                 var asm = ImportAssembly(file.FullName);
-                if (asm is null) { return false; }
+                if (asm is null) 
+                    return false;
 
                 foreach (var type in GetLoadableTypes(asm))
                 {
--- a/OpenTabletDriver/PluginManager.cs
+++ b/OpenTabletDriver/PluginManager.cs
@@ -51,7 +51,7 @@ namespace OpenTabletDriver
             {
                 return AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
             }
-            catch (BadImageFormatException e)
+            catch (BadImageFormatException)
             {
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/PluginManager.cs
```diff
@@ -45,7 +47,15 @@ public static bool TypeIsSupported(Type type)
 
         private static Assembly ImportAssembly(string path)
         {
-            return AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+            try
+            {
+                return AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+            }
+            catch (BadImageFormatException e)
```
InfinityGhost, line 54: Remove unused variable `e`

InfinityGhost (changes requested):

OpenTabletDriver/PluginManager.cs
```diff
@@ -24,6 +24,8 @@ public static bool AddPlugin(FileInfo file)
             if (file.Extension == ".dll")
             {
                 var asm = ImportAssembly(file.FullName);
+                if (asm is null) { return false; }
```
InfinityGhost, line 27: Formatting nitpick lol
```cs
if (asm is null)
    return false;
```

InfinityGhost (changes requested):

OpenTabletDriver/PluginManager.cs
skyborgff: tbh i wasnt sure the best way to make that. still think it's ugly

InfinityGhost: Agreed, a better way would to add null checking to `GetLoadableTypes()` but this is just a quick and easy way to do it

skyborgff: That seems inefficient, as we know that that dll `asm` is useless before entering that function.

InfinityGhost (approved):

PR #384, opened 2020-10-17
Add buttons for FAQ and Plugins

X9VoiD:
This makes it easier for people to reach places that are *not* a single click away

### DIFF
--- a/OpenTabletDriver.Native/SystemInfo.cs
+++ b/OpenTabletDriver.Native/SystemInfo.cs
@@ -42,5 +42,22 @@ namespace OpenTabletDriver.Native
                     break;
             }
         }
+
+        public static void OpenFileDirectory(string path)
+        {
+            switch (CurrentPlatform)
+            {
+                case RuntimePlatform.Windows:
+                    Process.Start("explorer.exe", path);
+                    break;
+                case RuntimePlatform.Linux:
+                    Process.Start("dbus-send", $"--session --dest=org.freedesktop.FileManager1 --type=method_call /org/freedesktop/FileManager1 org.freedesktop.FileManager1.ShowFolders array:string:\"file:{path}\" string:\"\"");
+                    break;
+                case RuntimePlatform.MacOS:
+                case RuntimePlatform.FreeBSD:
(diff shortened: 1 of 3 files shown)


OpenTabletDriver.Native/SystemInfo.cs
```diff
@@ -42,5 +42,22 @@ public static void Open(string path)
...
+                    Process.Start("explorer.exe", path);
+                    break;
+                case RuntimePlatform.Linux:
+                    Process.Start("dbus-send", $"--session --dest=org.freedesktop.FileManager1 --type=method_call /org/freedesktop/FileManager1 org.freedesktop.FileManager1.ShowFolders array:string:\"file:{path}\" string:\"\"");
+                    break;
+                case RuntimePlatform.MacOS:
+                case RuntimePlatform.FreeBSD:
+                    Process.Start("open", $"{path}");
+                    break;
+            }
```
InfinityGhost, line 60: This shouldn't be required, `Open(string path)` should handle MIME types such as directories and files

InfinityGhost (changes requested):

OpenTabletDriver.Native/SystemInfo.cs
X9VoiD: Yeah, opening folders with `Open()` also works on my Windows, but on my Linux it opens up as a VSCode workspace, so I directly used dbus to open a folder using the default file manager

InfinityGhost: Your MIME types are misconfigured then, it should use your file manager. `xdg-open` for a directory path opens in `nemo` for me like it is supposed to.

InfinityGhost: Relying on dbus to do this causes problems with i3 and other minimal window managers as well.

X9VoiD: Change pushed

InfinityGhost:
lg2m

InfinityGhost (approved):

PR #386, opened 2020-10-18
Improve reflection type caching

InfinityGhost:
# Changes
- Only cache a type when it inherits a type from `OpenTabletDriver.Plugin`
  > This should reduce memory usage and improve performance of reflection


InfinityGhost:
Broken with types with generic arguments `ISomeInterface<T>` which are implemented
This for example fails `IsAssignableFrom()` because it implements `T`, unsure of a workaround. 
```cs
OpenTabletDriver.Vendors.XP_Pen.XP_PenReportParser : IReportParser<IDeviceReport>
```

PR #388, opened 2020-10-18
Fix OutputReportLength of VEIKK S640

X9VoiD:
Not sure how I missed that before on the first PR for VEIKKs


InfinityGhost:
Is this really necessary?

X9VoiD:
Yes, this is so that OTD will send the OutputInitReport to the correct endpoint

InfinityGhost:
We'll update output init report lengths where it is actually needed, how about that?

X9VoiD:
Updated the PR to only change S640 where the change is needed.

InfinityGhost:
lg2m now

PR #399, opened 2020-10-24
Improve plugin logs

skyborgff:


InfinityGhost:
This may double up when the GUI reconnects, lets confirm that doesn't happen first.

skyborgff:
image
closing and restarting GUI didn't doubled it up

Issue #404, opened 2020-10-28
Support for Genius i608x tablet


Kijekin:
## Description
I'd like to see a support for Genius i608x tablet here, or a sort of template for me to do it myself.

X9VoiD:
Uninstall all tablet drivers associated with the device, restart, then OTD -> Help -> Export Diagnostics, upload the file here.

X9VoiD:
Any news on this?

Kijekin:
Was away from home, gimme sec

Kijekin:
Is it correct?

[diagnostics.zip](https://github.com/InfinityGhost/OpenTabletDriver/files/5487276/diagnostics.zip)

X9VoiD:
[MousePen i608X.zip](https://github.com/InfinityGhost/OpenTabletDriver/files/5488659/MousePen.i608X.zip)

Extract zip contents to `Configurations\Genius` then restart OTD

X9VoiD:
Please do note that the tablet dimensions and max sensor values might not be accurate and are only there as dummy values to let OTD detect the tablet. Those values will be filled in later when OTD Tablet Debugger have detected it and starts receiving reports.

Kijekin:
It doesn's seem to detect it. Shall i do any other steps in that case?

X9VoiD:
Make sure that you've restarted OTD and no other configs reference the same i608X.

Then OTD -> Help -> Export Diagnostics

Kijekin:
Done:

[diagnostics2.zip](https://github.com/InfinityGhost/OpenTabletDriver/files/5489500/diagnostics2.zip)

X9VoiD:
Delete your current OTD, then download and extract [latest dev build](https://github.com/InfinityGhost/OpenTabletDriver/actions/runs/342954837) to any folder except admin requiring folders. Add the configuration back then try again.

Kijekin:
Ok, it detects the tablet now

X9VoiD:
Check Tablet Debugger and see if OTD actually opened the correct `Endpoint` by seeing if `<Position>` is correct

X9VoiD:
Most tablets have upper left as 0,0 and bottom right as the sensor's maximum value, around `10000 px/mm` on 2540 LPI

Kijekin:
Tablet debugger shows only four blank boxes

Kijekin:
Can it be connected to fact that system reads tablet as multiple instances instead of one?

[...]

PR #425, opened 2020-11-05
Match device to config attributes to improve device selection

X9VoiD:
## Changes

- Add new Windows-specific `Attributes`.
  - `WinInterface` to match interface using the HID path.
  - `WinUsage` to match HID usage using the HID path.
- Future proof possible additional attributes by matching differently on Aux and Digitizer.

After fiddling around with the initial `WinDeviceMatchesEndpoint()` idea, I think we need to put `Attributes` inside `DigitizerIdentifiers` and `AuxIdentifiers`. This attribute change will be useful for when Windows somehow decides to separate touch, sliders, wheels, stuff, and to clean up the code this PR currently have.

## Issues

Helps #404

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using HidSharp;
 using OpenTabletDriver.Devices;
+using OpenTabletDriver.Interop;
 using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Output;
 using OpenTabletDriver.Plugin.Platform.Display;
@@ -91,7 +93,7 @@ namespace OpenTabletDriver
             digitizerIdentifier = default;
             foreach (var identifier in config.DigitizerIdentifiers)
             {
-                var matches = FindMatches(identifier);
+                var matches = FindMatchingDigitizer(identifier, config.Attributes);
 
                 if (matches.Count() > 1)
                     Log.Write("Detect", "More than 1 matching digitizer has been found.", LogLevel.Warning);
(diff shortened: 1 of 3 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -192,17 +192,62 @@ internal void InitializeAuxDevice(HidDevice auxDevice, DeviceIdentifier identifi
...
+                        conditionsMet = conditionsMet && Regex.IsMatch(devName, $"&col{attributes["WinUsage"]}");
+                    break;
+                }
+                case RuntimePlatform.Linux:
+                    break;
+                case RuntimePlatform.MacOS:
+                    break;
+                case RuntimePlatform.FreeBSD:
+                    break;
+            }
```
InfinityGhost, line 241: Clean up to this, or something similar. Worst case scenario, replace the breaks with `default` and return true.
```cs
private bool DigitizerMatchesAttribute(HidDevice device, Dictionary<string, string> attributes)
{
    var devName = device.GetFileSystemName();

    bool interfaceMatches = attributes.ContainsKey("WinInterface") ? Regex.IsMatch(devName, $"&mi_{attributes["WinInterface"]}") : true;
    bool keyMatches = attributes.ContainsKey("WinUsage") ? Regex.IsMatch(devName, $"&col{attributes["WinUsage"]}") : true;

    return SystemInfo.CurrentPlatform switch 
    {
        RuntimePlatform.Windows => interfaceMatches && keyMatches,
        _ => true
    };
}
```

OpenTabletDriver/Driver.cs
```diff
@@ -192,17 +192,62 @@ internal void InitializeAuxDevice(HidDevice auxDevice, DeviceIdentifier identifi
...
+
+            return conditionsMet;
+        }
+
+        private bool AuxMatchesAttribute(HidDevice device, Dictionary<string, string> attributes)
+        {
+            return true; // Future proofing
```
InfinityGhost, line 248: Might as well implement this ahead of time, but realistically this function shouldn't need to exist because it takes the same arguments as `DigitizerMatchesAttribute()` already covers it and should be used as a generic case.

InfinityGhost (changes requested):

X9VoiD:
Requested change pushed, but can't remove the similar functions since we have to differentiate between digitizer and aux. The aux don't need the WinUsage and WinInterface *yet*. When that time comes we will have to move configuration `Attributes` to inside `DigitizerIdentifier` and `AuxiliaryIdentifier`

InfinityGhost (approved):

PR #420, opened 2020-11-03
Rework plugin loading

X9VoiD:
## Changes

- Rewritten loading logic from scratch.
- Allow plugins (contained in a subdirectory) to load unmanaged dll dependency of their dependencies.
- Unified loading logic for Daemon and UX.
- Lock-less parallel loading wherever possible.

## Tested

- [x] Plugin loading
- [x] Unmanaged dll loading

Issue #182, opened 2020-07-12
[Feature Request] Add Plugin Manager


InfinityGhost:
## Description

Adding a plugin manager will decrease the need for background knowledge of how OpenTabletDriver works.

### General Functions
- Management
  - Install plugins
  - Uninstall plugins
  - Deactivate plugins
- Maintainence
  - Plugin settings
  - Update plugins (?)

PR #429, opened 2020-11-07
[WIP] Plugin Manager

X9VoiD:
# Changes

- [x] Plugin Manager window
  - [x] Show pending states
- [x] Drag and drop functionality (on supported platforms, see [Eto #1812](https://github.com/picoe/Eto/issues/1812))
- [x] Update UX of the plugin changes
- [x] Manual `File -> Install plugin...`

Depends on #420

Helps #182


X9VoiD:
[(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧](https://streamable.com/q0rm9f#)

InfinityGhost:
Still more to do, a method to uninstall plugins would also be a nice touch. I'll probably rework the list to have those additional buttons and handle them appropriately.

X9VoiD:
Uninstalling will be hard since plugins are open files when they are loaded. Windows can't delete them during runtime. Can probably do it during a daemon restart but that would mean Linux will be left behind as it can't restart Daemon (since we rely on the user on how to start the daemon as "daemon"). Same thing with updating loaded plugins, we can only do that with daemon restart.

X9VoiD:
Will merge `plugin-rework` changes after it is merged, to avoid unneeded extra merge commits.

InfinityGhost:
I'm fine with this in its current state, the other features will be added with the addition of a plugin repository.

OpenTabletDriver.UX/MainForm.cs
```diff
@@ -19,6 +19,8 @@ namespace OpenTabletDriver.UX
 
     public partial class MainForm : Form
     {
+        public static MainForm FormInstance { get; private set; }
```
InfinityGhost, line 22: Use `Application.Instance.MainForm` instead and cast it to `MainForm` where needed.

InfinityGhost (changes requested):

OpenTabletDriver.UX/MainForm.cs
X9VoiD: Pushed

OpenTabletDriver.Desktop/Reflection/PluginStateManager.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+using OpenTabletDriver.Plugin;
+
+namespace OpenTabletDriver.Desktop.Reflection
+{
+    public class PluginStateManager
+    {
+        public FileInfo UninstallCommands = new FileInfo(Path.Join(AppInfo.Current.PendingPluginDirectory, "uninstallCommands.txt"));
```
InfinityGhost, line 11: This is a bad way to go about this, is this the only option for enqueuing deletion/update?

InfinityGhost (changes requested):

OpenTabletDriver.Desktop/Reflection/PluginStateManager.cs
X9VoiD: Yeah, the only reasonable way we can do deletion and update is by doing it before `PluginManager` loads the plugins.

The global mutex is there to ensure that whichever runs first (daemon or ux), `PluginStateManager` will complete (and cleanup all enqueued stuff) before the current and other `PluginManager` loads plugins.

We may explore other methods to enqueue deletion, but the update procedure is the simplest it could be. Atleast that's what I think with what I currently know.

InfinityGhost: I think the best method we could have is to move any to-be-deleted files to a trash directory. From there, on next startup that trash directory could be cleared, and wouldn't be any need to check if the file is queued to be deleted. Not only that, but it could be used for client updates as well.

InfinityGhost (approved):

PR #457, opened 2020-11-19
Fix G430 name and ActiveReportID range

X9VoiD:
There were multiple reports saying that their G430 is detected as G430S, so I did a bit of review and found out that G430 and G430S have the same identifiers. So we will assume that these identifiers that came from hawku is not verified.

G430S is renamed to G430.
G430S_B is renamed to G430S.

And ActiveReportID is updated to use XP-Pen's usual [80..96) range.


InfinityGhost:
I want some verification on this

X9VoiD:
Will probably need the "G430S_B" tablet to verify that, but the G430S -> G430 rename already happened 4 times in discord so that part is verified.

SebbyLaw (approved):

PR #462, opened 2020-11-20
Implement rotation in RelativeMode, move deltaing to the end of a pipeline

AbstractQbit:
Rotation is done via a transform matrix, along with a sensitivity scaling.
Moving delta creation to the end allows filters to work with position instead of velocity, which they were not designed to handle.

### DIFF
--- a/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
@@ -56,6 +56,7 @@ namespace OpenTabletDriver.Plugin.Output
 
         private void UpdateTransformMatrix()
         {
+            _lastReceived = default;  // Prevents cursor from jumping on sensitivity change
             _transformationMatrix = Matrix3x2.CreateRotation(
                 (float)(_rotation * System.Math.PI / 180));
 
@@ -106,7 +107,7 @@ namespace OpenTabletDriver.Plugin.Output
             var delta = pos - _lastPos;
             _lastPos = pos;
 
-            return (difference > ResetTime && _lastReceived != default) ? null : delta;
+            return (difference > ResetTime) ? null : delta;
         }
     }
 }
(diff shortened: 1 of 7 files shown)


X9VoiD:
Am not against the change, but just wanna say that this breaks [PrecisionControl](https://github.com/X9VoiD/VoiDPlugins/blob/master/PrecisionControl/PrecisionControl.cs#L50).

Will clean up the code tho since we don't have to check if it's RelativeOutputMode anymore. Will also fix all other filters since only PrecisionControl handles relative data lol

AbstractQbit:
The idea was that you won't have to fix all the other filters, since now they should work the same as in absolute mode

OpenTabletDriver/Settings.cs
```diff
@@ -192,6 +192,13 @@ public float YSensitivity
             get => _yS;
         }
 
+        [JsonProperty("RelRotation")]
+        public float RelRotation
+        {
+            set => this.RaiseAndSetIfChanged(ref _relRot, value);
+            get => _relRot;
+        }
```
InfinityGhost, line 200: Rename to `RelativeRotation`, avoid shortened names for major properties like this.

InfinityGhost (changes requested):
Looks good, just one rename and I'm happy with it

InfinityGhost (approved):

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -32,24 +32,42 @@ public IEnumerable<IFilter> Filters
...
+                UpdateTransformMatrix();
+            }
+            get { return _rotation; }
+        }
+
+        private void UpdateTransformMatrix()
+        {
+            _lastReceived = default;  // Prevents cursor from jumping on sensitivity change
+            _transformationMatrix = Matrix3x2.CreateRotation(
+                (float)(_rotation * System.Math.PI / 180));
```
InfinityGhost, line 61: Invert rotation to match AbsoluteOutputMode

InfinityGhost (changes requested):

InfinityGhost (approved):

PR #464, opened 2020-11-21
LogView Rework

X9VoiD:
- Make LogView thread-safe while remaining lock-less.
- Internally synchronize all calls to AddItem to avoid race conditions.

### DIFF
--- a/OpenTabletDriver.UX/Controls/LogView.cs
+++ b/OpenTabletDriver.UX/Controls/LogView.cs
@@ -11,6 +11,7 @@ namespace OpenTabletDriver.UX.Controls
 {
     public class LogView : StackLayout
     {
+        private object logLock = new object();
         public LogView()
         {
             this.Orientation = Orientation.Vertical;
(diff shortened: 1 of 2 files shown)


InfinityGhost (approved):

OpenTabletDriver.UX/Controls/LogView.cs
```diff
@@ -162,26 +164,28 @@ private void Update(int startIndex, int endIndex)
 
         private void AddItem(LogMessage message)
         {
-            Application.Instance.AsyncInvoke(() => 
+            Application.Instance.AsyncInvoke(() =>
             {
-                Messages.Add(message);
+                lock (logLock)
```
InfinityGhost, line 169: After some testing, locking in this way deadlocks the UX thread. A concurrent collection should be used instead to avoid this issue.

InfinityGhost (changes requested):

OpenTabletDriver.UX/Controls/LogView.cs
X9VoiD: Requested changes pushed.

image

Not only did it lower memory usage from 3.1GB to 100MB, but the test itself run from 2-3 minutes down to just 2 seconds.

InfinityGhost (approved):
Absolutely fantastic changes here. Have yet to test it, but this looks like it should resolve the thread blocking being that it uses concurrent collections.

Issue #407, opened 2020-10-29
Support for Wacom CTE-460 Tablet


cl20103B11:
I only have this very old tablet :c

skyborgff:
Open OTD
Help -> export diagnostics and upload here

cl20103B11:
[diagnostics.txt](https://github.com/InfinityGhost/OpenTabletDriver/files/5458449/diagnostics.txt)

skyborgff:
tablet -> open configuration editor -> generate new configuration -> select your tablet. Do tell if it works btw

X9VoiD:
OTD in this case won't be able to properly generate a correct configuration since it had no access to report length information. Run both OTD Daemon and UX as sudo, and then export diagnostics again.

cl20103B11:
[test.txt](https://github.com/InfinityGhost/OpenTabletDriver/files/5463383/test.txt)

"Running as root and export diagnostics"

X9VoiD:
Use latest [dev build](https://github.com/InfinityGhost/OpenTabletDriver/actions/runs/336984458) (sign in to access download links) then download the zip below
> [CTE-460.zip](https://github.com/InfinityGhost/OpenTabletDriver/files/5465148/CTE-460.zip)

Extract content to `Configuration\Wacom`, please do note that the config contains dummy values for tablet dimensions and is only for the purpose of getting the tablet detected in OTD. The dimensions and specs will be updated later.

> [30-opentabletdriver.zip](https://github.com/InfinityGhost/OpenTabletDriver/files/5465194/30-opentabletdriver.zip)

Extract content to `/etc/udev/rules.d/`, then run `sudo udevadm control --reload-rules && sudo udevadm trigger`

X9VoiD:
Any news on this?

InfinityGhost:
@cl20103B11

cl20103B11:
Sorry for the late cuz too busy :c , It works, thx

X9VoiD:
Now that it's working, we should confirm the specs of the tablet.

Open OTD -> Tablet -> Tablet Debugger, then hover your pen slowly to bottom right most part of the tablet. Try to get the highest possible `Position` you can get.

And then check what `ReportID` it says when hovering pen, and when pen is out of range.

Lastly, check the maximum `Pressure` of the pen.

X9VoiD:
@cl20103B11 any news on the three values?

cl20103B11:
[...]

PR #471, opened 2020-11-23
Add config for CTE-460

X9VoiD:
## Verification

#407

Tablet dimension and sensor specs assumed correct


InfinityGhost:
Will mark as untested when merged

Issue #101, opened 2020-03-13
Incorrect report ID on CTH-690


InfinityGhost:
## Description

Incorrect report ID, resting is `29`

## System Information:

Option            | Value
------------------|------------------
Tablet            | CTH-690

PR #474, opened 2020-11-24
Fix CTL-6100 configuration 

InfinityGhost:
# Changes
- Configuration was broken, forcing the cursor to lock in the top center of the selected output area when inactive

# Verification
- PM with `Sam~#0101` on Discord

I want a secondary verification if at all possible.


InfinityGhost:
Based on how screwed this configuration was beforehand, I think we aught to consider this verified well enough.

PR #476, opened 2020-11-25
Interpolator patch

X9VoiD:
## Changes

- Do not set `interpolator.Enabled` from inside DriverDaemon.cs. This causes an unused running timer to be leaked.
- Remove delay of starting timers.
- Explicitly implement `IDisposable`.
- A bit of code cleanup and memory leak fix.

### DIFF
--- a/OpenTabletDriver/Interop/Timer/FallbackTimer.cs
+++ b/OpenTabletDriver/Interop/Timer/FallbackTimer.cs
@@ -57,7 +57,7 @@ namespace OpenTabletDriver.Interop.Timer
             this.runTimer = false;
         }
 
-        void ThreadMain()
+        private void ThreadMain()
         {
             float nextNotification = 0;
             float elapsedMilliseconds;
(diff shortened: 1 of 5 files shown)


OpenTabletDriver/Interop/Timer/FallbackTimer.cs
```diff
@@ -57,7 +57,7 @@ public void Stop()
             this.runTimer = false;
         }
 
-        void NotificationTimer()
+        void ThreadMain()
```
InfinityGhost, line 60:
```cs
private void ThreadMain()
```

InfinityGhost (changes requested):

OpenTabletDriver/Interop/Timer/FallbackTimer.cs
X9VoiD: Requested change pushed.

InfinityGhost (approved):

PR #481, opened 2020-11-27
Fix DeviceReader not notifying Driver of failure to open stream

X9VoiD:
[Driver.cs#L89](https://github.com/InfinityGhost/OpenTabletDriver/blob/1a8da16b03a13765301567e248726929fc86c08f/OpenTabletDriver/Driver.cs#L89) does not work since the exception that we want to detect gets captured inside DeviceReader instead. This results to the loop breaking too early without looping to all matching devices until it succeeds.

## Other changes

- Remove unused code, and a bit of cleanup.
- Increase DeviceReader thread priority to AboveNormal to improve performance consistency. Input processing happens in this specific thread.


OpenTabletDriver/DeviceReader.cs
```diff
@@ -17,16 +17,24 @@ public DeviceReader(HidDevice device, IReportParser<T> reportParser)
             {
                 Name = "OpenTabletDriver Device Reader",
                 IsBackground = true,
-                Priority = ThreadPriority.BelowNormal,
+                Priority = ThreadPriority.AboveNormal,
```
InfinityGhost, line 20: I think we previously discussed about removing this argument entirely, since its just used within HIDSharp for multiple opens.

X9VoiD: It was the `OpenPriority` thing that I removed in this PR. So we're good on that. This priority is for the actual thread that the OS will launch.

InfinityGhost: Ah gotcha, I was looking at the wrong thing lmao

InfinityGhost (approved):

PR #488, opened 2020-11-29
Move all OpenTabletDriver desktop-oriented classes to OpenTabletDriver.Desktop

InfinityGhost:
# Changes
- Moves all classes related to the OpenTabletDriver desktop experience to its own library
- Bindings are only used in the OpenTabletDriver.Desktop implementation
- Many naming changes for clarity

| Old                | New                |
| ------------------ | ------------------ |
| `IVirtualTablet`   | `IAbsolutePointer` |
| `IVirtualMouse`    | `IRelativePointer` |
| `IVirtualPointer`  | `IVirtualMouse`    |
| `IPressureHandler` | `IVirtualTablet`   |


OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -0,0 +1,19 @@
...
+using OpenTabletDriver.Plugin.Output;
+using OpenTabletDriver.Plugin.Platform.Display;
+using OpenTabletDriver.Plugin.Platform.Pointer;
+
+namespace OpenTabletDriver.Desktop.Output
+{
+    [PluginName("Artist Mode"), SupportedPlatform(PluginPlatform.Linux)]
+    public class LinuxArtistMode : AbsoluteOutputMode, IPointerOutputMode<IAbsolutePointer>
+    {
+        private Lazy<EvdevVirtualTablet> penHandler = new Lazy<EvdevVirtualTablet>();
```
X9VoiD, line 14: Since this is not a static class we can lose `Lazy<>` here
```cs
private EvDevVirtualTablet penHandler = new EvDevVirtualTablet()
```

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        public override IVirtualScreen VirtualScreen => throw new NotImplementedException();
```
X9VoiD, line 16: Is this intended? or
```cs
public override IVirtualScreen VirtualScreen => Platform.VirtualScreen;
```

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -0,0 +1,19 @@
...
+        public override IAbsolutePointer Pointer => penHandler.Value;
```
X9VoiD, line 17:
```cs
public override IAbsolutePointer Pointer => penHandler;
```

X9VoiD (changes requested):

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
InfinityGhost: Good catch

OpenTabletDriver.Plugin/Tablet/TabletStatus.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace OpenTabletDriver.Plugin.Tablet
+{
+    public class TabletState
```
InfinityGhost, line 3: Rename `TabletStatus.cs` -> `TabletState.cs`

OpenTabletDriver.UX/Windows/ConfigurationEditor.cs
```diff
@@ -141,10 +143,33 @@ private TabletConfiguration SelectedConfiguration
...
+        private void SerializeConfiguration(FileInfo file, TabletConfiguration configuration)
+        {
+            if (file.Exists)
+                file.Delete();
+            
+            using (var sw = file.CreateText())
+            using (var jw = new JsonTextWriter(sw))
+                ConfigurationSerializer.Serialize(jw, configuration);
+        }
+
```
InfinityGhost, line 168: We need a serialization helper class instead of this jank mess

OpenTabletDriver/Reflection/PluginManager.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                    var constructor = matchingConstructors.FirstOrDefault();
+                    return (T)constructor?.Invoke(args) ?? null;
+                }
+                catch
+                {
+                    Log.Write("Plugin", $"Unable to construct object '{name}'", LogLevel.Error);
+                }
+            }
+            return null;
```
InfinityGhost, line 60: Probably would be better to throw an exception instead of returning null, could reduce some pains.

InfinityGhost: Will change this later in a different PR where needed. Too much depends on this returning null, so we just will leave it be for now.

PR #491, opened 2020-11-29
Fix Intuosv3 aux buttons

AbstractQbit:
Fixed parsed index of IntuosV3 devices


InfinityGhost:
IIRC this issue is only present when using the Wacom driver, correct?

AbstractQbit:
No, it tries to read buttons from the same byte that report parser uses for switching, that happens regardless of skip byte

InfinityGhost:
Gotcha, so good to merge now?

AbstractQbit:
Yeah, will need to move to having dynamic button count eventually, but for now this should work

InfinityGhost (approved):

PR #502, opened 2020-12-02
Add FilterStage.PreInterpolate

X9VoiD:
This is required for filters that need to work on the rawest form of the tablet data.

An example of this is the [Reconstructor](https://github.com/X9VoiD/VoiDPlugins/tree/master/Reconstructor) filter where changing the report rate or the position by any means will affect its accuracy negatively.


OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
```diff
@@ -18,6 +20,7 @@ public Interpolator(ITimer scheduler)
 
         protected double reportMsAvg = 5.0f;
         protected bool enabled;
+        protected List<IFilter> filters = new List<IFilter>();
```
InfinityGhost, line 23: Whats the sense in storing this as a list?

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
X9VoiD: Same reason as to why we keep other filters as list per stages. Enumerating over an IEnumerable with `foreach` calls the constructor of the filter, making any settings and filter internal state to be reset.

X9VoiD: May also probably just use `IFilter[]` instead of `List<IFilter>` if you prefer that.

InfinityGhost: Use `IList<IFilter>` in place of `List<IFilter>`, it would cover our use case. `T[]` derives `IList<T>`, its the bare minimum for what we need.

X9VoiD: Requested changes pushed

OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
```diff
@@ -57,6 +60,12 @@ public virtual bool Enabled
             get => this.enabled;
         }
 
+        public virtual IEnumerable<IFilter> Filters
+        {
+            set => this.filters = value.ToArray();
+            get => this.filters;
+        }
```
InfinityGhost, line 67: `IList<IFilter>`, I'd prefer to avoid internally casting or converting

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Tablet/Interpolator/Interpolator.cs
X9VoiD: a

InfinityGhost (approved):

PR #542, opened 2020-12-11
Add PreInterpolate filters as PreTranspose when no active interpolator is running

X9VoiD:


OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -21,7 +21,10 @@ public IList<IFilter> Filters
             set
             {
                 this.filters = value;
-                this.preFilters = Filters.Where(t => t.FilterStage == FilterStage.PreTranspose).ToList();
+                if (Info.Driver.InterpolatorActive)
+                    this.preFilters = Filters.Where(t => t.FilterStage == FilterStage.PreTranspose).ToList();
+                else
+                    this.preFilters = Filters.Where(t => t.FilterStage == FilterStage.PreTranspose || t.FilterStage == FilterStage.PreInterpolate).ToList();
```
InfinityGhost, line 27: Why not `HasFlag()`?

InfinityGhost: If FilterStage doesn't have the Flags attribute, add it.

X9VoiD: I tried looking into it and I think it won't help us with this since Flags allows us to set multiple enums at once but that's not what we want. I may change the code to use `HasFlag` tho if you prefer that over `==`, but won't add `Flags`. If we use `Flags` then it means we have to add additional code to ensure that a filter only gets added once to whatever the earliest stage the filter is set and the `if statement block` above will still be needed.

InfinityGhost: Let's just leave it for now, we'll address it later in a code cleanup prior to release.

InfinityGhost (approved):

PR #551, opened 2020-12-13
Use single-file publish for Windows CI and update dotnet to latest minor version

AbstractQbit:


InfinityGhost (changes requested):
Single file embedding breaks assembly file location paths, revert that commit.

InfinityGhost (approved):

PR #575, opened 2020-12-20
Use HPET-enabled Stopwatch class

AbstractQbit:
It replaces DateTime.Now in ms-critical sections,
as per .net documentation, DateTime.Now's resolution
can have an impact on sub 100ms time intervals.
https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=net-5.0#datetime-resolution

### DIFF
--- a/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
@@ -12,9 +12,16 @@ namespace OpenTabletDriver.Plugin.Output
     [PluginIgnore]
     public abstract class RelativeOutputMode : IOutputMode
     {
+        public RelativeOutputMode()
+        {
+            stopwatch.Start();
+        }
+        
         private IList<IFilter> filters, preFilters, postFilters;
         private Vector2? lastPos;
         private Stopwatch stopwatch = new Stopwatch();
+        private TimeSpan lastElapsed = default;
+        private bool skipReport = false;
         private Matrix3x2 transformationMatrix;
 
         public IList<IFilter> Filters
@@ -68,7 +75,7 @@ namespace OpenTabletDriver.Plugin.Output
 
         private void UpdateTransformMatrix()
         {
-            this.stopwatch.Reset();     // Prevents cursor from jumping on sensitivity change
+            this.skipReport = true;     // Prevents cursor from jumping on sensitivity change
 
             this.transformationMatrix = Matrix3x2.CreateRotation(
                 (float)(-Rotation * System.Math.PI / 180));
@@ -94,8 +101,9 @@ namespace OpenTabletDriver.Plugin.Output
 
         public Vector2? Transpose(ITabletReport report)
         {
-            var elapsed = stopwatch.Elapsed;
-            stopwatch.Restart();
+            var currElapsed = stopwatch.Elapsed;
+            var deltaTime = currElapsed - lastElapsed;
+            lastElapsed = currElapsed;
 
             var pos = report.Position;
 
@@ -112,7 +120,12 @@ namespace OpenTabletDriver.Plugin.Output
             var delta = pos - this.lastPos;
             this.lastPos = pos;
 
-            return (elapsed > ResetTime || elapsed.Ticks != 0) ? null : delta;
+            if (skipReport)
+            {
+                skipReport = false;
+                return null;
+            }
+            return (deltaTime > ResetTime) ? null : delta;
         }
     }
 }
--- a/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Numerics;
 using OpenTabletDriver.Plugin.Attributes;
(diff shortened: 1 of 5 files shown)


OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -11,9 +12,16 @@ namespace OpenTabletDriver.Plugin.Output
...
+        {
+            stopwatch.Start();
+        }
+        
         private IList<IFilter> filters, preFilters, postFilters;
         private Vector2? lastPos;
-        private DateTime lastReceived;
+        private Stopwatch stopwatch = new Stopwatch();
+        private TimeSpan lastElapsed = default;
+        private bool skipReport = false;
```
InfinityGhost, line 24: I think it might be a good idea to create a base class for classes that need HPET

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
AbstractQbit: Change pushed

InfinityGhost:
The stopwatch based timer is only more accurate than `DateTime.Now` when the hardware supports it, and having a constantly running stopwatch also can have performance penalties.

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.ishighresolution?view=net-5.0#remarks

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public TimeSpan Reset()
+        {
+            var delta = Stop();
+            start = end = default;
+            return delta;
+        }
+        public double ResetMs() => Reset().TotalMilliseconds;
+
+        private static Stopwatch internalWatch = Stopwatch.StartNew();
```
InfinityGhost, line 66: Not a good idea to mark this as static, since it will always be running, even when no HPET timers are needed.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
```diff
@@ -0,0 +1,71 @@
...
+                return delta;
+            }
+            else
+            {
+                var delta = end - start;
+                Start();
+                return delta;
+            }
+        }
+        public double RestartMs() => Restart().TotalMilliseconds;
```
InfinityGhost, line 45: Remove these unnecessary helper methods, since they're only referenced like twice ever

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using System.Diagnostics;
+
+namespace OpenTabletDriver.Plugin.Utils
+{
+    public class DeltaStopwatch
```
InfinityGhost, line 6: Change namespace to `OpenTabletDriver.Plugin.Timing` and class name to `HighPrecisionEventStopwatch`

InfinityGhost: Do we really need a wrapper for an already existing class? This seems kinda pointless.

InfinityGhost (changes requested):

AbstractQbit:
> The stopwatch based timer is only more accurate than `DateTime.Now` when the hardware supports it, and having a constantly running stopwatch also can have performance penalties.

HPET was introduced by intel in ~2005 and is included in basically every platform since. It starts runing at a system startup and only ever costs you anything when you query it. The wrapper is there so that we can "lap" the stopwatch with just 1 kernel call/HPET query instead of two when querying elapsed and then restarting. This also eliminates the possibility of missing some ticks between checking the time and restarting.

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
AbstractQbit: If it's never used, then it just means one HPET query at static init phase, nothing more. There is really no difference between having it static vs non-static, apart from having a couple more objects and not being able to query total app runtime in non-static case.

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
AbstractQbit: DeltaStopwach name was used because it allows for measuring laps without redundant queries, in contrast to regular Stopwatch. HighPrecisionEventStopwatch doesn't really reflect the difference between those classes, and is very long imo

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
AbstractQbit: If it looks out of place, then sure, will remove those. Made them because most of the time milliseconds are what's being used in the end.

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
InfinityGhost: Leave this as is then.

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
InfinityGhost: Just looks odd for it to be exposed like that the main method going unused

OpenTabletDriver.Plugin/Utils/DeltaStopwatch.cs
InfinityGhost: Class name length isn't an issue. Can we compromise on `HPETDeltaStopwatch` as the class name?

InfinityGhost (approved):

PR #581, opened 2020-12-23
Completely specify the buggy G430 revision

X9VoiD:
The G430 with 10160LPI spec is most likely specific to the certain revision that it was found

### DIFF
--- a/OpenTabletDriver/Configurations/XP-Pen/G430.json
+++ b/OpenTabletDriver/Configurations/XP-Pen/G430.json
@@ -21,7 +21,7 @@
       "FeatureInitReport": null,
       "OutputInitReport": "ArAC",
       "DeviceStrings": {
-        "4": "V2\\.0"
+        "4": "V2\\.0 20170423"
       },
       "InitializationStrings": []
     },


OpenTabletDriver/Configurations/XP-Pen/G430.json
```diff
@@ -21,7 +21,7 @@
       "FeatureInitReport": null,
       "OutputInitReport": "ArAC",
       "DeviceStrings": {
-        "4": "V2\\.0"
+        "4": "V2\\.0 20170423"
```
InfinityGhost, line 24: Should we instead use a regex for the date? I feel like we've seen this sort of thing before and this is just the firmware build date.

InfinityGhost (changes requested):

OpenTabletDriver/Configurations/XP-Pen/G430.json
X9VoiD: No, in this case it's that specific firmware that is tied to that identifier.

InfinityGhost (approved):

PR #582, opened 2020-12-24
Add support for additional tablet report data

AbstractQbit:
Eraser is basically only detected on intuosv3 right now, on intuosv2 toolID determines if it's an eraser, and for other tablets idk where the eraser bit is.


InfinityGhost:
Move tilt and eraser into their own interfaces, `ITiltReport` and `IEraserReport`

OpenTabletDriver/Tablet/TabletReport.cs
```diff
@@ -4,18 +4,24 @@
 
 namespace OpenTabletDriver.Tablet
 {
-    public struct TabletReport : ITabletReport
+    public struct TabletReport : ITabletReport, ITiltReport
```
InfinityGhost, line 7: Not all generic HID tablets support tilt from what I can tell, revert this to standard. Create a separate IntuosV1Report for the 470, 471, 480, etc. since they do contain this data.

InfinityGhost (changes requested):

InfinityGhost:
All that is really left for this to be complete is making it work and readable with the tablet debugger.

OpenTabletDriver/Tablet/TabletReport.cs
X9VoiD: UC Logic tablets uses this same logic so it may be named as `ExtendedTabletReport` (signifying that it is just an extension of the standard parser) or something . Or probably just check if report size is more than 10.

InfinityGhost: > Or probably just check if report size is more than 10.

Maybe in the report parser, but not directly within the report class. IntuosV1 also reports eraser, tilt, and hover height detection, so that's something we should also try to confirm if its used in UCLogic tablets as well.

AbstractQbit:
Redid this pr because changes here were incompatible with visualizer

OpenTabletDriver.Plugin/Tablet/IDeviceReport.cs
```diff
@@ -1,7 +1,11 @@
+using System;
+
 namespace OpenTabletDriver.Plugin.Tablet
 {
     public interface IDeviceReport
     {
         byte[] Raw { get; }
+        string GetStringRaw() => BitConverter.ToString(Raw).Replace('-', ' ');
+        string GetStringFormat() => "";
```
InfinityGhost, line 9: I like this change, however maybe we should just make it pull the raw report when its just an IDeviceReport.
```suggestion
        string GetStringFormat() => GetStringRaw();
```

OpenTabletDriver/Tablet/TabletReportParser.cs
```diff
@@ -6,7 +6,10 @@ public class TabletReportParser : IReportParser<IDeviceReport>
     {
         public virtual IDeviceReport Parse(byte[] data)
         {
-            return new TabletReport(data);
+            if (data.Length < 12)
+                return new TabletReport(data);
+            else
+                return new TiltTabletReport(data);
```
InfinityGhost, line 12: Might need to test this out, could potentially cause huge issues.

InfinityGhost: We should instead add another report parser, keeping this one vanilla without any frills. Add a `TabletTiltReportParser` class inheriting this class with these changes with an override of this method.

InfinityGhost (changes requested):

InfinityGhost (approved):
LG2M

PR #593, opened 2020-12-29
Add Configuration For Wacom Intuos 4 XL ( PTK-1240 )

insomnyawolf:
I've been using this for a while and seems to cause no issues yet so far, if you need anything else, tell me

### DIFF
--- a/OpenTabletDriver/Configurations/Wacom/PTK-1240.json
+++ b/OpenTabletDriver/Configurations/Wacom/PTK-1240.json
@@ -8,8 +8,8 @@
       "MaxY": 60960.0,
       "MaxPressure": 2047,
       "ActiveReportID": {
-        "Start": 0,
-        "StartInclusive": true,
+        "Start": 1,
+        "StartInclusive": false,
         "End": 62,
         "EndInclusive": false
       },
@@ -30,8 +30,8 @@
       "MaxY": 60960.0,
       "MaxPressure": 2047,
       "ActiveReportID": {
-        "Start": 0,
-        "StartInclusive": true,
+        "Start": 1,
+        "StartInclusive": false,
         "End": 62,
         "EndInclusive": false
       },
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/Configurations/Wacom/PTK-1240.json
```diff
@@ -0,0 +1,51 @@
...
+  "DigitizerIdentifiers": [
+    {
+      "Width": 487.68,
+      "Height": 304.8,
+      "MaxX": 97536.0,
+      "MaxY": 60960.0,
+      "MaxPressure": 2047,
+      "ActiveReportID": {
+        "Start": 0,
+        "StartInclusive": true,
```
X9VoiD, line 12: Other IntuosV2 tablets needed
```
"Start": 1,
"StartInclusive": false
```
to avoid occasional cursor teleport when tablet is not in use but still connected. Can you verify that the same issue is not happening with this tablet? If it's happening then change the range to the one I wrote here.

OpenTabletDriver/Configurations/Wacom/PTK-1240.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    {
+      "Width": 487.68,
+      "Height": 304.8,
+      "MaxX": 97536.0,
+      "MaxY": 60960.0,
+      "MaxPressure": 2047,
+      "ActiveReportID": {
+        "Start": 0,
+        "StartInclusive": true,
```
X9VoiD, line 34: Same here

X9VoiD (changes requested):

OpenTabletDriver/Configurations/Wacom/PTK-1240.json
insomnyawolf: I didn't had this issue yet, the only similar thing that happen is cursor teleport if i click the buttons on the tablet.

I'm testing it with this edit just in case, i'll keep you updated.

InfinityGhost: > the only similar thing that happen is cursor teleport if i click the buttons on the tablet

That might be due to this range issue.

insomnyawolf:
I've tested it a little bit and it seems to be working fine, thanks for the feedback

X9VoiD (approved):

InfinityGhost (approved):

PR #609, opened 2020-12-31
Allow only one instance of UX

### LINKED ISSUE #479: Bind OpenTabletDriver menu to button
Currently there is no way to open the driver menu without clicking the icon on task bar. Please add option to bind it to keyboard button and tablet express keys. Thanks

X9VoiD:
## Changes

- Single instance UX
- `Info.GeUXInstance()`
- New binding `UXBinding : "Show Window"`

Code were structured for easy addition of new capabilities that may be invoked from `RpcClient<IUXDriver>` (requesting for better interface name).

Plugins may use that proxy object to invoke methods to run on UX thread.

## Issues

Fixes #479

### DIFF
--- a/OpenTabletDriver.UX/App.cs
+++ b/OpenTabletDriver.UX/App.cs
@@ -17,6 +17,13 @@ namespace OpenTabletDriver.UX
     {
         public static void Run(string platform, string[] args)
         {
+            _ = UXInstance.Invoke(async ux =>
+            {
+                await ux.ShowClient();
+                UXInstance.Dispose();
+                Environment.Exit(0);
+            });
+
             var root = new RootCommand("OpenTabletDriver UX")
             {
                 new Option<bool>(new string[] { "-m", "--minimized" }, "Start the application minimized")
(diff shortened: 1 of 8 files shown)


OpenTabletDriver.UX/App.cs
```diff
@@ -53,6 +55,38 @@ public static void Run(string platform, string[] args)
             app.Run(mainForm);
         }
 
+        private static void RequestShowClient()
+        {
+            var handle = new EventWaitHandle(false, EventResetMode.AutoReset, @"Global\OpenTabletDriver.UX.Wpf", out var host);
```
InfinityGhost, line 60: Should just use our `Instance` class here, since it does the same thing without requiring an extra thread.

InfinityGhost (changes requested):

OpenTabletDriver.UX/App.cs
X9VoiD: We can't use Instance here since it cannot detect when a new Instance is spawned. It can only work one way. And the thread handling the showing of the window runs on the host, not on the new one.

InfinityGhost: You just detect if an instance already exists, literally copy what the daemon does.

InfinityGhost: Scratch that, misread the second part. If it has issues in the future we'll deal with it then.

InfinityGhost (approved):

X9VoiD:
Closing. Superseded by #634

PR #612, opened 2021-01-01
Change Area Converter labels accordingly

X9VoiD:
Also changed Group to update layout when changing `Group.Text`, and to avoid performing unneeded layout update during ctor by using the private member instead of the public property during ctor.

### DIFF
--- a/OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
+++ b/OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
@@ -4,6 +4,8 @@ namespace OpenTabletDriver.Plugin.Tablet
     {
         DeviceVendor Vendor { get; }
 
+        string[] Label { get; }
+
         Area Convert(TabletState tablet, double left, double top, double right, double bottom);
     }
 }
--- a/OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
+++ b/OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
@@ -4,8 +4,11 @@ namespace OpenTabletDriver.Plugin.Tablet
     {
         DeviceVendor Vendor { get; }
 
-        string[] Label { get; }
+        string Top { get; }
+        string Left { get; }
+        string Bottom { get; }
+        string Right { get; }
 
-        Area Convert(TabletState tablet, double left, double top, double right, double bottom);
+        Area Convert(TabletState tablet, double top, double left, double bottom, double right);
     }
 }
--- a/OpenTabletDriver.UX/Controls/Generic/Group.cs
+++ b/OpenTabletDriver.UX/Controls/Generic/Group.cs
@@ -15,8 +15,8 @@ namespace OpenTabletDriver.UX.Controls.Generic
         public Group(string text, Control content, Orientation orientation = DEFAULT_ORIENTATION, bool expand = true)
             : this()
         {
-            this.text = text;
-            this.content = content;
+            this.Text = text;
+            this.Content = content;
             this.Orientation = orientation;
             this.ExpandContent = expand;
         }
@@ -57,6 +57,9 @@ namespace OpenTabletDriver.UX.Controls.Generic
 
         protected void UpdateControlLayout()
         {
+            if (!this.Loaded)
+                return;
+
             switch (Orientation, SystemInterop.CurrentPlatform)
             {
                 case (_, PluginPlatform.MacOS):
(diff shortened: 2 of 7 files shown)


OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
```diff
@@ -4,6 +4,8 @@ public interface IAreaConverter
     {
         DeviceVendor Vendor { get; }
 
+        string[] Label { get; }
```
InfinityGhost, line 7: Change this to make it more obvious what order they should be in
```cs
string Left { get; }
string Top { get; }
string Right { get; }
string Bottom { get; }
```

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Tablet/IAreaConverter.cs
X9VoiD: They go by the order of variables in `Convert()` since left, top, right, bottom don't quite fit XP-Pen's width, height, x-offset, y-offset

InfinityGhost: Doesn't matter, since here you can redefine it. XP-Pen is the odd one out on everything here right now

X9VoiD: Fixed

OpenTabletDriver.UX/Windows/AreaConverterDialog.cs
```diff
@@ -15,15 +16,42 @@ public AreaConverterDialog()
...
+                var group = new Group
+                {
+                    Orientation = Orientation.Horizontal,
+                    Content = numberBox
+                };
+
+                groups[i] = group;
+                numberBoxes[i] = numberBox;
+            }
+
```
InfinityGhost, line 54: This change is now overkill for what we need, since we're just using the properties. We can basically just revert to what we had before but with the label text being the property.

InfinityGhost (changes requested):

OpenTabletDriver.UX/Windows/AreaConverterDialog.cs
X9VoiD: We still have to keep a reference to the group somehow. I reverted those changes so it won't use arrays anymore, but kept the Group variable so we may refresh `Group.Text`.

OpenTabletDriver.UX/Controls/Generic/Group.cs
```diff
@@ -27,8 +27,17 @@ public Group(string text, Control content, Orientation orientation = DEFAULT_ORI
...
+        private string text;
+        public string Text
+        {
+            set
+            {
+                this.text = value;
+                UpdateControlLayout();
+            }
+            get => text;
+        }
```
InfinityGhost, line 40: Unnecessary change. Apply this diff instead:
```diff
diff --git a/OpenTabletDriver.UX/Controls/Generic/Group.cs b/OpenTabletDriver.UX/Controls/Generic/Group.cs
index eabc504..eb93847 100644
--- a/OpenTabletDriver.UX/Controls/Generic/Group.cs
+++ b/OpenTabletDriver.UX/Controls/Generic/Group.cs
@@ -57,6 +57,9 @@ namespace OpenTabletDriver.UX.Controls.Generic
 
         protected void UpdateControlLayout()
         {
+            if (!this.Loaded)
+                return;
+
             switch (Orientation, SystemInterop.CurrentPlatform)
             {
                 case (_, PluginPlatform.MacOS):
```

OpenTabletDriver.UX/Controls/OutputModeEditor.cs
```diff
@@ -276,7 +276,7 @@ public void Rebind(Settings settings)
                 private async Task ConvertAreaDialog()
                 {
                     var converter = new AreaConverterDialog();
-                    await converter.ShowModalAsync();
+                    await converter.ShowModalAsync(Application.Instance.MainForm);
```
InfinityGhost, line 279: `await converter.ShowModalAsync(this.ParentWindow);` should be fine

InfinityGhost (changes requested):

OpenTabletDriver.UX/Controls/Generic/Group.cs
X9VoiD: image

The control itself is not notified of the change to Text since it's not data binded I think.

InfinityGhost: Keep the `Text` change, but don't use the private field to update it's value in the constructor.

X9VoiD: 🍞

InfinityGhost (approved):

PR #638, opened 2021-01-10
Use latest plugin update and check driver compatibility

X9VoiD:
## Changes

- Check if plugin metadata is compatible to current driver
- If there is more than on matching metadata, sort it in descending order so `updatableFromRepository.First()` returns the latest version of that plugin
- Merge LINQ query

### DIFF
--- a/OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
+++ b/OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
@@ -216,6 +216,7 @@ namespace OpenTabletDriver.UX.Windows
             protected WeakReference<PluginMetadata> MetadataReference { set; get; } = new WeakReference<PluginMetadata>(null);
 
             private EmptyMetadataControl emptyMetadataControl = new EmptyMetadataControl();
+            private Version driverVersion = Assembly.GetExecutingAssembly().GetName().Version;
 
             public void Update(PluginMetadata metadata)
             {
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
```diff
@@ -239,13 +240,15 @@ public async void Refresh()
                     Repository ??= await PluginMetadataCollection.DownloadAsync();
 
                     bool isInstalled = contexts.Any(t => PluginMetadata.Match(t.GetMetadata(), metadata));
-                    bool canUpdate = Repository.Any(t => t.Name == metadata.Name && t.PluginVersion > metadata.PluginVersion);
 
                     var updatableFromRepository = from meta in Repository
-                        where meta.PluginVersion > metadata.PluginVersion
-                        orderby meta.PluginVersion
+                        where meta.Name == metadata.Name && meta.PluginVersion > metadata.PluginVersion
+                        where driverVersion >= meta.SupportedDriverVersion
```
InfinityGhost, line 246: Is `driverVersion` used in the installation case as well?

X9VoiD: As far as I can tell, this Update() function is also used for installation

InfinityGhost: lg2m then

InfinityGhost (approved):

PR #639, opened 2021-01-11
Wrong dimensions for G640S

25pwn:
https://www.xp-pen.com/product/54.html
The specs and the user manual say the active area is 6.5"*4", which is 165.1mm*101.6 instead of 165.0mm*103.0mm


X9VoiD:
We follow digital dimensions coming from the tablet descriptor itself, not the manufacturer's measurements in inches which is often *rounded*

PR #641, opened 2021-01-12
Added Huion H1161 configuration file

ficion:
Pretty much the same as every other Huion's.

### DIFF
--- a/OpenTabletDriver/Configurations/Huion/H1161.json
+++ b/OpenTabletDriver/Configurations/Huion/H1161.json
@@ -1,0 +1,35 @@
+{
+  "Name": "Huion H1161",
+  "DigitizerIdentifiers": [
+    {
+      "Width": 279.4,
+      "Height": 174.6,
+      "MaxX": 55880.0,
+      "MaxY": 34925.0,
+      "MaxPressure": 8191,
+      "ActiveReportID": {
+        "Start": 64,
+        "StartInclusive": true,
+        "End": 64,
+        "EndInclusive": true
+      },
+      "VendorID": 9580,
+      "ProductID": 109,
+      "InputReportLength": 12,
+      "OutputReportLength": null,
+      "ReportParser": "OpenTabletDriver.Tablet.TabletReportParser",
+      "FeatureInitReport": null,
+      "OutputInitReport": null,
+      "DeviceStrings": {
+        "201": "HUION_T191_\\d{6}$"
+      },
+      "InitializationStrings": [
+        200
+      ]
+    }
+  ],
+  "AuxilaryDeviceIdentifiers": [],
+  "Attributes": {
+    "libinputoverride": "1"
+  }
+}


OpenTabletDriver/Configurations/Huion/H1161.json
```diff
@@ -0,0 +1,35 @@
+{
+  "Name": "Huion H1161",
+  "DigitizerIdentifiers": [
+    {
+      "Width": 279.4,
+      "Height": 174.6,
+      "MaxX": 55880.0,
+      "MaxY": 34925.0,
```
X9VoiD, line 8: Can we get a verification for this MaxX and MaxY? Preferably from tablet debugger's output.

OpenTabletDriver/Configurations/Huion/H1161.json
```diff
@@ -0,0 +1,35 @@
...
+      "Width": 279.4,
+      "Height": 174.6,
+      "MaxX": 55880.0,
+      "MaxY": 34925.0,
+      "MaxPressure": 8191,
+      "ActiveReportID": {
+        "Start": 64,
+        "StartInclusive": true,
+        "End": 64,
+        "EndInclusive": true
```
X9VoiD, line 14: Is there some reason for deviating from the [64..null) range of other huion configs?

OpenTabletDriver/Configurations/Huion/H1161.json
ficion: image
image
That's as high as it gets on each border

OpenTabletDriver/Configurations/Huion/H1161.json
ficion: When buttons are pressed they send a 112 and 120 signal and that makes the cursor move to the top left. Since only cursor movement is supported for now, I thought it'd be fine to limit it to 64

X9VoiD (approved):

OpenTabletDriver/Configurations/Huion/H1161.json
Mrcubix: that's not something you can remove with reportID your tablet use UC logic which is known to send invalid report when pushing buttons causing the cursor to teleport, so by doing this, you might prevent your cursor from moving at all while buttons are pushed, stick to [64,null[

X9VoiD: No. Uhh.. @ficion's config is just fine as it is if we're not going to use the gaomon/huion parser yet. ReportID 64 is for tablet report, higher than that is an aux report. Check [GaomonReportParser.cs](https://github.com/InfinityGhost/OpenTabletDriver/blob/master/OpenTabletDriver/Vendors/Gaomon/GaomonReportParser.cs).

@ficion and I were talking about teleport due to reading aux report as tablet report. The click issue is a different thing and just offsets the cursor instead of teleporting to `<0, 0>`.

InfinityGhost (approved):

PR #657, opened 2021-01-17
Fix bugs and add MaxPenButtonCount

X9VoiD:
- [x] Add `MaxPenButtonCount` to tablet configurations

## Notes

`Driver.TabletChanged` was previously spamming nulls which locked UX on `bindingEditor.UpdateBindings()`, so this PR resolves that by triggering the event only when the `Driver.Tablet` actually changed. There were also some `STAThread` exceptions not found before.


X9VoiD:
Will close the PR and open a new dedicated PR for each bugs. The `MaxPenButtonCount` is expected to not be used this way so that part will be discarded.

PR #666, opened 2021-01-18
Import Timers from glibc instead of librt

X9VoiD:


InfinityGhost (changes requested):
```
[EntryPointNotFoundException:Error]	Unable to find an entry point named 'timer_create' in shared library 'libc'.
```

X9VoiD:
Closing for now.

PR #683, opened 2021-01-20
Fix race conditions of interpolators

X9VoiD:
## Changes

- Safe invoke `Elapsed`
- Dispose interpolators before performing changes on the input processing pipeline
- Ensure that timers/schedulers' `Start()` and `Stop()` are not redundantly called
- Fix race condition on `Interpolator.Dispose()`
- Fix `Interpolator.InRange` and `Interpolator.Enabled`. Do only as the variable states.
- Synchronously kill Windows timer

## Test

Before PR: Crash when applying settings with pen while an interpolator is active
After PR: Did not crash after spamming apply

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -141,6 +141,10 @@ namespace OpenTabletDriver.Daemon
 
         public Task SetSettings(Settings settings)
         {
+            foreach (var interpolator in Driver.Interpolators)
+                interpolator.Dispose();
+            Driver.Interpolators.Clear();
+
             Settings = SettingsMigrator.Migrate(settings);
 
             var pluginRef = Settings.OutputMode?.GetPluginReference() ?? AppInfo.PluginManager.GetPluginReference(typeof(AbsoluteMode));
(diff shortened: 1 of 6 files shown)


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -141,6 +141,10 @@ public async Task<TabletState> DetectTablets()
 
         public Task SetSettings(Settings settings)
         {
+            foreach (var interpolator in Driver.Interpolators)
+                interpolator.Dispose();
+            Driver.Interpolators.Clear();
```
InfinityGhost, line 146: Any reasoning for moving this out of the `SetInterpolatorSettings()` method? This is kind of hard to justify.

InfinityGhost (changes requested):
Haven't tested yet, but looks good.

OpenTabletDriver.Daemon/DriverDaemon.cs
InfinityGhost: Add this comment and I'll approve
```suggestion
            // Dispose all interpolators to begin changing settings
            foreach (var interpolator in Driver.Interpolators)
                interpolator.Dispose();
            Driver.Interpolators.Clear();
```

InfinityGhost:
@X9VoiD Resolve merge conflicts and I'll do a full review

InfinityGhost (approved):
Tested on Linux

PR #684, opened 2021-01-20
Invoke TabletChanged only when it actually changed

X9VoiD:
### Procedure

Replug tablet

### Before PR

`TabletChanged` is invoked once on disconnect, and then invoked for every configuration until a match is found.

### After PR

`TabletChanged` is invoked only once on both disconnect and reconnect.

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -47,11 +47,13 @@ namespace OpenTabletDriver
         {
             private set
             {
+                if (value != this.tablet)
+                    TabletChanged?.Invoke(this, value);
+
                 // Stored locally to avoid re-detecting to switch output modes
                 this.tablet = value;
                 if (OutputMode != null)
                     OutputMode.Tablet = Tablet;
-                TabletChanged?.Invoke(this, Tablet);
             }
             get => this.tablet;
         }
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -47,11 +47,13 @@ public TabletState Tablet
         {
             private set
             {
+                if (value != this.tablet)
+                    TabletChanged?.Invoke(this, value);
+
```
InfinityGhost, line 52: Invoking the event should happen after the value is set, just in case the consumer uses Driver.Tablet to get the property instead of the event's value.

InfinityGhost (changes requested):

OpenTabletDriver/Driver.cs
X9VoiD:
```cs
                if (value != this.tablet)
                {
                    // Stored locally to avoid re-detecting to switch output modes
                    this.tablet = value;
                    if (OutputMode != null)
                        OutputMode.Tablet = Tablet;
                    TabletChanged?.Invoke(this, value);
                }
```
Something like this? So it also only sets the `OutputMode.Tablet` when it changes. `OutputMode` perform some calculations when the tablet is set.

InfinityGhost: That would work fine, yes.

InfinityGhost (approved):

PR #689, opened 2021-01-20
Center relative to parent control and fix location jitter

X9VoiD:
## Changes

- Center all child form or dialog relative to parent window (mostly `MainForm`)
- Attach all child form or dialog to parent window (fixes problems like losing a form/dialog behind the `MainForm` or the form/dialog not restored in conjuction with `MainForm.Show()` after getting minimized)
- Fixed `MainForm` being larger than 1366x768 or smaller screens. (Previous logic checked if ClientSize is larger than Screen before applying but ClientSize does not account for window decoration, so the window's total size is still larger than the screen)
- Apply `MainForm`'s sizing and centering logic before it launches by retrieving the screen through the cursor


OpenTabletDriver.UX/MainForm.cs
```diff
@@ -18,13 +18,14 @@ namespace OpenTabletDriver.UX
 {
     using static App;
 
-    public partial class MainForm : DesktopForm
+    public partial class MainForm : Form
     {
         public MainForm()
             : base()
         {
+            InitializePlatform();
```
InfinityGhost, line 26: At least copy what was in the original `DesktopForm` class if you're going to outright remove it. This is called at `OnLoadComplete()` with the single run variable. Preferably, I'd like to keep the `DesktopForm` class so this is easier to maintain.

InfinityGhost (changes requested):

InfinityGhost (approved):

InfinityGhost (changes requested):
On GTK, dialogs are not even relatively close to being centered.
image

OpenTabletDriver.UX/DesktopForm.cs
```diff
@@ -1,66 +1,42 @@
...
+            var y = Owner.Location.Y + (Owner.ClientSize.Height / 2);
+            var center = new PointF(x, y);
 
-            if (!this.platformInit)
-            {
-                // Adjust to any platform quirks
-                OnInitializePlatform(e);
-            }
+            Location = new Point((int)(center.X - (ClientSize.Width / 2)), (int)(center.Y - (ClientSize.Height / 2)));
         }
```
InfinityGhost, line 30: Fix this to check if the platform is GTK, window positioning is handled automatically on that platform.
```cs
        protected virtual void InitializeForm()
        {
            if (!Platform.IsGtk)
            {
                var x = Owner.Location.X + (Owner.ClientSize.Width / 2);
                var y = Owner.Location.Y + (Owner.ClientSize.Height / 2);
                var center = new PointF(x, y);

                Location = new Point((int)(center.X - (ClientSize.Width / 2)), (int)(center.Y - (ClientSize.Height / 2)));
            }
        }
 ```

**Review by InfinityGhost — CHANGES_REQUESTED** [review]

**Comment by X9VoiD** [comment]
Fixed, and rebased to https://github.com/OpenTabletDriver/OpenTabletDriver/commit/2ddf0a75b455996ac3eba163a280c83c11b7ba4a.

image

It was a bug in Eto where `ClientSize` will always be zero until form is shown, all calls to `ClientSize._set` are ignored in Gtk as long as `Form.Show()` is not called yet

**Comment by gonX** [comment]
CodeFactor is crying - also, rebase and/or close? No idea if this PR is still interesting

**Comment by InfinityGhost** [comment]
Resolve conflicts and I'll test then merge.

**Comment by InfinityGhost** [comment]
Tablet debugger size appears entirely wrong on GTK, horizontal scrollbar visible.

File: OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs [hunk 3]
```diff
@@ -19,8 +19,10 @@ namespace OpenTabletDriver.UX.Windows.Tablet
     public class TabletDebugger : DesktopForm
     {
         public TabletDebugger()
+            : base(Application.Instance.MainForm)
         {
             Title = "Tablet Debugger";
+            this.ClientSize = new Size(640, 640);
```
gonX, line 25: Removing this line makes Tablet debugger size properly by default
```suggestion
```

gonX (changes requested):
Just 1 change to fix the issue reported by Ghost, otherwise LGTM

OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
X9VoiD: Removing this may also cause issues with platforms that rely on pre-determined size for jitter-less window creation (Windows, MacOS), but will add that change for now for the sake of moving things forward.

gonX: The tablet debugger already has a default width height and height specified if I'm not mistaken: https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs#L122-L123:
```cs
            #code snipped for brevity
            var splitter = new Splitter
            {
                Width = 660,
                Height = 800
            };
            this.Content = new Scrollable
            {
                Content = splitter
            };
```

InfinityGhost (changes requested):
Form windows are not centered when they are created on GTK, they're slightly offset lower than they should be. ideally, they should be handled by the window manager and should not be positioned by us on Linux altogether.

InfinityGhost (approved):

PR #702, opened 2021-01-22
Fix manually installed plugin metadata

### LINKED ISSUE #693: Manually installed plugins don't have metadata
## Description

No metadata is found in manually installed plugins, and this results in the plugin not showing in the plugin manager.

InfinityGhost:
# Changes
- Revised LINQ queries
  - Installed plugins are at the top of the list
  - Update looks for matching plugins correctly

# Issues
- Closes #693

### DIFF
--- a/OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
+++ b/OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
@@ -146,20 +146,14 @@ namespace OpenTabletDriver.UX.Windows
         protected async Task Uninstall(DesktopPluginContext context)
         {
             context.Directory.Refresh();
-            if (context.Directory.Exists && await App.Driver.Instance.UninstallPlugin(context.FriendlyName))
-            {
-                AppInfo.PluginManager.UnloadPlugin(context);
-                await Refresh();
-            }
-            else if (!context.Directory.Exists)
-            {
-                AppInfo.PluginManager.UnloadPlugin(context);
-                await Refresh();
-            }
-            else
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
```diff
@@ -144,7 +145,13 @@ protected async Task Install(string path)
...
+            context.Directory.Refresh();
+            if (context.Directory.Exists && await App.Driver.Instance.UninstallPlugin(context.FriendlyName))
+            {
+                AppInfo.PluginManager.UnloadPlugin(context);
+                await Refresh();
+            }
+            else if (!context.Directory.Exists)
             {
                 AppInfo.PluginManager.UnloadPlugin(context);
                 await Refresh();
```
X9VoiD, line 157: Is there some reason we're doing it like this? Because if none then:
```suggestion
            if (context.Directory.Exists)
                await App.Driver.Instance.UninstallPlugin(context.FriendlyName)

            AppInfo.PluginManager.UnloadPlugin(context);
            await Refresh();
```

X9VoiD (changes requested):

OpenTabletDriver.UX/Windows/PluginManagerWindow.cs
InfinityGhost: The reason it is like that was if a manual delete was performed, however that would work with the downside of losing UI interactivity in the `else` block. Will clean this up to be more clear though.

X9VoiD: Ah, right. Can't see beyond what GitHub shows lol.

X9VoiD (approved):

PR #703, opened 2021-01-22
Add Support for Deco Mini7

Devrulx:


OpenTabletDriver/Configurations/XP-Pen/Deco mini7.json
```diff
@@ -0,0 +1,31 @@
+{
+  "Name": "XP-Pen Deco Mini7",
+  "DigitizerIdentifiers": [
+    {
+      "Width": 177.8,
+      "Height": 111.1,
+      "MaxX": 35560.0,
+      "MaxY": 22219.0,
```
X9VoiD, line 8: Height and MaxY doesn't match correctly. I would like to get a verification for MaxX and MaxY.

Open Tablet Debugger, and hover your pen on the bottom right most part of the physical tablet to try to get the highest possible value for `<Position>`. You may try getting the max on one axis at a time.

X9VoiD (changes requested):

OpenTabletDriver/Configurations/XP-Pen/Deco mini7.json
Devrulx: The width and height are from the manufacturer, and those are the numbers I get from the debugger.

PR #732, opened 2021-01-29
Add Huion H1611 auxiliary button support 

TheKingDavid18:
to make pen buttons on H1611 work I changed
"End": 64,
to
"End": 66,
Also to make tablet buttons work I changed the parse to the gaomon parser and added:
      "ButtonCount": 10,
However the number didn't affect it
Also the report IDs for the hs611 pen buttons are also 65 and 66, so that config file should have:
"End": 66,
"EndInclusive": true


X9VoiD:
Please do the PR branching off from master if you want it merged to master, and ButtonCount is not part of current settings format.

NoahDTheHorse:
I only had to option that let me make a new pull request, here and  add-wizrdpen-5x4config

InfinityGhost:
This needs to be rebased to current master before I'll do any review.

InfinityGhost:
Looks like this PR is now stale, marking as draft until I have time to fix all of this myself.

PR #744, opened 2021-01-31
Updated H610 Pro V2 configuration

Mrcubix:
# Is this a 420? :

No

# Verification:

https://discord.com/channels/615607687467761684/615611007951306863/805564430003339346

https://www.amazon.ca/dp/B00GWYYXV0

### DIFF
--- a/OpenTabletDriver/Configurations/Huion/H610 Pro V2.json
+++ b/OpenTabletDriver/Configurations/Huion/H610 Pro V2.json
@@ -17,7 +17,7 @@
       "ProductID": 109,
       "InputReportLength": 12,
       "OutputReportLength": null,
-      "ReportParser": "OpenTabletDriver.Tablet.TabletReportParser",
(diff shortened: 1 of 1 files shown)


InfinityGhost (approved):

InfinityGhost:
Appears to not include auxiliary support based on the config, can we get that supported?

X9VoiD:
This shouldn't be named `V2` as it has the same device string identifier. We have a lot of other configurations with PID 109 and 110 merged into a single one since they don't really differ. For something to be V2 for Huions it has to be different in some other way like device string.

InfinityGhost (changes requested):
We need to find some sort of other identifier to avoid configuration conflicts.

X9VoiD:
Oh, I thought it added a new configuration with `V2`. My bad, this is fine.

InfinityGhost:
> Oh, I thought it added a new configuration with `V2`. My bad, this is fine.

.

InfinityGhost (approved):
pepega

PR #746, opened 2021-02-01
Fix PreInterpolate Filters getting applied twice when an Interpolator is active

X9VoiD:
Since we are early disposing interpolators now, the previous `InterpolatorActive` is now broken with where it is used.  This PR solves it by checking the `PluginSettingStoreCollection Interpolators` for any active interpolator instead of instances of interpolators (which is to be constructed later)

**Pre-PR**

PreInterpolate filters get applied twice when an interpolator is active, since InterpolatorActive is always false during `Filter { set; }`
It exists on both interpolator's Filter list and the output mode's preFilter list.

**Post-PR**

PreInterpolate filters applied only once when an interpolator is active.
It properly exists on just the interpolator's Filter list.

### DIFF
--- a/OpenTabletDriver.Plugin/IDriver.cs
+++ b/OpenTabletDriver.Plugin/IDriver.cs
@@ -1,6 +1,5 @@
 using System;
 using OpenTabletDriver.Plugin.Output;
-using OpenTabletDriver.Plugin.Platform.Display;
 using OpenTabletDriver.Plugin.Tablet;
 
 namespace OpenTabletDriver.Plugin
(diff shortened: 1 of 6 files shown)


OpenTabletDriver.Plugin/IDriver.cs
```diff
@@ -12,7 +11,6 @@ public interface IDriver
         event EventHandler<TabletState> TabletChanged;
 
         bool EnableInput { set; get; }
-        bool InterpolatorActive { get; }
```
InfinityGhost, line 15: API breaking change, this would have to wait until we close the `v0.5.x` milestone.

InfinityGhost (commented):
Did a quick review, introduces API breaking changes so it's gonna have to wait until the next tag unless they're resolved.

X9VoiD:
This PR is superseded by #812 if it'll be merged in the far future. I may close this one right now so we don't have to go through two API breaking changes.

X9VoiD:
This PR is not needed anymore, will close.

PR #751, opened 2021-02-02
Added 1060 Plus configuration

Mrcubix:
téléchargement

# Verification :

https://discord.com/channels/615607687467761684/789348845372178482/806256756044726312


InfinityGhost (changes requested):
Configuration name and filename do not match. if this is a New 1060 Plus, it should be on that configuration.

InfinityGhost (approved):

PR #754, opened 2021-02-04
Fix area rotation bug

InfinityGhost:
Fixed rotation using top left and bottom right corners, instead of max points. This caused the bounds to be incorrectly calculated for rotations that aren't divisible by 90 degrees.


OpenTabletDriver.UX/Controls/Area/AreaDisplay.cs
```diff
@@ -276,5 +279,13 @@ private static bool IsValid(RectangleF rect)
...
             return rect.Width > 0 && rect.Height > 0;
         }
+
+        private static bool IsInBounds(RectangleF bounds, RectangleF rect)
+        {
+            return bounds.Left >= rect.Left &&
+                bounds.Top >= rect.Top &&
+                bounds.Bottom <= rect.Bottom &&
+                bounds.Right <= rect.Right;
+        }
```
InfinityGhost, line 289: Was hoping this would resolve our bound clipping issue, but it doesn't seem to help. I think our only options at this point are to manually calculate these based on the original points, due to floating point error.

PR #770, opened 2021-02-13
Fix area sticking to boundaries while dragging

### LINKED ISSUE #769: Area dragging sticks to boundaries
## Description

When dragging areas, the deltas appear to be within floating point precision bounds causing the area to stick on the bottom and right of areas. 

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Software Version | 0ddc86ce36ea8a27d5f2f39b4f8fc355c01d689f

AbstractQbit:
Area position now gets corrected by the error amount instead of getting set to the nearest valid position.

Resolves #769.


InfinityGhost:
Precision still appears to be an issue, since max bounds aren't lined up correctly.
Right clicking and aligning manually goes past the boundaries, so this fix needs some sort of adjustment to make up for that.

### Dragged
image

### Aligned
image

InfinityGhost (approved):

PR #779, opened 2021-02-17
Fix area resize

X9VoiD:
## Changes

- Add "Update proportionally to display area change" on tablet area editor context menu.
- Fix "Tablet area resize is performed in reverse when changing display area size".

## Test

1. Enable Lock Aspect Ratio (required for Pre-PR to trigger the bug)
2. Change display area to quarter area (manual editing or changing mapped monitor of different resolution also triggers the bug)

### Pre-PR

The tablet area becomes bigger instead of smaller to maintain the *feel* or physical distance travelled to reach the same point on screen.

### Post-PR

Tablet area resizes to a quarter of its original size, reflecting the changes that happened on display area.

### DIFF
--- a/OpenTabletDriver.Desktop/Settings.cs
+++ b/OpenTabletDriver.Desktop/Settings.cs
@@ -20,7 +20,7 @@ namespace OpenTabletDriver.Desktop
 
         private float _dW, _dH, _dX, _dY, _tW, _tH, _tX, _tY, _r, _xS, _yS, _relRot, _tP;
         private TimeSpan _rT;
-        private bool _lockar, _sizeChanging, _autoHook, _clipping, _areaLimiting, _lockUsableAreaDisplay, _lockUsableAreaTablet, _updateTAreatoDArea;
+        private bool _lockar, _sizeChanging, _autoHook, _clipping, _areaLimiting, _lockUsableAreaDisplay, _lockUsableAreaTablet;
         private PluginSettingStore _outputMode, _tipButton;
 
         private PluginSettingStoreCollection _filters = new PluginSettingStoreCollection(),
@@ -77,8 +77,7 @@ namespace OpenTabletDriver.Desktop
             {
                 var prevValue = DisplayWidth;
                 RaiseAndSetIfChanged(ref _dW, value);
-                if (UpdateTabletAreaToDisplayChanges)
-                    TabletWidth *= DisplayWidth / prevValue;
+                TabletWidth *= DisplayWidth / prevValue;
             }
             get => _dW;
         }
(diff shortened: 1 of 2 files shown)


InfinityGhost:
Not sure if I'm a big fan of the additional option, its not very clear why you would have both lock aspect ratio and this. Maybe the default behavior should be this instead.

X9VoiD:
Toggle removed

OpenTabletDriver.Desktop/Settings.cs
```diff
@@ -75,9 +75,9 @@ public float DisplayWidth
         {
             set
             {
+                var prevValue = DisplayWidth;
                 RaiseAndSetIfChanged(ref _dW, value);
-                if (LockAspectRatio)
-                    TabletHeight = DisplayHeight / DisplayWidth * TabletWidth;
+                TabletWidth *= DisplayWidth / prevValue;
```
InfinityGhost, line 80: Doesn't this make the aspect ratio *always* locked?

OpenTabletDriver.Desktop/Settings.cs
```diff
@@ -87,9 +87,9 @@ public float DisplayHeight
         {
             set
             {
+                var prevValue = DisplayHeight;
                 RaiseAndSetIfChanged(ref _dH, value);
-                if (LockAspectRatio)
-                    TabletWidth = DisplayWidth / DisplayHeight * TabletHeight;
+                TabletHeight *= DisplayHeight / prevValue;
```
InfinityGhost, line 92: Same here

OpenTabletDriver.UX/Controls/OutputModeEditor.cs
```diff
@@ -255,7 +255,8 @@ public class TabletAreaEditor : AreaEditor
                 public TabletAreaEditor()
                     : base()
                 {
-                    this.ToolTip = "You can right click the area editor to enable aspect ratio locking, adjust alignment, or resize the area.";                }
+                    this.ToolTip = "You can right click the area editor to enable aspect ratio locking, adjust alignment, or resize the area.";
+                }
```
InfinityGhost, line 259: What changed here to trigger this?

InfinityGhost (changes requested):

OpenTabletDriver.Desktop/Settings.cs
X9VoiD: I can probably change the math a little bit so it won't resize the area and just change the ratios, because if I add it back to `LockAspectRatio` if block then it's not really lock aspect ratio, but proportional resize. That's why there was a new toggle prior to the review.

InfinityGhost:
Don't like the direction of this PR, it's not a good solution to what we're trying to solve.

PR #784, opened 2021-02-19
Metadata timeout

X9VoiD:
## Test 1

Open plugin manager without internet.

### Pre-PR

- OTD crashes.

### Post-PR

- OTD shows a warning about internet, and then proceeds to only show installed plugins.

## Test 2

Open plugin manager with a bad internet.

### Pre-PR

- OTD will either hang or crash depending on the nature of the problem.

### Post-PR

- OTD timeouts after 5 seconds and then says that it timed-out fetching metadata.
- It will capture any other unhandled exception, show in message box and log to console.
- OTD then proceeds to only show installed plugins.

### DIFF
--- a/OpenTabletDriver.Desktop/Reflection/Metadata/PluginMetadataCollection.cs
+++ b/OpenTabletDriver.Desktop/Reflection/Metadata/PluginMetadataCollection.cs
@@ -30,7 +30,7 @@ namespace OpenTabletDriver.Desktop.Reflection.Metadata
 
         protected static GitHubClient GitHub { get; } = new GitHubClient(new ProductHeaderValue("OpenTabletDriver"));
 
-        public static readonly PluginMetadataCollection Empty = new PluginMetadataCollection();
+        public static PluginMetadataCollection Empty => new PluginMetadataCollection();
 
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Desktop/Reflection/Metadata/PluginMetadataCollection.cs
```diff
@@ -29,6 +30,8 @@ protected PluginMetadataCollection(IEnumerable<PluginMetadata> source)
 
         protected static GitHubClient GitHub { get; } = new GitHubClient(new ProductHeaderValue("OpenTabletDriver"));
 
+        public static readonly PluginMetadataCollection Empty = new PluginMetadataCollection();
```
InfinityGhost, line 33: This should be a property, otherwise we're using the same empty instance with every empty collection.
```suggestion
        public static PluginMetadataCollection Empty => new PluginMetadataCollection();
```

InfinityGhost (changes requested):

InfinityGhost (approved):
Quick overview looks good, still yet to test however

PR #786, opened 2021-02-19
Add Download Count badge

Vendicated:
Not sure if this is something you want, just figured this might be useful to some people


InfinityGhost (approved):
This is actually a nice addition, thanks for the contribution.

Issue #700, opened 2021-01-22
Unplugging tablet while data is reporting throws OOR


InfinityGhost:
## Description

Exception is thrown when unplugging while the tablet is reporting data.
```
[ArgumentOutOfRangeException:Error]     Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'startIndex')
   at System.BitConverter.ToInt16(Byte[] value, Int32 startIndex)
   at System.BitConverter.ToUInt16(Byte[] value, Int32 startIndex)
   at OpenTabletDriver.Tablet.TabletReport..ctor(Byte[] report) in /repos/OpenTabletDriver/OpenTabletDriver/Tablet/TabletReport.cs:line 15
   at OpenTabletDriver.Tablet.TabletReportParser.Parse(Byte[] data) in /repos/OpenTabletDriver/OpenTabletDriver/Tablet/TabletReportParser.cs:line 9
   at OpenTabletDriver.Devices.DeviceReader`1.Main() in /repos/OpenTabletDriver/OpenTabletDriver/Devices/DeviceReader.cs:line 63
```

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System | Arch Linux
| Software Version | 5fadee0efa81f9c05fd06de07e6f6eda9cc0d51e
| Tablet           | CTL-470

PR #790, opened 2021-02-21
Addresses #700 - Unplugging Tablet Generates Incomplete Report Data

George-lewis:
If the tablet is unplugged while reading data, `byte[] report` may be incomplete causing the parser to throw an `ArgumentOutOfBoundsException`. We can handle this by catching the exception.

### DIFF
--- a/OpenTabletDriver/Devices/DeviceReader.cs
+++ b/OpenTabletDriver/Devices/DeviceReader.cs
@@ -72,6 +72,10 @@ namespace OpenTabletDriver.Devices
             {
                 Log.Write("Device", "Device disconnected.");
             }
+            catch (ArgumentOutOfRangeException)
+            {
+                Log.Write("Device", "Not enough report data returned by the device. Was it disconnected?");
+            }
             catch (Exception ex)
             {
                 Log.Exception(ex);


OpenTabletDriver/Tablet/TabletReport.cs
```diff
@@ -8,6 +8,11 @@ public struct TabletReport : ITabletReport
     {
         internal TabletReport(byte[] report)
         {
+            if (report.Length < 8)
+            {
+                throw new ArgumentException("Not enough report data.");
+            }
+
```
X9VoiD, line 15: This only solves it for a single report parser and adds an `if` on a hot path. It doesn't solve the issue for the wacom parsers. Remove this and just catch `ArgumentOutOfRangeException` instead on `DeviceReader.cs`. That way, it'll fix OOR issue on all parsers and avoid a branch on a hot path.

OpenTabletDriver/Devices/DeviceReader.cs
```diff
@@ -72,6 +72,10 @@ protected void Main()
             {
                 Log.Write("Device", "Device disconnected.");
             }
+            catch (ArgumentException aex) when (aex.Message == "Not enough report data.")
+            {
+                Log.Write("Device", "Not enough report data returned by the device. Was it disconnected?");
+            }
```
X9VoiD, line 78:
```suggestion
            catch (ArgumentOutOfRangeException)
            {
                Log.Write("Device", "Not enough report data returned by the device. Was it disconnected?");
            }
```

X9VoiD (changes requested):

X9VoiD (approved):

OpenTabletDriver/Tablet/TabletReport.cs
George-lewis: Thank you for the quick reply. I agree this is much better, now there is no need to proactively throw an exception.

AbstractQbit:
I'd recommend squashing this pr, this level of detail seems really unnecessary for 4 lines of change.

George-lewis:
> I'd recommend squashing this pr, this level of detail seems really unnecessary for 4 lines of change.

I think I got it. Look good to you?

AbstractQbit (approved):

InfinityGhost (approved):
Always good to have more descriptive messages for these sorts of issues.

PR #794, opened 2021-02-22
Fix duplicate context menu entries on OTD Guide

X9VoiD:
Apparently, `OnLoadComplete` is called on instantiated pages every time you navigate a page on `DocumentPage`. This is most likely an Eto bug.

## Test

- Open OTD Guide
- Navigate to Area Editor
- Navigate away and then back to Area Editor

### Pre-PR

image

### Post-PR

image


InfinityGhost (approved):

InfinityGhost:
This is a hacky workaround considering the issue (might actually be potential memory leaks there) but it works, so w/e

PR #795, opened 2021-02-22
Add IntuosV2AuxReport, bump aux buttons to 8, discard idle reports

AbstractQbit:


OpenTabletDriver/Vendors/Wacom/IntuosV3ReportParser.cs
```diff
@@ -6,10 +6,12 @@ public class IntuosV3ReportParser : IReportParser<IDeviceReport>
...
         {
-            if (data[0] == 0x11)
-                return new IntuosV3AuxReport(data);
-            else
-                return new IntuosV3Report(data);
+            return data[0] switch
+            {
+                0x10 => new IntuosV3Report(data),
+                0x11 => new IntuosV3AuxReport(data),
+                _ => null
```
InfinityGhost, line 13: This might have potential side effects we're not aware of, do output modes handle null reports?

AbstractQbit: If the report is null it won't be emitted 
https://github.com/InfinityGhost/OpenTabletDriver/blob/55b21bd58c66ca2c46b5c86f42b5c1f0ea4a24ce/OpenTabletDriver/Devices/DeviceReader.cs#L64-L65

OpenTabletDriver/Devices/DeviceReader.cs
```diff
@@ -61,7 +61,8 @@ protected void Main()
                 {
                     var data = ReportStream.Read();
                     var report = Parser.Parse(data);
-                    Report?.Invoke(this, report);
+                    if (report is not null)
```
InfinityGhost, line 64:
```suggestion
                    if (report != null)
```

InfinityGhost (changes requested):
One minor change, otherwise this is good.

InfinityGhost (approved):

PR #797, opened 2021-02-23
Handle Missing Config Directory

George-lewis:
I found this bug while investigating a related issue. In my source tree the config directory didn't exist and this code was causing a hard crash. Now, if the directory doesn't exist, the function returns an empty list. The directory would then be created later in `WriteConfigurations`.

### DIFF
--- a/OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
+++ b/OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
@@ -117,6 +117,7 @@ namespace OpenTabletDriver.UX.Windows.Configurations
 
         private ObservableCollection<TabletConfiguration> ReadConfigurations(DirectoryInfo dir)
         {
+            dir.Refresh();
             if (dir.Exists)
             {
                 var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
--- a/OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
+++ b/OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
@@ -117,9 +117,17 @@ namespace OpenTabletDriver.UX.Windows.Configurations
 
         private ObservableCollection<TabletConfiguration> ReadConfigurations(DirectoryInfo dir)
         {
-            var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
```diff
@@ -117,9 +117,16 @@ orderby config.Name
 
         private ObservableCollection<TabletConfiguration> ReadConfigurations(DirectoryInfo dir)
         {
-            var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
-                select Serialization.Deserialize<TabletConfiguration>(file);
-            return new ObservableCollection<TabletConfiguration>(configs);
+            if (dir.Exists)
+            {
+                var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
+                              select Serialization.Deserialize<TabletConfiguration>(file);
```
InfinityGhost, line 123:
```suggestion
                var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
                    select Serialization.Deserialize<TabletConfiguration>(file);
```

OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
```diff
@@ -117,9 +117,16 @@ orderby config.Name
 
         private ObservableCollection<TabletConfiguration> ReadConfigurations(DirectoryInfo dir)
         {
-            var configs = from file in dir.GetFiles("*.json", SearchOption.AllDirectories)
-                select Serialization.Deserialize<TabletConfiguration>(file);
-            return new ObservableCollection<TabletConfiguration>(configs);
+            if (dir.Exists)
```
InfinityGhost, line 120: Depending on where the `dir` variable came from, the `Exists` property isn't always up to date, however in this case it *should* be fairly safe to use. Would be a good idea to add a `dir.Refresh()` call to ensure its up to date.

InfinityGhost (changes requested):

OpenTabletDriver.UX/Windows/Configurations/ConfigurationEditor.cs
George-lewis: I've added a `.Refresh()` call.

InfinityGhost (approved):

PR #799, opened 2021-02-23
Add string dump to device string reader

X9VoiD:
[Sample output](https://github.com/InfinityGhost/OpenTabletDriver/files/6031481/a.txt)

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -417,21 +417,17 @@ namespace OpenTabletDriver.Daemon
 
         public Task<string> RequestDeviceString(int index)
         {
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -417,21 +417,17 @@ public Task SetTabletDebug(bool enabled)
 
         public Task<string> RequestDeviceString(int index)
         {
-            return Task.FromResult(Driver.TabletReader?.Device?.GetDeviceString(index) ?? null);
+            if (Driver.TabletReader?.Device != null)
+                return Task.FromResult(Driver.TabletReader?.Device?.GetDeviceString(index));
```
George-lewis, line 421:
```suggestion
                return Task.FromResult(Driver.TabletReader.Device.GetDeviceString(index));
```

We know these fields aren't null due to the guard

InfinityGhost (approved):

PR #804, opened 2021-02-26
Round area width and height before displaying value

X9VoiD:
Values inside the rectangle are rounded off.

image


InfinityGhost (approved):
Reasonable.

PR #801, opened 2021-02-25
Fix redundant settings rebind

X9VoiD:
## Changes

- Set `AreaViewModel` and related setup during ctor instead of OnLoadComplete

Fixes slow down and hang issues when switching output modes, also fixes #992

> Full breakdown of the bug: https://discord.com/channels/615607687467761684/628651971267788800/835818812913090580

## Test

Switch output modes

### Pre-PR

Around the 5th switch OTD will either start to slow down on switching or just completely hang

### Post-PR

Works fine and no redundant binds on `Settings`

PR #699, opened 2021-01-21
Harden against possible settings loading error

X9VoiD:
As discussed from the Discord, all known errors from loading a settings is fixed in this PR:

- Delete settings that are inaccessible by `PluginManager` (Keep structure if `AuxButtons` or `PenButtons`)
- Migrate settings after loading plugins to avoid null reference after uninstalling an active plugin through UX Plugin Manager
- Never let `Settings` become null after Daemon's initialization (UX side settings loading not needed anymore)
- Recover what we can from the settings file and log reasons for not being able to

Closes #696

PR #812, opened 2021-02-28
Rework input

X9VoiD:
Depends on #699 and #801

## Changes

- `Interpolator`  -> `IAsyncFilter`
- `IVirtualTablet` -> `IPressureHandler`
- `IAsyncFilter` is handled by `AsyncFilterHandler` inside `OutputMode`
  - Only one async filter will be active at any given time
  - Removed the need for filtering synthetic tablet report
- `AbsoluteOutputMode` and `RelativeOutputMode` common logic merged to `BaseOutputMode`
- Common configurations for `IOutputMode` moved to `IOutputMode.OutputModeConfig`
- AbsoluteMode and RelativeMode now uses the same UX for editing
  - `ResetTime` replaced by `AsyncFilterHandler.WithinDelayTolerance` (simplified logic from now-deleted `Limiter`)
- `IAbsolutePointer` and `IRelativePointer` replaced by `IPointer`
- and some other changes that I might have forgotten

## Benchmark
Pre-Rework
| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
| Output | 66.07 ns | 0.859 ns | 0.718 ns |

Rework
| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
| Output | 42.64 ns | 0.877 ns | 1.109 ns |


X9VoiD:
CodeFactor bad :<

OpenTabletDriver.Plugin/Output/Async/AsyncFilterHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+                }
+            }
+        }
+
+        public static bool WithinDelayTolerance(double reportMs, double reportMsAvg)
+        {
+            if (reportMsAvg < 3)
+                return reportMs < 3;
+            else
+                return reportMs < (reportMsAvg * 1.5);
```
AbstractQbit, line 107: This function looks kinda sus
image
Maybe we can just simplify it to
```suggestion
            return (reportMs < 4.5f) | (reportMs < (reportMsAvg * 1.5));
```
?

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -1,126 +1,65 @@
...
-                        pressureHandler.SetPressure((float)tabletReport.Pressure / (float)Tablet.Digitizer.MaxPressure);
+            var pos = base.Transpose(report);
+            var deltaTime = stopwatch.Restart().TotalMilliseconds;
+            reportMsAvg += (deltaTime - reportMsAvg) * 0.1;
+            Vector2? delta;
 
-                    if (Transpose(tabletReport) is Vector2 position)
-                        Pointer.Translate(position);
-                }
+            if (lastPos.HasValue && pos.HasValue && !skipReport && AsyncFilterHandler.WithinDelayTolerance(deltaTime, reportMsAvg))
```
AbstractQbit, line 51: In this case of just checking nullable's value lazy eval will only hurt performance by unnecessary branching
```suggestion
            if (lastPos.HasValue & pos.HasValue & !skipReport & AsyncFilterHandler.WithinDelayTolerance(deltaTime, reportMsAvg))
```

OpenTabletDriver.Plugin/HPETDeltaStopwatch.cs
```diff
@@ -11,7 +11,7 @@ public HPETDeltaStopwatch(bool startRunning = true)
             start = isRunning ? internalWatch.Elapsed : default;
         }
 
-        public static TimeSpan RunitmeElapsed => internalWatch.Elapsed;
+        public TimeSpan RunitmeElapsed => internalWatch.Elapsed;
```
AbstractQbit, line 14: `internalWatch` is static, there is no need for `HPETDeltaStopwatch` instance to query app uptime
```suggestion
        public static TimeSpan RunitmeElapsed => internalWatch.Elapsed;
```

AbstractQbit (commented):
Some thoughts on what caughed my eye

OpenTabletDriver.Plugin/HPETDeltaStopwatch.cs
X9VoiD: Ah, this change was not intended. It was when I'm debugging a weird relative mode bug where it jumps even after skipping one, or two reports.

OpenTabletDriver.Plugin/Output/Async/AsyncFilterHandler.cs
X9VoiD: Is something like this fine?
```cs
            return reportMs < Math.Max(3, reportMsAvg * 1.5);
```

AbstractQbit: Hmm, yeah, that looks even better

X9VoiD:
Dropping the PR for now, but will keep the branch on my repo as reference.

PR #821, opened 2021-03-02
Fix lost data context on some context menu entries

### LINKED ISSUE #818: Some context menu entries are not updated
## Description
After enabling area aspect ratio lock and loading settings that have it unlocked it becomes impossible to disable it (even though the checkbox isn't checked) and the area isn't the same as saved. The only way is to disable the lock is to do it before loading the settings.

## System Information:
| Name             | Value |
| ---------------- | ----- |
| Operating System | Windows 10 64-bit 10.0.19042 Build 19042 |
| Software Version | 0.5.0 |
| Tablet           | Wacom CTL-472 |

X9VoiD:
Fixes #818
Depends on #801

## Test

Save a separate settings with unlocked aspect ratio with non-matching ratio of display and tablet, lock aspect ratio then load the separate settings with `File -> Load Settings`

### Pre-PR

Some context menu don't update appropriately and is left pointing to the old Settings instance.

Affected:
- "Lock aspect ratio"
- "Area clipping"
- "Ignore input outside area"

### Post-PR

Works as expected.

### DIFF
--- a/OpenTabletDriver.UX/Controls/Area/AreaDisplay.cs
+++ b/OpenTabletDriver.UX/Controls/Area/AreaDisplay.cs
@@ -11,9 +11,14 @@ namespace OpenTabletDriver.UX.Controls.Area
 {
     public class AreaDisplay : TimedDrawable, IViewModelRoot<AreaViewModel>
     {
+        public AreaDisplay(AreaViewModel viewModel)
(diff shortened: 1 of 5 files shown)


X9VoiD:
Superseded by #1059

PR #825, opened 2021-03-06
Move PluginManager to OpenTabletDriver.Desktop

InfinityGhost:
This should provide a decent performance increase at the driver constructor, which was previously around 600ms when loading all of the assemblies for a single reference.

Considering:
- Enum instead of the type path

or

- Leaving as is, figuring out a better system later.


peppy (approved):
Looks to fix the profiled issue

peppy:
Initial checks shows this removes the startup overhead, looks good.

PR #826, opened 2021-03-06
Remove virtual modifier of PluginManager methods

X9VoiD:
This allows the JIT compiler to probably inline the methods or apply some other optimizations not available when declaring them as `virtual`. We don't currently have any need for it, nor are we overriding these methods somewhere so we may remove it.

Depends on #825


X9VoiD:
Superseded by #827

PR #827, opened 2021-03-06
Reduce PluginManager ctor time

X9VoiD:
### Before
|            Method |     Mean | Error |
|------------------ |---------:|------:|
| PluginManagerCtor | 425.5 ms |    NA |

### After
|            Method |     Mean | Error |
|------------------ |---------:|------:|
| PluginManagerCtor | 104.8 ms |    NA |

> `[MethodImpl(MethodImplOptions.AggressiveOptimization)]` staves off an extra 50ms

## Changes

- Avoided unnecessary use of `TypeInfo`
- Retrieve OTD plugin library type from just the OTD plugin module/assembly
- Retrieve OTD internally implemented plugins from just the OTD module/assembly
- As a side-effect of the changes, we can also now load types from a not `IsLoadable` assembly
  - If the type can't really be loaded, we can also now log which type, and for what reason

### DIFF
--- a/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs
+++ b/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs
@@ -108,9 +108,8 @@ namespace OpenTabletDriver.Desktop.Reflection
 
                 try
                 {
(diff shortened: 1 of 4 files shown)


InfinityGhost:
Resolve merge conflicts and I'll review.

PR #832, opened 2021-03-08
Implement dictionary-based plugin manager

X9VoiD:
## Changes

- Use `Dictionary<Type, List<Type>>` as the store
- Perform initial reflection asynchronously

Supersedes #827


X9VoiD:
We should fix the current plugin manager issues first so we would know which issues came from this PR.

PR #865, opened 2021-03-15
Added PTK-640 and PTH-650 Configurations

danielyko1017:
[Verification (discussion starts from here)](https://discord.com/channels/615607687467761684/787072060563128321/821103128463015978)

This is my first time doing this so I apologize if there are any major errors or things that I forgot to change

### DIFF
--- a/OpenTabletDriver/Configurations/Wacom/PTH-650.json
+++ b/OpenTabletDriver/Configurations/Wacom/PTH-650.json
@@ -1,0 +1,63 @@
+{
+  "Name": "Wacom PTH-650",
+  "DigitizerIdentifiers": [
+    {
+      "Width": 223.52,
+      "Height": 139.7,
+      "MaxX": 44704.0,
+      "MaxY": 27940.0,
+      "MaxPressure": 2047,
+      "ActiveReportID": {
+        "Start": 0,
+        "StartInclusive": true,
+        "End": null,
+        "EndInclusive": false
+      },
+      "VendorID": 1386,
+      "ProductID": 39,
+      "InputReportLength": 10,
+      "OutputReportLength": null,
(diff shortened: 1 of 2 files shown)


InfinityGhost (approved):
Looks good, anything nonfunctional on the tablet?

danielyko1017:
Overall, the tablet appears to be pretty much entirely functional (hovering, dragging without a large tilt angle, pen pressure, pen compatibility, etc.). Testing with the tablet debugger and normally playing osu with the tablet seemed to be fully functional on my end. 

However, I want to note that both tablets have some slight issues when you tilt the pen while dragging (eg. ~60 degrees for the PTK-640 and ~30 degrees for the PTH-650), since it gives a reportID of 0 and that's the same reportID as the no signal reportID. Since there was such a small angle needed for the PTH-650, I made it so that the 0 reportID is inclusive so the tablet will work exactly as intended BUT the cursor will jump to the top-left of the screen (ie. set a position of (0,0)) when the pen is lifted off. However, I made it non-inclusive for the PTK-640 since the tilt angle needed for a 0 reportID is very large. [You can see the discussion about this in the server here, along with images of how much tilt is needed for both](https://discord.com/channels/615607687467761684/787072060563128321/821118342647250955)

This also happens with some other pens that are compatible with the tablets (such as the LP180) but it's usually much worse since the tilt angle needed to get a 0 reportID is much smaller. Also, the 0 reportID is basically not an issue with hovering since the reportID is much higher compared to dragging. I believe Kuuube was talking to AbstractQbit about having some sort of fix made for the reportID 0 issue because it appears to be a consistent issue for a majority of the Pro-series tablets so this won't be an issue once that is implemented.

Issue #864, opened 2021-03-14
Settings for Huion Q11K (Non-V2)


seritools:
## Description
Hey, I've checked out the configuration for the Q11K (not V2). Since it's in the list as "not tested" I expected to debug myself a bit :)

I had to change a few minor things to get it working:

- My Q11K has the same Product ID as the Q11K V2: I changed `"ProductID": 109,` to `"ProductID": 110,`.
- There seems to be a typo in the device string: It says `"201": "HUION_TI64_\\d{6}$"`, but it should be `"201": "HUION_T164_\\d{6}$"` (note the `I` vs `1`)

After these two changes the device is detected and regular pen operation (including both pen buttons) work fine. I've checked the values in the tablet debugger though, and the `MaxX` and `MaxY` as well as the `Height` should match those of the V2 variant.

The 8 buttons on the tablet itself do not work yet. Pressing them does not change anything in the tablet debugger raw data view either :(

Any advice on how to debug that?

I've added my dump here: [dump.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6137881/dump.txt)
And my current, modified config here: [Inspiroy Q11K.json.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6137907/Inspiroy.Q11K.json.txt)

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System | Windows 10 Insider 21322
| Software Version | 0.5.2.2
| Tablet           | Huion Q11K (not V2)

Mrcubix:
first of all, don't change setting from an existing variation, copy paste the identifier and you'll obtain something such as
```
{
  "Name": "Huion Inspiroy Q11K",
  "DigitizerIdentifiers": [
    {
      "Width": 279.4,
      "Height": 146.6,
      "MaxX": 51680.0,
      "MaxY": 33020.0,
      "MaxPressure": 8191,
      "ActiveReportID": {
        "Start": 64,
        "StartInclusive": true,
        "End": null,
        "EndInclusive": false
      },
      "VendorID": 9580,
      "ProductID": 109,
      "InputReportLength": 12,
      "OutputReportLength": null,
      "ReportParser": "OpenTabletDriver.Tablet.TabletReportParser",
      "FeatureInitReport": null,
      "OutputInitReport": null,
      "DeviceStrings": {
        "201": "HUION_TI64_\\d{6}$"
      },
      "InitializationStrings": [
        200
      ]
    },
[...]

PR #886, opened 2021-03-20
Change GaomonReportParser to UCLogicReportParser

X9VoiD:
To better represent the range that the parser supports, rename it to `UCLogicReportParser`

## Changes
- `GaomonReportParser` -> `UCLogicReportParser`
- `GaomonAuxReport` -> `UCLogicAuxReport`
- Use `BitVector32` for retrieving active aux buttons
- Simplify code for aux detection

Helps with #864


InfinityGhost (changes requested):
I'm fine with these changes, but we also need to update EVERY configuration which relied on the old parser path.

X9VoiD:
Configurations updated.

InfinityGhost (approved):

PR #887, opened 2021-03-20
Update XP-Pen report parser

X9VoiD:
## Changes

- Support 8 buttons
- Use `BitVector32` wherever profitable
- Simplify aux report detection
  - Remove redundant check on `XP_PenAuxReport` (already performed in `XP_PenReportParser`)
- Skip BitConverter's unneeded safety checks by directly calling `Unsafe.ReadUnaligned<T>()`
  > We know in advance that the bytes we're parsing are complete

### DIFF
--- a/OpenTabletDriver/Vendors/XP_Pen/XP_PenAuxReport.cs
+++ b/OpenTabletDriver/Vendors/XP_Pen/XP_PenAuxReport.cs
@@ -1,3 +1,4 @@
+using System.Collections.Specialized;
 using OpenTabletDriver.Plugin.Tablet;
 
 namespace OpenTabletDriver.Vendors.XP_Pen
@@ -7,20 +8,19 @@ namespace OpenTabletDriver.Vendors.XP_Pen
         public XP_PenAuxReport(byte[] report)
         {
             Raw = report;
-            var ReportID = (uint)report[1] >> 1;
-            AuxButtons = new bool[6];
-            if (ReportID == 120) 
+
+            var bitVector = new BitVector32(report[2]);
+            AuxButtons = new bool[]
             {
-                AuxButtons = new bool[] 
-                {
-                    (report[2] & (1 << 0)) != 0,
-                    (report[2] & (1 << 1)) != 0,
-                    (report[2] & (1 << 2)) != 0,
-                    (report[2] & (1 << 3)) != 0,
-                    (report[2] & (1 << 4)) != 0,
-                    (report[2] & (1 << 5)) != 0
-                };
-            }
+                bitVector[1 << 0],
+                bitVector[1 << 1],
+                bitVector[1 << 2],
+                bitVector[1 << 3],
+                bitVector[1 << 4],
+                bitVector[1 << 5],
+                bitVector[1 << 6],
+                bitVector[1 << 7]
+            };
         }
 
         public bool[] AuxButtons { set; get; }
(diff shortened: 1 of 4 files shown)


OpenTabletDriver/Vendors/XP_Pen/XP_PenAuxReport.cs
```diff
@@ -7,20 +8,19 @@ public struct XP_PenAuxReport : IAuxReport
...
-            }
+                bitVector[1 << 0],
+                bitVector[1 << 1],
+                bitVector[1 << 2],
+                bitVector[1 << 3],
+                bitVector[1 << 4],
+                bitVector[1 << 5],
+                bitVector[1 << 6],
+                bitVector[1 << 7]
+            };
```
InfinityGhost, line 23: How does this actually benefit performance? It certainly makes it more readable however.

InfinityGhost (commented):
If there is actually tangible evidence this improves performance, this could be applied to other report parsers as well.

OpenTabletDriver/Vendors/XP_Pen/XP_PenAuxReport.cs
X9VoiD: Before:

|               Method |     Mean |    Error |   StdDev |
|--------------------- |---------:|---------:|---------:|
|         ReportParser | 35.34 ns | 0.602 ns | 0.563 ns |
| SkipByteReportParser | 50.80 ns | 0.686 ns | 0.608 ns |
|    **XPPenReportParser** | **35.84 ns** | 0.266 ns | 0.208 ns |

After:

|               Method |     Mean |    Error |   StdDev |
|--------------------- |---------:|---------:|---------:|
|         ReportParser | 35.31 ns | 0.445 ns | 0.371 ns |
| SkipByteReportParser | 51.45 ns | 0.773 ns | 0.685 ns |
|    **XPPenReportParser** | **26.05 ns** | 0.255 ns | 0.213 ns |

So basically the new parser is 37.58% faster

X9VoiD: ## XPPenTiltReportParser

Before:

|                Method |     Mean |    Error |   StdDev |
|---------------------- |---------:|---------:|---------:|
| XPPenTiltReportParser | 45.83 ns | 0.939 ns | 0.922 ns |

After:

|                Method |     Mean |    Error |   StdDev |
|---------------------- |---------:|---------:|---------:|
| XPPenTiltReportParser | 39.63 ns | 0.372 ns | 0.310 ns |

The new tilt report parser is 15.64% faster, most of the time is spent on memory operations (heap alloc and boxing)

## XPPenAux

Before:

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| XPPenAuxParser | 35.29 ns | 0.588 ns | 0.521 ns |

After:

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| XPPenAuxParser | 25.95 ns | 0.449 ns | 0.420 ns |

The new aux parser is 36% faster

> Note: Had to set benchmark's data[1] to 0x00 for the first benchmark and 0xc0 for the aux benchmark

InfinityGhost: Worth the change, we should go ahead and make these changes to other report parsers while we're at it, ideally in separate branches so we can merge on a per-manufacturer basis.

InfinityGhost (approved):
> Worth the change, we should go ahead and make these changes to other report parsers while we're at it, ideally in separate branches so we can merge on a per-manufacturer basis.

PR #888, opened 2021-03-20
Rename XP_Pen to XP

X9VoiD:
This is mostly just aesthetic change so I moved this change to this separate PR in case this change is unwanted

## Changes

- `XP_Pen` -> `XP`

Depends on #887


InfinityGhost:
I hate this lol

Issue #892, opened 2021-03-20
Better user-facing error messages about common issues


bdach:
After integrating OTD in lazer, we are starting to receive issue reports about common issues that are mentioned [in the FAQ](https://github.com/OpenTabletDriver/OpenTabletDriver/wiki/General-FAQ) already (see https://github.com/ppy/osu/issues/12101 for an example). For everyone's sanity, I'm thinking it would be good if the errors were described better and if there was a way to guide the users to the destination they need to head to directly, instead of having them post a lazer issue, only then to be directed to an OTD FAQ page.

It would be nice if these common errors were re-exposed in a more user-friendly manner, so that we can tell the user what's wrong exactly and how to fix it. I'm imagining that alongside a human-readable error, an optional URL to the appropriate FAQ section/to the issue creation page could be provided by OTD, so that we may for example redirect the user to said page when they click the error notification. Other applications that might want to integrate OTD may benefit from the same information as well.

InfinityGhost:
#893 should resolve this now, still want an outsider POV on this.

bdach:
If it's my perspective you're looking for, then I think that error message looks way better now. Thanks!

PR #893, opened 2021-03-20
Improve exception message of common issues

X9VoiD:
Intercept DeviceIOException and change it to something more descriptive. Include link to instructions as well.

Helps with #892

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using HidSharp;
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -100,6 +101,12 @@ public bool TryMatch(TabletConfiguration config)
                     return true;
                 }
             }
+            catch (IOException iex) when (iex.Message.Contains("Unable to open HID class device"))
```
InfinityGhost, line 104: This doesn't take into consideration what platform this is running on.

InfinityGhost (changes requested):

OpenTabletDriver/Driver.cs
X9VoiD: Fixed. But the problem with our approach is that these logs are spammed. Tried moving it to DriverDaemon to implement breaks but that means you'll have to implement the same try catch on lazer side.

InfinityGhost (approved):
lg2m

PR #910, opened 2021-03-24
Match digitizer/aux parser and remove wild configs

### LINKED ISSUE #909: CTL-4100 Buttons always pressed after receiving positional input
## Description

with the tablet buttons bound to anything, simply placing the pen above the tablet will cause the buttons to be pressed until they're actually physically pressed.
## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System | Arch Linux (happens on windows too though) |
| Software Version | v0.5.2.3  |
| Tablet           | Wacom CTL-4100 |

X9VoiD:
## Changes

- Match the parser used in aux with the digitizer's
- Remove "wild" configs that identifies as InputReportLength: null, OutputReportLength: null. These configs match anything and causes overlaps

Fixes #909


InfinityGhost (commented):
None of these configurations changed have VID: null nor PID: null, only null InputReportLength and OutputReportLength. This is also a batch change for known working configurations, not sure if this is the best option.

X9VoiD:
Edited. Was a typo.

The change regarding parsers are due to parsing overlap in certain parsers which caused the issue in #909. The other InputReportLength and OutputReportLength nulls will also cause problem for other tablets especially those that use it in aux. This PR will force us to fix these configurations rather than rely on "wild" matching.

tl; dr, 909 happened because it opened the same endpoint using different parsers.

peppy:
Any resolution to this one? Would be great to be able to get it fixed at our end as right now this is causing bindings to be unusable (buttons are always reported as down).

InfinityGhost (approved):
Might have to just approve this and revert if need be.

gonX (approved):

gonX:
Anything stopping this from getting merged?

PR #914, opened 2021-03-25
I/O pipeline overhaul

InfinityGhost:
Output pipeline rework, with GStreamer-like I/O pipeline.

- [x] Implement `IPipelineElement<T>`
- [x] Add Interpolator helper class
- [x] Fix debugger data modified by pipeline


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -113,9 +114,14 @@ public Task LoadPlugins()
...
             {
                 AppInfo.PluginManager.Load();
+
                 // Migrate if settings is available to avoid invalid settings
                 if (Settings != null)
                     Settings = SettingsMigrator.Migrate(Settings);
+
+                // Add services to inject on plugin construction
+                AppInfo.PluginManager.ResetServices();
+                AppInfo.PluginManager.AddService(() => SystemInterop.Timer);
```
X9VoiD, line 124: Cross-referencing to PluginManager.cs#L67
```cs
var service = GetService(property.PropertyType);
```
This and DependencyInjectionTest#L18 will fail because T here is `Func<ITimer>`
```cs
serviceManager.AddService<ITimer>(() => new CLRTimer());
```
The GetService() will therefore also return a `Func<ITimer>` object

X9VoiD (changes requested):

OpenTabletDriver.Daemon/DriverDaemon.cs
X9VoiD: I misread the parameters for `AddService()`. Will retract the review lol

PR #926, opened 2021-03-26
Make use of all the bit flags to detect aux

X9VoiD:
```
Aux:      b11110000 F0
OOR:      b00110000 C0
Previous: b00110000 C0
PR:       b11110000 F0
```
Previous method matches both the OutOfRange report and Aux (the problem was worked around before by performing two separate checks)

### DIFF
--- a/OpenTabletDriver/Vendors/XP_Pen/XP_PenReportParser.cs
+++ b/OpenTabletDriver/Vendors/XP_Pen/XP_PenReportParser.cs
@@ -7,7 +7,7 @@ namespace OpenTabletDriver.Vendors.XP_Pen
     {
         public IDeviceReport Parse(byte[] data)
         {
-            return (data[1] & 0xc0) == 0xc0 ? new XP_PenAuxReport(data) : new TabletReport(data);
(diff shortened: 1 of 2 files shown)


InfinityGhost (approved):
Makes sense, but still want at least someone else to confirm this works.

X9VoiD:
Referencing v0.5.2.3's [XP_PenAuxReport.cs#L10-12](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/8ade463019101b8ef3b3d6b521f4a3406ca7b134/OpenTabletDriver/Vendors/XP_Pen/XP_PenAuxReport.cs#L10)
```cs
            var ReportID = (uint)report[1] >> 1;
            AuxButtons = new bool[6];
            if (ReportID == 120) 
```
`F0` is equivalent to 240, and `report[1] >> 1` essentially just divides it to two before the check, so it matches the old behaviour except that we're only using the the 5th-8th bit (check only the bit flags that actually change, instead of the whole byte in case a certain tablet uses the first four bits for other purposes).

The bitshift from before was unnecessary.

PR #927, opened 2021-03-26
Fix issues on pipeline rework

X9VoiD:
## Issues fixed

- [x] OutputMode was not connected by default, so the benchmark fails
- [x] Null reference exception when AuxReader is null (for devices that don't have a separate aux interface)
- [x] Millimeter and PreTranspose is not actually in milimeter coordinates (it was in raw device coordinates)
- [x] RelativeOutputMode is broken (cursor on bottom right)
  - [x] Transform absolute report to delta report
  - [x] Fix ResetTime (remove jumps)
- [x] Broken AsyncPositionedPipelineElement<T> scheduler behaviour (`Start()` is called before Interval is set)
- [x] AsyncPositionedPipelineElement<T> implementer have no way of controlling when `Emit` is invoked
- [x] Rename `IVirtualMouse` and `IVirtualTablet` to `IMouseButtonHandler` and `IPressureHandler`
- [x] `PreTransform` is not connected to the end of pipeline if there is no `PostTransform` element present
- [x] `PostTransform` is not connected to the start of the pipeline if there is no `PreTransform` element present
- [x] Every access to IVirtualDisplay causes ctor

Initial benchmarks for reworked pipeline:

| Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| Output | 101.8 ns | 0.78 ns | 0.76 ns |     - |     - |     - |         - |

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
+++ b/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
@@ -22,7 +22,6 @@ namespace OpenTabletDriver.Desktop.Interop
         }
 
         private static IVirtualScreen virtualScreen;
-
         private static IAbsolutePointer absolutePointer;
         private static IRelativePointer relativePointer;
         private static IVirtualKeyboard virtualKeyboard;
--- a/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
+++ b/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
@@ -21,13 +21,7 @@ namespace OpenTabletDriver.Desktop.Interop
         {
         }
 
-        private static IVirtualScreen virtualScreen = CurrentPlatform switch
-        {
-            PluginPlatform.Windows => new WindowsDisplay(),
-            PluginPlatform.Linux => ConstructLinuxDisplay(),
-            PluginPlatform.MacOS => new MacOSDisplay(),
-            _ => null
-        };
+        private static IVirtualScreen virtualScreen;
 
         private static IAbsolutePointer absolutePointer;
         private static IRelativePointer relativePointer;
@@ -98,7 +92,13 @@ namespace OpenTabletDriver.Desktop.Interop
             _                      => null
         };
 
-        public static IVirtualScreen VirtualScreen => virtualScreen;
+        public static IVirtualScreen VirtualScreen => CurrentPlatform switch
+        {
+            PluginPlatform.Windows => virtualScreen ??= new WindowsDisplay(),
+            PluginPlatform.Linux => virtualScreen ??= ConstructLinuxDisplay(),
+            PluginPlatform.MacOS => virtualScreen ??= new MacOSDisplay(),
+            _ => null
+        };
 
         private static IVirtualScreen ConstructLinuxDisplay()
         {
--- a/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
+++ b/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
@@ -21,6 +21,14 @@ namespace OpenTabletDriver.Desktop.Interop
         {
         }
 
+        private static IVirtualScreen virtualScreen = CurrentPlatform switch
+        {
+            PluginPlatform.Windows => new WindowsDisplay(),
+            PluginPlatform.Linux => ConstructLinuxDisplay(),
+            PluginPlatform.MacOS => new MacOSDisplay(),
+            _ => null
+        };
+
         private static IAbsolutePointer absolutePointer;
         private static IRelativePointer relativePointer;
         private static IVirtualKeyboard virtualKeyboard;
@@ -90,13 +98,7 @@ namespace OpenTabletDriver.Desktop.Interop
             _                      => null
         };
 
-        public static IVirtualScreen VirtualScreen => CurrentPlatform switch
-        {
-            PluginPlatform.Windows => new WindowsDisplay(),
-            PluginPlatform.Linux   => ConstructLinuxDisplay(),
-            PluginPlatform.MacOS   => new MacOSDisplay(),
-            _                      => null
-        };
+        public static IVirtualScreen VirtualScreen => virtualScreen;
 
         private static IVirtualScreen ConstructLinuxDisplay()
         {
--- a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
+++ b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
@@ -9,6 +9,7 @@ namespace OpenTabletDriver.Desktop.Output
     [PluginName("Artist Mode"), SupportedPlatform(PluginPlatform.Linux)]
     public class LinuxArtistMode : AbsoluteOutputMode, IPointerProvider<IAbsolutePointer>
     {
-        public override IAbsolutePointer Pointer { set; get; } = new EvdevVirtualTablet();
+        private static IAbsolutePointer pointer = new EvdevVirtualTablet();
+        public override IAbsolutePointer Pointer { set; get; } = pointer;
     }
 }
--- a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
+++ b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
@@ -9,6 +9,7 @@ namespace OpenTabletDriver.Desktop.Output
     [PluginName("Artist Mode"), SupportedPlatform(PluginPlatform.Linux)]
     public class LinuxArtistMode : AbsoluteOutputMode, IPointerProvider<IAbsolutePointer>
     {
-        public override IAbsolutePointer Pointer { set; get; } = new EvdevVirtualTablet();
+        private static IAbsolutePointer pointer;
+        public override IAbsolutePointer Pointer { set; get; } = pointer ??= new EvdevVirtualTablet();
     }
 }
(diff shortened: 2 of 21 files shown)


OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -9,7 +9,6 @@ namespace OpenTabletDriver.Desktop.Output
     [PluginName("Artist Mode"), SupportedPlatform(PluginPlatform.Linux)]
     public class LinuxArtistMode : AbsoluteOutputMode, IPointerProvider<IAbsolutePointer>
     {
-        private static IAbsolutePointer pointer;
-        public override IAbsolutePointer Pointer { set; get; } = pointer ??= new EvdevVirtualTablet();
+        public override IAbsolutePointer Pointer { set; get; } = new EvdevVirtualTablet();
```
InfinityGhost, line 12: Revert this commit, this is entirely on purpose. We store the pointer statically so when modifying settings or switching output modes, the virtual tablet doesn't get reconstructed. Applications such as GIMP require restart to make use of the new virtual tablet.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Output/OutputMode.cs
```diff
@@ -44,6 +50,9 @@ public IList<IPositionedPipelineElement<IDeviceReport>> Elements
 
                 if (Elements != null && Elements.Count > 0)
                 {
+                    this.DeviceOutput -= this.Consume;
+                    this.Emit -= this.OnFinalReport;
```
InfinityGhost, line 54: Not sure if this is the best way to handle this potential bug, elements can still be cleared.

A private function to control this bypass technique would be a much safer and maintainable alternative with a field that stores the bypass state.

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
X9VoiD: Resolved. The null check is removed by statically setting the private pointer.

OpenTabletDriver.Plugin/Output/OutputMode.cs
X9VoiD: Made a private function that sets and unsets the passthrough.

X9VoiD:
Ready for merge.

For a test element, build [this branch](https://github.com/X9VoiD/OpenTabletDriver/tree/output-pipeline-test-element) (devocub filter ported as a pipeline element)

InfinityGhost:
>     *  Every access to IVirtualDisplay causes ctor

This isn't necessarily an issue really, since its intended to be stored in the state in which it is assigned.

However, it could be resolved in the same way we store the `IVirtualKeyboard` and the `IRelativePointer`/`IAbsolutePointer` on Linux where recreating the pointers creates desktop environment issues. Doing it this way will construct it on the first request to the `IVirtualScreen` but will always return that instance from that point forward until application exit.

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/7953a99582ec9882fa7ea7e2bb9fdc682f74e7d0/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs#L88

X9VoiD:
Yeah, my bad, forgot to update the check boxes. I used almost the same method. But since we'll always need the virtual display I used a switch statement on it and just returned it from the getter.

The IVirtualDisplay re-ctor was a problem due to Windows timer that we have to detect resolution change (somehow we already have it even tho the change only propagates to the pointer). Without the fix, there are two instances of the timer running, and spamming "DPI Awareness enabled" twice every second.

OpenTabletDriver.Plugin/Platform/Pointer/IMouseButtonHandler.cs
```diff
@@ -1,6 +1,6 @@
 namespace OpenTabletDriver.Plugin.Platform.Pointer
 {
-    public interface IVirtualMouse
+    public interface IMouseButtonHandler
```
InfinityGhost, line 3: Revert this name change, it is out of scope for this PR.

OpenTabletDriver.Plugin/Platform/Pointer/IPressureHandler.cs
```diff
@@ -1,6 +1,6 @@
 namespace OpenTabletDriver.Plugin.Platform.Pointer
 {
-    public interface IVirtualTablet
+    public interface IPressureHandler
```
InfinityGhost, line 3: Revert this name change as well, out of scope for this PR.

OpenTabletDriver.Plugin/Output/OutputMode.cs
```diff
@@ -80,9 +107,26 @@ public virtual TabletState Tablet
...
+
+        private void UnsetPassthrough()
+        {
+            if (isPassthrough)
+            {
+                this.entryElement = null;
+                this.Emit -= this.OnOutput;
+                isPassthrough = false;
+            }
+        }
```
InfinityGhost, line 130: Make this a property, sort of like so. Probably more needed to be changed here however.
```cs
        private bool passthrough;
        protected bool Passthrough
        {
            private set
            {
                if (value && !passthrough)
                {
                    this.entryElement = this;
                    this.Emit += this.OnOutput;
                    this.passthrough = true;
                }
                else if (!value && passthrough)
                {
                    this.entryElement = null;
                    this.Emit -= this.OnOutput;
                    this.passthrough = false;
                }
            }
            get => this.passthrough;
        }
```

OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
```diff
@@ -21,6 +21,14 @@ protected DesktopInterop()
         {
         }
 
+        private static IVirtualScreen virtualScreen = CurrentPlatform switch
+        {
+            PluginPlatform.Windows => new WindowsDisplay(),
+            PluginPlatform.Linux => ConstructLinuxDisplay(),
+            PluginPlatform.MacOS => new MacOSDisplay(),
+            _ => null
+        };
```
InfinityGhost, line 30: Implement this the same way we do for the other interfaces within this class for cleanliness (such as the pointers and `IVirtualKeyboard`) and so that the behavior is the same throughout. This will initialize at the same time as the static constructor, whereas the others will be initialized when requested.

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -9,7 +9,7 @@ namespace OpenTabletDriver.Desktop.Output
     [PluginName("Artist Mode"), SupportedPlatform(PluginPlatform.Linux)]
     public class LinuxArtistMode : AbsoluteOutputMode, IPointerProvider<IAbsolutePointer>
     {
-        private static IAbsolutePointer pointer;
-        public override IAbsolutePointer Pointer { set; get; } = pointer ??= new EvdevVirtualTablet();
+        private static IAbsolutePointer pointer = new EvdevVirtualTablet();
+        public override IAbsolutePointer Pointer { set; get; } = pointer;
```
InfinityGhost, line 13: This will initialize the pointer even when not ever used since the static constructor will be invoked initializing this field, resulting in wasted memory. This also will be constructed on non-linux platforms because of this.

OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
```diff
@@ -2,60 +2,100 @@
...
+            }
+        }
+
         /// <summary>
-        /// The current state of the <see cref="AsyncPositionedPipelineElement{T}"/>
+        /// Sets the internal state.
         /// </summary>
-        protected T State { set; get; }
+        /// <param name="value"></param>
+        protected abstract void ConsumeState();
```
InfinityGhost, line 74: Whats the purpose in this function? The documentation needs to be improved here.

OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
```diff
@@ -2,60 +2,100 @@
...
+        protected abstract void UpdateState();
 
         /// <summary>
-        /// Sets the internal state.
+        /// Determines if pen is in tablet hover range
         /// </summary>
-        /// <param name="value"></param>
-        public virtual void Consume(T value)
+        /// <returns><see cref="true"/> if pen is in range</returns>
+        protected bool PenIsInRange()
```
InfinityGhost, line 90: Could benefit from adding a remark explaining how this is determined.

InfinityGhost (changes requested):

OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
X9VoiD: Notifies the asynchronous element when a state is consumed,
> Lets the element to update its internal state, not to be confused with `State`)

And does so within a lock to avoid race conditions against `UpdateState`. Overriding `Consume` is not viable as that means the override will happen outside of the lock.

OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
X9VoiD: We determine that a pen is out of range when `PenIsInRange` is called within `UpdateState` after a time equivalent to a report and a half has already passed.

If however, the report interval is faster than 3ms, we instead check if 3ms has already passed before declaring that the pen is out of range.

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
InfinityGhost: We still need to store the virtual tablet statically, otherwise it will re-initialize the pointer on every settings change. Many applications require restart when the virtual tablet is re-initialized so this becomes an unnecessary inconvenience.

InfinityGhost (commented):
Almost ready, just need the linux artist mode fixed.

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
X9VoiD: Fixed

InfinityGhost (approved):

PR #958, opened 2021-04-05
Fix PTH-451 Config

Kuuuube:
Add featureinit and specify aux endpoint


InfinityGhost:
Resolve merge conflicts.

PR #962, opened 2021-04-06
Add Wacom touch support

AbstractQbit:
I have only tested this with cth-470, pth-850 and pth-860; 
8th and 9th gen consumer tablets need testing before merging this.

Also fixes #935.

PR #963, opened 2021-04-06
Fix intuosv2

AbstractQbit:
Depends on #962, because 64 byte endpoints shouldn't be opened with IntuosV2ReportParser.

~~CTx-x90 tablets need to be tested.~~ [CTL490 works, so other IntuosHT2 should work too](https://discord.com/channels/615607687467761684/628651971267788800/829070161721950238)

### DIFF
--- a/OpenTabletDriver/Configurations/Wacom/CTH-490.json
+++ b/OpenTabletDriver/Configurations/Wacom/CTH-490.json
@@ -8,7 +8,7 @@
       "MaxY": 9500.0,
       "MaxPressure": 2047,
       "ActiveReportID": {
-        "Start": 0,
+        "Start": 1,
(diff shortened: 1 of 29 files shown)


AbstractQbit:
Changed this pr to fix the entire Intuosv2 lineup overall. Missing dimensions data was scraped from [here](https://github.com/linuxwacom/input-wacom/blob/3cf44674021c2c80d514973de0daa2f93639e182/4.5/wacom_wac.c#L4327-L4365).

AbstractQbit:
Also this now supersedes #958

Kuuuube (approved):
Checked all the values for all the new configs and they all match what they should be.

InfinityGhost (approved):

Issue #960, opened 2021-04-05
Support Request: XP-Pen G430S & Huion H610PRO V2 undetected


SavvyKat:
# Support Request

Both my XP-Pen G430S and Huion H610PRO V2 tablets are not being detected in OTD. 
[Console Pastebin](https://pastebin.com/AX6sA5DZ)
I am aware of issues with XP-Pen G430S #920 #698, but I cannot get my H610PRO V2 tablet to connect either. I previously used hawku drivers; however, I made sure to uninstall the vmulti drivers and such. I'm unsure if that would complicate anything. 
I would prefer to use my G430s, but I'd be happy to connect either of them.

## What I've already tried

Made sure to uninstall any previous drivers.
[Uninstalled corresponding HID devices in device manager. ](https://github.com/OpenTabletDriver/OpenTabletDriver/wiki/Windows-FAQ#my-gaomonhuion-tablet-isnt-detected-but-is-listed-as-supported)
Uninstalled all phantom devices.
Tried looking in the discord for any information, and tried to use the device string reader to identify the tablet.
Lots of troubleshooting with restarting my computer and plugging in my tablets.

## Diagnostic Information

[Diagnostic Dumps.zip](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6259972/Diagnostic.Dumps.zip)
Dumped diagnostics three times; however, they saved as empty files and crashed the GUI. Decided to upload them anyway but I don't see how they would be of any use.

X9VoiD:
Are you running OTD inside some form of sandbox or a limited environment? Because `Failed to get info` is a very rare error on Windows. It means that the library we're using (HidSharp) failed to read the characteristics of a device, either because it got denied of access or you're running in an unsupported environment.

This issue is external to OTD and cannot be fixed from within OTD.

Mrcubix:
Talking about the issue on discord rn in misc @X9VoiD

SavvyKat:
Issue was caused by my Wooting Two keyboard. Unpluggin the keyboard before starting OTD lets my tablets connect just fine. Going to look into reporting this issue to Wooting.

PR #965, opened 2021-04-06
Fix bogus HID reports breaking tablet detection

BigBrainAFK:
This fix prevents devices from aborting the tablet detection prematurely.

The fix mainly applies to cases like #960 where another USB device hands out bogus HID reports with `CanOpen` set to `true`.
So we filter the `IEnumerable<HidDevice>` in `OpenTabletDrivers.Drivers.FindMatches` before we even check for a `CanOpen`.
This way we rule out broken third party devices that are not the tablets we are interested in.

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -330,14 +330,28 @@ namespace OpenTabletDriver
 
         private IEnumerable<HidDevice> FindMatches(DeviceIdentifier identifier)
         {
-            return from device in DeviceList.Local.GetHidDevices()
-                where identifier.VendorID == device.VendorID
-                where identifier.ProductID == device.ProductID
-                where device.CanOpen
-                where identifier.InputReportLength == null || identifier.InputReportLength == device.GetMaxInputReportLength()
-                where identifier.OutputReportLength == null || identifier.OutputReportLength == device.GetMaxOutputReportLength()
-                where DeviceMatchesAllStrings(device, identifier)
-                select device;
+            foreach (HidDevice device in DeviceList.Local.GetHidDevices().ToList())
+            {
+                try
+                {
+                    if (identifier.VendorID != device.VendorID ||
+                        identifier.ProductID != device.ProductID ||
+                        !device.CanOpen ||
+                        (identifier.InputReportLength != null && identifier.InputReportLength != device.GetMaxInputReportLength()) ||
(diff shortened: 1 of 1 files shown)


X9VoiD (approved):
This fixes the issue for most cases, however we'll have to keep in mind that the actual problem is in HidSharpCore. If in the future we encounter a tablet that HidSharpCore can't perform report reconstruction on then it'll crash on those cases.

InfinityGhost:
> This fixes the issue for most cases, however we'll have to keep in mind that the actual problem is in HidSharpCore. If in the future we encounter a tablet that HidSharpCore can't perform report reconstruction on then it'll crash on those cases.

Should add a try/catch block to handle that situation and report it explicitly into the log for that situation. Hard crashes are always  more difficult to diagnose.

X9VoiD (changes requested):
As per Infinity's suggestion, probably change the function to this or something similar
```cs
        private IEnumerable<HidDevice> FindMatches(DeviceIdentifier identifier)
        {
            foreach (var device in DeviceList.Local.GetHidDevices())
            {
                try
                {
                    if (identifier.VendorID != device.VendorID
                        || identifier.ProductID != device.ProductID
                        || !device.CanOpen
                        || identifier.InputReportLength != null && identifier.InputReportLength != device.GetMaxInputReportLength()
                        || identifier.OutputReportLength != null && identifier.OutputReportLength != device.GetMaxOutputReportLength()
                        || !DeviceMatchesAllStrings(device, identifier))
                    {
                        continue;
                    }
                }
                catch (Exception e)
                {
                    Log.Exception(e);
                    continue;
                }

                yield return device;
            }
        }
```

X9VoiD (approved):

InfinityGhost (changes requested):
I mean, yeah this works but it looks horrible compared to the previous LINQ method. Since the issue is specifically the `HidDevice.CanOpen` property, we should just check that in a separate try/catch block in a private function in the meantime.

BigBrainAFK:
> 
> 
> I mean, yeah this works but it looks horrible compared to the previous LINQ method. Since the issue is specifically the `HidDevice.CanOpen` property, we should just check that in a separate try/catch block in a private function in the meantime.

I seperated out the `device.CanOpen` and reverted back to the first commit style of LINQ so that the console doesnt get spammed with all the exceptions but will still log errors if needed.

InfinityGhost (approved):

PR #975, opened 2021-04-10
Fix Consistency in Various Config Names

Kuuuube:
XP-Pen:
- Many XP-Pen tablets which are a part of the `Star` lineup (`G430`, `G430S`, `G540 Pro`, `G540`, `G640`, `G640 V2`, `G960 V2 (now G960)`) did not include this in their name. The `Star` is referenced everywhere with these tablets by XP-Pen including as a part of their model numbers.
- The `Artist 12 Pro` had it's name backwards and was `Artist Pro 12`.
- The `[Star] G960 V2` does not exist and there is only the `Star G960`. There is no config for the `Star G960` so the `[Star] G960 V2` config is likely meant for the `Star G960` and it has been renamed as such.
- Various capitalization inconsistencies were also fixed.

Huion:
- The `Kamvas Pro 13` and `GT-133` are the same tablet but the name `GT-133` is nearly never used by Huion and using `Kamvas` puts it in line with the other `Kamvas` configs. Although, all of the `Kamvas` configs could be renamed to `GT-xxx` but it is much less confusing to name them by what Huion uses 95% of the time.
- The `Inspiroy Q11K` and `Inspiroy Q11K V2` configs have been renamed to `Q11K` and `Q11K V2`. They should not have `Inspiroy` in the name. `Inspiroy` seems to be something Huion throws onto their tablets nearly at random and they don't refer to any of their tablets using the `Inspiroy` name in any actual documentation or in the official drivers for these tablets. Also, none of the other Huion configs use the `Inspiroy` name even though many do have it mentioned in Huion's marketing.


OpenTabletDriver/Configurations/XP-Pen/Deco 01 v2.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Name": "XP-Pen Deco 01 v2",
+  "Name": "XP-Pen Deco 01 V2",
```
InfinityGhost, line 2: Match casing of this field to the filename.

InfinityGhost (changes requested):

gonX (approved):

InfinityGhost (approved):

PR #976, opened 2021-04-10
Add CTF-430 Config

BK8000L:


OpenTabletDriver/Configurations/Wacom/CTF-430.json
```diff
@@ -0,0 +1,29 @@
+{
+  "Name": "Wacom CTF-430",
+  "DigitizerIdentifiers": [
+    {
+      "Width": 127.6,
+      "Height": 92.8,
+      "MaxX": 5104.0,
+      "MaxY": 3712.0,
```
Kuuuube, line 8: Are you sure these are correct? Have you verified this in OTD debugger?

BK8000L: Untitled
my pen is in the lower right corner of the surface.
this tablet has an aspect ratio of 5:4 and is quite old, it has a low resolution of 1040lpi

Kuuuube: Ok looks good. Needed to ask because according to this the CTF-430 is not 1040LPI. It is actually 1016LPI which goes against wacom's listed specs.

OpenTabletDriver/Configurations/Wacom/CTF-430.json
```diff
@@ -0,0 +1,29 @@
...
+      "Width": 127.6,
+      "Height": 92.8,
+      "MaxX": 5104.0,
+      "MaxY": 3712.0,
+      "MaxPressure": 511,
+      "ActiveReportID": {
+        "Start": 12,
+        "StartInclusive": true,
+        "End": null,
+        "EndInclusive": true
```
Kuuuube, line 14: We put `false` when configs use `null` on reportID to keep things consistent across configs.

OpenTabletDriver/Configurations/Wacom/CTF-430.json
BK8000L: sorry.
"Width": 124.6,
"Height": 90.6,
is correct?

Kuuuube: I think what you have right now is still correct. Would you be able to physically measure it accurately enough to check?

OpenTabletDriver/Configurations/Wacom/CTF-430.json
BK8000L: ok, sorry.

OpenTabletDriver/Configurations/Wacom/CTF-430.json
BK8000L: my measurements are very close to the initial version

OpenTabletDriver/Configurations/Wacom/CTF-430.json
BK8000L: can you fix it?

OpenTabletDriver/Configurations/Wacom/CTF-430.json
Kuuuube: Ok then I think sticking to `"Width": 127.6,"Height": 92.8,` is the right thing to do. It's more likely wacom messed up the LPI than the physical size measurement which matches what you have. Wacom's hover height is never listed as a max hover height just as an average or a guarantee it will go at least that high.

OpenTabletDriver/Configurations/Wacom/CTF-430.json
Kuuuube: I don't have the permissions to edit your pr. You can just push another commit to fix it.

Kuuuube (approved):

InfinityGhost (approved):
I can't find any useful information on this tablet, so I'm taking your word for how well it is supported. Is anything currently nonfunctional on the tablet, such as auxiliary buttons, pen tilt, etc?

BK8000L:
> I can't find any useful information on this tablet, so I'm taking your word for how well it is supported. Is anything currently nonfunctional on the tablet, such as auxiliary buttons, pen tilt, etc?

nothing special about this tablet, everything works well

PR #987, opened 2021-04-13
Fix async element disposal

X9VoiD:
Fixes daemon crash when disabling a previously enabled async element.


OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
```diff
@@ -112,7 +112,7 @@ protected void OnEmit()
         public void Dispose()
         {
             Scheduler?.Dispose();
-            Scheduler = null;
+            scheduler = null;
```
InfinityGhost, line 115: Should instead null check in the property

X9VoiD: Fixed

InfinityGhost (approved):

PR #1013, opened 2021-04-19
Add a page about running OpenTabletDriver in the background to the greeter

MishterKirby:


OpenTabletDriver.UX/Windows/Greeter/Pages/BackgroundingPage.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public BackgroundingPage()
+        {
+            this.Content = new StackedContent
+            {
+                new PaddingSpacerItem(),
+                new Bitmap(App.Logo.WithSize(150, 150)),
+                new StylizedText("Running in the background", SystemFonts.Bold(12), new Padding(0, 0, 0, 8)),
+                "You can run OpenTabletDriver in the background.",
+                "All you have to do is minimize it, and it will stay running in your system tray.",
```
InfinityGhost, line 18: Maybe better wording this way? The "all you have to do" part implies it's an optional feature and not the standard behavior
```suggestion
                "OpenTabletDriver will run in the background when minimized.",
                "You can show the window again from the system tray.",
```

OpenTabletDriver.UX/Windows/Greeter/Pages/BackgroundingPage.cs
```diff
@@ -0,0 +1,23 @@
+using Eto.Drawing;
+using OpenTabletDriver.UX.Attributes;
+using OpenTabletDriver.UX.Controls.Generic;
+
+namespace OpenTabletDriver.UX.Windows.Greeter.Pages
+{
+    [PageName("Running in the background")]
+    public class BackgroundingPage : StylizedPage
```
InfinityGhost, line 8: Maybe we could make this page about the system tray as a whole? We could discuss the options available from within it along with how it runs in the background when minimized.

OpenTabletDriver.UX/Windows/Greeter/StartupGreeterWindow.cs
```diff
@@ -25,6 +25,7 @@ protected override void OnLoadComplete(EventArgs e)
                     new AreaEditorPage(),
                     new BindingPage(),
                     new PluginPage(),
+                    new BackgroundingPage(),
```
InfinityGhost, line 28: Windows/WPF is currently the only platform that supports the system tray correctly, where commands actually invoke as intended so this really should only appear on that platform.

OpenTabletDriver.UX/Windows/Greeter/Pages/BackgroundingPage.cs
Kuuuube: Best wording for this is probably:
```
When minimized, OpenTabletDriver will run in the background.
The window can be restored from the system tray.
```

InfinityGhost (approved):

PR #1018, opened 2021-04-20
Lock plugin manager buttons during plugin install/update/uninstall operation

X9VoiD:
Prevents all issues associated with parallel installs/uninstalls.


InfinityGhost (approved):
Works well, resolves a lot of issues too

PR #1019, opened 2021-04-20
Convert some Forms to a singleton window

X9VoiD:
## Changes

- Only allow a single instance of the following:
  - Plugin Manager
  - Configuration Editor
  - Tablet Debugger

Mainly prevents issues that arises from multiple plugin managers being open

### DIFF
--- a/OpenTabletDriver.UX/Windows/WindowSingleton.cs
+++ b/OpenTabletDriver.UX/Windows/WindowSingleton.cs
@@ -1,32 +1,30 @@
-using System;
 using Eto.Forms;
 
 namespace OpenTabletDriver.UX.Windows
 {
-    public class WindowSingleton<T> where T : Form, new()
+    public class WindowSingleton<T> where T : Window, new()
     {
-        public WindowSingleton()
+        private T window;
+
+        public void Show()
         {
-            RefreshInternalForm();
-        }
+            if (window == null)
+            {
+                window = new T();
+                window.Closed += (_, _) => window = null;
+            }
 
-        private T form;
-        private bool isClosed;
+            switch (window)
+            {
+                case Form:
+                    (window as Form).Show();
+                    break;
+                case Dialog:
+                    (window as Dialog).ShowModal();
+                    break;
+            }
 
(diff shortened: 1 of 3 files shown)


OpenTabletDriver.UX/Windows/WindowSingleton.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using Eto.Forms;
+
+namespace OpenTabletDriver.UX.Windows
+{
+    public class WindowSingleton<T> where T : Form, new()
```
InfinityGhost, line 6: Should use the `Window` class as a constraint instead so this could potentially be used on dialogs too.

OpenTabletDriver.UX/Windows/WindowSingleton.cs
```diff
@@ -0,0 +1,32 @@
...
+            form.Show();
+            form.Focus();
+        }
+
+        private void RefreshInternalForm()
+        {
+            isClosed = false;
+            form = new T();
+            form.Closed += (_, _) => isClosed = true;
+        }
```
InfinityGhost, line 30: We could completely omit the need for the `isClosed` field by nulling out the form when it is closed.
If we do it this way, we won't initialize `form` early (only initialized we ever call this function) and potentially reduces memory usage.

Also, why not just call this `Show()`? The class itself implies its a Singleton so the suffix is kinda pointless IMO
```suggestion
        public void Show()
        {
            if (form == null)
            {
                form = new T();
                form.Closed += (_, _) => form = null;
            }

            form.Show();
            form.Focus();
        }
```

InfinityGhost (changes requested):
Really like the WindowSingleton idea, just a few tweaks and I'll be happy with this.

OpenTabletDriver.UX/Windows/WindowSingleton.cs
X9VoiD: Code cleaned up to this + switch case for Form and Dialog

InfinityGhost (approved):

PR #1025, opened 2021-04-21
Catch binding validation exceptions

X9VoiD:
## Changes

- Fixes crashes encountered during retrieval of binding validation.
- Adds a log entry specific to plugin developers forgetting to use static validator.


InfinityGhost (approved):
Always nice to have exception handling lol

PR #1028, opened 2021-04-21
Ignore OpenTabletDriver.Plugin.dll when loading plugins

gonX:
Loading `OpenTabletDriver.Plugin.dll` inexplicably breaks further plugin loading (at least for the current folder), so by ignoring that we (hopefully) solve the issue where plugins state they've been loaded while they actually don't work (e.g. missing from Filter list).


gonX:
Force pushed some code convention changes requested on Discord

InfinityGhost (approved):

PR #1036, opened 2021-04-22
Lock area dragging to original clicked point

X9VoiD:
## Changes

- Affix ViewModel to cursor's grabbed pixel during area dragging.


InfinityGhost (commented):
Doesn't seem to make an effect on my end, but the code looks good. What does this do?

X9VoiD:
Pre-PR, if you move your cursor outside the viewmodel's bounds (lock to usable area is on), and then move back a little bit, the viewmodel moves relative to the movement of the cursor.

In this PR, the viewmodel will stick to the cursor no matter what. The pixel that you grabbed will stay under your cursor whatever happens. "Lock to usable area" still applies

InfinityGhost:
Now I see what you mean, good change.

InfinityGhost (approved):

PR #1046, opened 2021-04-24
Load MenuBar on daemon connection

adryzz:
# Changes
The menu bar now only loads when the daemon has connected, preventing the user from performing actions that require the daemon connected


InfinityGhost (changes requested):
Looks good, just one minor change.

InfinityGhost (approved):

PR #1054, opened 2021-04-25
Fix linking issues

X9VoiD:
## Changes

- Fix `PostTransform` only using the last element of its group.
- Provide `LinkAll` function to link all parameters and the reverse in `UnlinkAll`.
- Unlink existing elements. Fixes event handler leaks on `this`.
- Fix possible null reference on `PreTransformElements` and `PostTransformElements`.

### Tested by

- [Kuuube](https://discord.com/channels/615607687467761684/628651971267788800/838096546964570142)


InfinityGhost:
Ensure that this does what we intend, iirc I ran into a bug where it wouldn't actually link this way for some reason. Likely unrelated but still worth the check.

InfinityGhost (approved):

PR #1062, opened 2021-04-26
Apply optimization to Wacom parsers

X9VoiD:
Did not apply Unsafe optimizations to some parsers that rely on specific bit operations.

- [x] Depends on #1060


OpenTabletDriver/Vendors/Wacom/IntuosV3ReportParser.cs
```diff
@@ -7,13 +7,20 @@ public class IntuosV3ReportParser : IReportParser<IDeviceReport>
     {
         public virtual IDeviceReport Parse(byte[] data)
         {
-            return data[0] switch
+            if (data.Length < 10)
```
AbstractQbit, line 10: bruh moment
```suggestion
            if (data.Length >= 10)
```

AbstractQbit (approved):

PR #1065, opened 2021-04-28
Add support for VEIKK VK640

TLRS69:
I would really love to use this driver but it does not support my tablet


Mrcubix:
Please take a look at the [compatibility list](https://github.com/OpenTabletDriver/OpenTabletDriver/projects/4) and then make an **issue** not a pull request, This should be closed.

PR #1077, opened 2021-05-01
Fix pen in range detection on async pipeline elements

X9VoiD:
## Changes

- Fix `reportRateMs` always null.


gonX:
This also fixes inter/extra-polators drifting infinitely after liftoff for me, would suggest this being merged before 0.6.x

gonX (approved):

InfinityGhost (approved):

PR #1082, opened 2021-05-02
Fix eraser status report for wacom bamboo

Keltere:
On the wacoom CTH-470 the eraser status was always true even when using the tip.
Credits to @X9VoiD for the solution.

### DIFF
--- a/OpenTabletDriver/Configurations/Wacom/CTH-470.json
+++ b/OpenTabletDriver/Configurations/Wacom/CTH-470.json
@@ -17,7 +17,7 @@
       "ProductID": 222,
       "InputReportLength": 10,
       "OutputReportLength": null,
-      "ReportParser": "OpenTabletDriver.Tablet.TabletReportParser",
+      "ReportParser": "OpenTabletDriver.Vendors.Wacom.BambooCaptureReportParser",
       "FeatureInitReport": "AgI=",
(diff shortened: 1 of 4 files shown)


InfinityGhost (changes requested):
Only the Wacom MTE-450 uses this report parser, nothing else. This shouldn't have made any effect on the CTH-470.

This also breaks the eraser on the MTE-450.

gonX:
For context, a change to the CTH-470 parser was intended (based off a conversation today on Discord) but it does indeed seem that it is absent from this PR

InfinityGhost:
> For context, a change to the CTH-470 parser was intended (based off a conversation today on Discord) but it does indeed seem that it is absent from this PR.
> 
> Good to know that it breaks MTE-450 support, this requires the change to be split into its own parser and then making CTH-470 use that.

I own a CTH-470 as well and I'm unable to reproduce the parent issue either way.

X9VoiD:
Move this to another parser.

Kuuuube (changes requested):
Parser name is misleading. The CTH-470 does not only go by the "Bamboo Capture" name and not all "Bamboo Capture" tablets are CTH-470's.

InfinityGhost:
This has since been resolved in a different PR.

PR #1083, opened 2021-05-02
Add OpenTabletDriver.Analyzers project

X9VoiD:
## Changes

- Implements tablet configuration pre-compilation

The pre-compiled tablet configurations are meant to be used on osu! Lazer and other consumers of `OpenTabletDriver`.
It removes the need to parse any json.

- [x] Implement emitters
- [x] Add tests


OpenTabletDriver/Driver.cs
```diff
@@ -82,6 +82,11 @@ private set
         public DeviceReader<IDeviceReport> TabletReader { private set; get; }
         public DeviceReader<IDeviceReport> AuxReader { private set; get; }
 
+        public static IEnumerable<TabletConfiguration> GetPreCompiledConfigurations()
+        {
+            return CompiledTabletConfig.GetCompiledConfigs();
+        }
```
InfinityGhost, line 88: Do this instead. It would provide a default method of grabbing tablet configurations which can be overridden to provide configurations manually, such as in the desktop implementation.

This also allows us to directly write a virtual `Detect()` method which will run through all configurations and match for each. Making this virtual allows the consumer to handle detection manually as well if desired.

```suggestion
        public virtual IEnumerable<TabletConfiguration> GetConfigurations() => CompiledTabletConfigurations.GetConfigurations();
```

OpenTabletDriver.Tests/TabletConfigurationCompilerTests.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [TestClass]
+    public class TabletConfigurationCompilerTests
+    {
+        [TestMethod]
+        public void GetPreCompiledConfiguration()
+        {
+            var tabletConfig = Driver.GetPreCompiledConfigurations().FirstOrDefault();
+            Assert.IsNotNull(tabletConfig);
+            Assert.IsTrue(!string.IsNullOrEmpty(tabletConfig.Name));
```
InfinityGhost, line 14: Might be a good idea to make sure *all* configurations enumerated are available, instead of just the first one. Wouldn't increase test time much and would make this much more reliable for test coverage.

OpenTabletDriver.Analyzers/TabletConfigurationGenerator.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.CodeAnalysis;
+using Newtonsoft.Json;
+using OpenTabletDriver.Plugin.Tablet;
+
+namespace OpenTabletDriver.Analyzers
+{
+    [Generator]
+    public class TabletConfigurationGenerator : ISourceGenerator
+    {
+        private const string CLASS_NAME = "CompiledTabletConfig";
```
InfinityGhost, line 14:
```suggestion
        private const string CLASS_NAME = "CompiledTabletConfigurations";
```

OpenTabletDriver.Analyzers/TabletConfigurationGenerator.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static class CompiledTabletConfig
+    {
+        public static IEnumerable<TabletConfiguration> GetCompiledConfigs()
+        {
+";
+
+        private const string FOOTER = @"        }
+    }
+}
+";
```
InfinityGhost, line 31: Would like to see this made prettier if at all possible (could use `\n` and string addition), along with using the `CLASS_NAME` constant to avoid a possible mix up.

InfinityGhost (changes requested):

OpenTabletDriver.Analyzers/TabletConfigurationGenerator.cs
X9VoiD: It'll be messier if I did it with push-pop style for indenting, so instead I started the string on a new line.

InfinityGhost:
Will have to wait until omnisharp resolves its issues with analyzers before I'll merge this. I'm still too comfortable with vscode to leave it broken.

InfinityGhost:
Is this even usable since we've switched to the separate project for this instead? I'd like to get a move on this.

X9VoiD:
Can still be used on the OpenTabletDriver.Configurations project. Though to fix slow omnisharp we need to update to .NET 6 for incremental generation feature (to skip code generation when nothing changed on the config files).

jamesbt365:
Well we are on .NET 6 now

InfinityGhost:
> Well we are on .NET 6 now

Might as well, just needs rebased to master now

doubledge-spec:
Update?

InfinityGhost:
This needs bumped to the latest master branch

OpenTabletDriver/OpenTabletDriver.csproj
```diff
@@ -23,10 +23,4 @@
     <ProjectReference Include="..\OpenTabletDriver.Native\OpenTabletDriver.Native.csproj" />
   </ItemGroup>
 
-  <ItemGroup Label="Configurations">
-    <EmbeddedResource Include=".\Configurations\*\*.json">
-      <IncludeInPackage>true</IncludeInPackage>
-    </EmbeddedResource>
-  </ItemGroup>
-
```
X9VoiD, line 31: This points to a non-existent folder, so I removed it.

OpenTabletDriver.Configurations/OpenTabletDriver.Configurations.csproj
```diff
@@ -19,12 +19,13 @@
...
   </ItemGroup>
 
   <ItemGroup Label="Configurations">
-    <EmbeddedResource Include=".\Configurations\*\*.json">
-      <IncludeInPackage>true</IncludeInPackage>
-    </EmbeddedResource>
+    <AdditionalFiles Include=".\Configurations\**\*.json" TabletConfiguration="true" />
   </ItemGroup>
 
+  <Import Project="..\OpenTabletDriver.Analyzers\TabletConfigurationGenerator.props" />
```
X9VoiD, line 29: We're not referencing the generator as a nuget package so the generator prop has to be imported manually.

OpenTabletDriver.Configurations/Configurations/Huion/Kamvas Pro 16 (2.5k).json
```diff
@@ -29,7 +29,7 @@
       "DeviceStrings": {
         "201": "HUION_M214_\\d{6}$"
       },
-      "InitializationStrings": null
+      "InitializationStrings": []
```
X9VoiD, line 32: `InitializationStrings` should never be null according to [`DeviceIdentifier`](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/10a3c07206028f7df5befbdeca6aadb30efe4cb3/OpenTabletDriver/Tablet/DeviceIdentifier.cs), it caused issues to the generator as well.

OpenTabletDriver.Analyzers/TabletConfigurationGenerator.cs
```diff
@@ -0,0 +1,69 @@
...
+            var tabletConfigurationJsons = json
+                .Combine(context.AnalyzerConfigOptionsProvider)
+                .Where(t => IsTabletConfiguration(t.Left, t.Right))
+                .Select((t, _) => ParseConfiguration(t.Left))
+                .Collect();
+
+            context.RegisterSourceOutput(tabletConfigurationJsons, (ctx, jsons) =>
+            {
+                // TODO: automatically get the namespace and class
+                var target = new ProviderTarget("DeviceConfigurationProvider", "OpenTabletDriver.Configurations");
```
X9VoiD, line 28: I don't expect that we'll change the name of that class nor add another target so this is fine for now.

OpenTabletDriver.Analyzers/Formatters/WhitespaceHumanizer.cs
```diff
@@ -0,0 +1,169 @@
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace OpenTabletDriver.Analyzers.Formatters
+{
+    public class WhitespaceHumanizer : CSharpSyntaxRewriter
```
X9VoiD, line 8: This is a fun one, remove this, then see the test output lol.

OpenTabletDriver.Analyzers.Tests/Verifiers/IncrementalGeneratorVerifier.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+
+namespace OpenTabletDriver.Analyzers.Tests.Verifiers
+{
+    public partial class IncrementalGeneratorVerifier<TSourceGenerator>
```
X9VoiD, line 14: This class is designed to be reusable for other projects that use and verify incremental generators (tho it doesn't support testing incremental changes yet, like the one I did in [osu-framework#5562](https://github.com/ppy/osu-framework/pull/5562)). Roslyn-provided verifiers do not support testing metadatas of `AdditionalText` which this one offers.

InfinityGhost (approved):

InfinityGhost:
Something's busted on my end with my rider install and .NET 7 so I'm unable to verify that the analyzers don't break development environments like in the past in vscode's C# extension.

doubledge-spec:
Boid has merge perms :trolley:

PR #1088, opened 2021-05-04
Fix hex-based input boxes

X9VoiD:
## Changes

- Fixed normal hex parsing (was previously always zero, a regression from fixing hex byte arrays)
- Optimized `Allow` function
- Stricter hex input on both normal hex and hex byte array
- Removed duplicate code


OpenTabletDriver.UX/Windows/Configurations/Controls/DeviceIdentifierEditor.cs
```diff
@@ -129,8 +129,8 @@ public DeviceIdentifierEntry()
 
             protected StackLayout layout;
 
-            private MaskedTextBox<int> vendorId, productId;
-            private MaskedTextBox<uint> inputReportLength, outputReportLength;
+            private HexNumberBox vendorId, productId;
+            private UnsignedIntegerNumberBox inputReportLength, outputReportLength;
```
InfinityGhost, line 133: Why was this changed?

OpenTabletDriver.UX/Controls/Generic/Text/Providers/RegexBasedTextProvider.cs
```diff
@@ -0,0 +1,17 @@
+using System.Text.RegularExpressions;
+
+namespace OpenTabletDriver.UX.Controls.Generic.Text.Providers
+{
+    public abstract class RegexBasedTextProvider<T> : MaskedTextProvider<T>
```
InfinityGhost, line 5: Would prefer this class name
```suggestion
    public abstract class RegexTextProvider<T> : MaskedTextProvider<T>
```

OpenTabletDriver.UX/Windows/Configurations/Controls/DeviceIdentifierEditor.cs
X9VoiD: To make it clear what their type really are, instead of finding the `new()` just to know.

InfinityGhost: Reasonable enough

InfinityGhost (approved):

PR #1093, opened 2021-05-08
Run DetectTablets after resuming Windows (from hibernation)

stjeong:
After resuming Windows from hibernation, the tablet can be recognized automatically.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Numerics;
 using System.Threading.Tasks;
 using HidSharp;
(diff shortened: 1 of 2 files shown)


X9VoiD:
While we're at it, I would like to see the same thing for Linux, and MacOS if possible.

adryzz:
Shouldn't platform-specific code be in its own class, and with an abstraction layer?

InfinityGhost:
> Shouldn't platform-specific code be in its own class, and with an abstraction layer?
> Something like `OpenTabletDriver.Daemon.PowerManagement`?

Yeah we should be abstracting this out, ideally into the `OpenTabletDriver.Desktop` lib instead of the daemon.

InfinityGhost:
Superceded by #1290

PR #1098, opened 2021-05-13
Add support for multiple concurrent tablets

InfinityGhost:
# Changes
- Added support for multiple concurrent tablets
- Added profiles for separate tablet settings

## Todo
- [x] UX
- [x] Daemon
- [x] Console


OpenTabletDriver.UX/Windows/DeviceStringReader.cs
```diff
@@ -165,17 +165,8 @@ private async Task<string> SendRequest(string strIndex, string strVid, string st
...
-                        return await App.Driver.Instance.RequestDeviceString(vid, pid, index);
-                    else
-                        return await App.Driver.Instance.RequestDeviceString(index);
-                }
-                catch
-                {
-                    throw;
-                }
+                if (int.TryParse(strVid, out var vid) && int.TryParse(strPid, out var pid))
+                    return await App.Driver.Instance.RequestDeviceString(vid, pid, index);
```
X9VoiD, line 169: This breaks the send string to current tablet when we don't specify vid and pid, but specified index.

```suggestion
                if (int.TryParse(strVid, out var vid) && int.TryParse(strPid, out var pid))
                    return await App.Driver.Instance.RequestDeviceString(vid, pid, index);
                else
                    return await App.Driver.Instance.RequestDeviceString(index);
```

The suggestion might be broken on your implementation, but it still is a regression from master.

InfinityGhost: I'm intentionally removing it in the first place, most of the use case is to get device strings from new devices rather than existing ones.

PR #1117, opened 2021-05-23
Add full init sequence to intuos3 tablets

AbstractQbit:
Seems like this was meant to be added when multi init was implemented, but everyone forgot about it, lol.

Works on ptz-630: https://discord.com/channels/615607687467761684/628651971267788800/836495244484149269
and ptz-930 (source: myself), probably safe to assume all intuos3 tablets share this init.


X9VoiD (approved):

InfinityGhost (approved):

InfinityGhost:
Odds are that this is safe to do anyway because the tablet firmware can reject feature reports.

PR #1127, opened 2021-05-28
Fix empty/null string check in KeyBinding

ashblk:
This check wasn't negated and worked in the opposite way to that intended - virtual keys would get pressed only if the key string was null or empty.


X9VoiD (approved):

InfinityGhost (approved):

InfinityGhost:
Whoops

PR #1128, opened 2021-05-28
Add support for auxilliary buttons on Huion H430P

ashblk:


X9VoiD (changes requested):
Have you already tried using UCLogicReportParser? Looking at what you did with H430 parser it seems that it does the exact same thing as UCLogicReportParser.

X9VoiD (approved):

AbstractQbit (approved):

InfinityGhost (approved):

PR #1164, opened 2021-06-10
New Feature: Hide daemon window

jobs-git:
This file will execute the daemon but hide its window.

Signed off: Dr. James Dominic P. Guana <jdpguana@histark.com>


InfinityGhost:
Odd decision to use a visual basic script, I think this potentially could be done directly in the daemon rather than externally through a script.

jobs-git:
Closing, wrong sign off.

PR #1165, opened 2021-06-10
New Feature: Hide daemon window

jobs-git:
This file will execute the daemon but hide its window.

This way, users can still utilize the console app.

Signed off: Dr. James Dominic P. Guana <jdpguana@histark.com>

### DIFF
--- a/OpenTabletDriver.Daemon/OpenTabletDriver.No.Window.vbs
+++ b/OpenTabletDriver.Daemon/OpenTabletDriver.No.Window.vbs
@@ -1,0 +1,3 @@
+Set Shell = CreateObject("Shell.Application")
+Shell.ShellExecute "OpenTabletDriver.Daemon.exe", , , , 0
+Set Shell = Nothing


InfinityGhost:
Appears to be a duplicate of #1164, mistake by chance?

InfinityGhost:
Scratch that, saw the comment on #1164

jobs-git:
Its possible to program the hidden window internally, but the daemon will lose the console output via CMD.

gonX:
I think this fits the wiki better for users who are interested in such a solution. I'm concerned a vbs file will trigger anti-virus false positives.

InfinityGhost:
> I think this fits better on the wiki, for the few users who are interested in such a solution. I'm concerned a vbs file will trigger anti-virus false positives.

I agree, should be added to [this](https://github.com/OpenTabletDriver/OpenTabletDriver.Web/blob/master/OpenTabletDriver.Web/Views/Wiki/FAQ/Windows.cshtml)

gonX:
Closing for above reasons. Thanks for the PR though, we hope to see more from you in the future!

PR #1204, opened 2021-06-19
Add support for UC-Logic PF1209

### LINKED ISSUE #1197: Add support for UC-Logic PF1209
# Tablet Information

| Name            | Value |
| --------------- | ----- |
| Manufacturer    | UC-Logic |
| Product Name    | PF1209 |
| Width           | 305 mm |
| Height (active)          | 229 mm |
| Pressure Levels | 1024 |
| LPI             | 2000 |

AKA Monoprice #6815

I've figured out a configuration (attached) which works for this tablet, but I think it might need some more work that I'm not sure how to do.
(I think the tablet is supposed to have a strip of "buttons" at the top, but the cover is long gone. There's a ~10mm strip that all registers as the very top of the active area, but I don't see any difference in the reports by tapping in that area.)
[PF1209.json.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6668475/PF1209.json.txt)
[90-opentabletdriver_PF1209.rules.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6668480/90-opentabletdriver_PF1209.rules.txt)

## Diagnostic Information

[diagnostics.json.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6668465/diagnostics.json.txt)

peterfab9845:
Just basic support - binding for tip contact while using the buttons is possible, but not included in this PR since it depends on >2 pen buttons.

Partially resolves #1197


InfinityGhost:
Looking at product images, it appears this tablet supports a mouse? Not really worth mentioning in TABLETS.md but interesting nonetheless.

PR #1231, opened 2021-07-01
Expose tilt to libinput (Linux evdev)

gonX:
Tilt doesn't work on Linux in Artist mode - this PR fixes that.

# How to verify/check for tilt compatibility:

1. Launch `libinput debug-gui`
2. Observe cursor dot stretching (or lack thereof) according to tilt

# Things to consider before merging

- Minimums and maximus are currently considered to be between -127 to 128 at 1 unit per degree, totaling 255 degrees, which probably doesn't make sense as I don't know of any tablets with more than 60 degrees of tilt per direction (totaling ~120 degrees). However, I doubt this causes any degree inaccuracy, given the `resolution` parameter exists.
    - It'd probably be worth adding `TiltResolution` (or similar naming) to tablet configuration files if any tablets have more than 1 unit per degree resolution. Support for this isn't included in the PR.
- ~~Krita does not seem to work with tilt done this way yet, investigating.~~ Fixed by increasing resolution from 1 to ~~100~~ 57

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -45,18 +45,18 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
             var xTilt = new input_absinfo
             {
-                minimum = -127,
-                maximum = 128,
-                resolution = 100
+                minimum = -64,
+                maximum = 63,
+                resolution = 57
             };
             input_absinfo* xTiltPtr = &xTilt;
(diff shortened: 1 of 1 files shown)


gonX:
Increasing `resolution` from 1 to 100 seems to fix the Krita issue. Having some other branches going right now but will update this one ASAP

InfinityGhost:
Is this ready?

gonX:
Yes

gonX:
I'm gonna rescind the ready statement. I think it needs some testing on other tablets, though all Wacoms should work fine.

If someone on a UCLogic-based tablet could test for tilt that would be really swell. `libinput debug-events` is the magic command - and don't forget to enable Artist mode.

gonX:
Tested working correctly on a XP-Pen G960S Plus

gonX:
Ready for merge IMO

InfinityGhost (approved):

PR #1232, opened 2021-07-01
Linux Artist Mode: Refactor Eraser detection logic and expose Proximity to libinput

gonX:
This PR adjusts some code quirks to bring it in line with what programs (particularly Linux' libinput) expects from a tablet driver, most notably changing when we emit the `BTN_TOOL_<name>` event (currently only pen and eraser).

Kernel documentation states:
> `BTN_TOOL_<name>` events must be reported when a stylus or other tool is active on the tablet

Pre-PR: This would only be set when pressure is applied, which deviates from this specification

Post-PR: This is instead set in `SetPosition()`, allowing for programs to detect the different tools before being put to use.

The Eraser variable is currently updated on every report, which can most likely be optimized.

As a part of the PR, proximity and hover distance is added and exposed to libinput as I don't know of another way to correctly report negative `BTN_TOOL_<name>` reports.

This has the side effect of effectively lowering hover distance on some tablets (at least on my GD-0608) as libinput will drop inputs when the device reports out of range.

This feature is from my current tablets perspective advantageous as it provides invalid tilt reports ("stuck" at last value) at extreme hover distances.

If the proximity feature is not desired I can remove it (or split it into its own PR for further analysis), and we'll most likely have to live with negative `BTN_TOOL_<name>` events never being reported to libinput, which I'd argue breaks spec.

Therefore, non-proximity based tablet owners that use Linux Artist Mode are encouraged to test out this PR.

### DIFF
--- a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
+++ b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
@@ -27,7 +27,8 @@ namespace OpenTabletDriver.Desktop.Output
             if (report is ITiltReport tiltReport)
                 VirtualTablet.SetTilt(tiltReport.Tilt);
 
-            VirtualTablet.SetEraser(report is IEraserReport); // FIXME: this is probably inefficient
+            if (report is IEraserReport eraserReport)
+                VirtualTablet.SetEraser(eraserReport.Eraser);
 
             if (report is IProximityReport proximityReport)
                 VirtualTablet.SetProximity(proximityReport.NearProximity, proximityReport.HoverDistance);
--- a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
+++ b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
@@ -26,6 +26,11 @@ namespace OpenTabletDriver.Desktop.Output
 
             if (report is ITiltReport tiltReport)
                 VirtualTablet.SetTilt(tiltReport.Tilt);
+
+            VirtualTablet.SetEraser(report is IEraserReport); // FIXME: this is probably inefficient
+
+            if (report is IProximityReport proximityReport)
+                VirtualTablet.SetProximity(proximityReport.NearProximity, proximityReport.HoverDistance);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


gonX:
forced pushed a rebase to master

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
```diff
@@ -26,6 +26,11 @@ protected override void OnOutput(IDeviceReport report)
 
             if (report is ITiltReport tiltReport)
                 VirtualTablet.SetTilt(tiltReport.Tilt);
+
+            VirtualTablet.SetEraser(report is IEraserReport); // FIXME: this is probably inefficient
```
InfinityGhost, line 30: `IEraserReport` just states that it reports whether or not the eraser is reporting pressure.
```suggestion
            if (report is IEraserReport eraserReport)
                VirtualTablet.SetEraser(eraserReport.Eraser);
```

InfinityGhost (changes requested):
This is my only issue here, however I should note that we're changing how this works in #1210 so this might end up being dropped in favor of that (or could base the PR on that branch once its fixed)

OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
gonX: Are you sure? Looking at report parsers, it seems that it's set whenever you've got the eraser end of the pen in action, and not just when pressure is present.
I can check tomorrow whether a G960S+ reports eraser state in hover or just on pressure as you said - I have no other tablets to test this.

Your change is however correct and does what the intention was

InfinityGhost: > Are you sure? Looking at report parsers, it seems that it's set whenever you've got the eraser end of the pen in action, and not just when pressure is present.

That's what I meant, 0 pressure is still reporting pressure.

gonX:
I've pushed another change that makes evdev/uinput ready for us using `BTN_TOOL_RUBBER` - without the change, eraser end movement would be discarded

InfinityGhost (approved):

PR #1261, opened 2021-07-13
Only show settings based on tablet specifications

InfinityGhost:
- Settings are only shown based on the specifications of the tablet


gonX (approved):
Looks pretty good, and works fine on a Wacom GD-0608-U

PR #1275, opened 2021-07-18
Increase pen button count to 3 for PTH-X60 tablets

### LINKED ISSUE #1257: add support for the third button on the wacom pro 3d pen
## Description
it would be nice to have an option to add a keystroke to a third pen button

InfinityGhost:
- Closes #1257 

Not sure if the report parser actually supports the third button however.


gonX:
> Not sure if the report parser actually supports the third button however.

Neither `OpenTabletDriver.Vendors.Wacom.Intuos` nor `OpenTabletDriver.Vendors.Wacom.IntuosV2` (the 2 used parsers in the affected tablets) have support for 3rd button.

We could make a "guess" on the 3rd button (probably `penByte.IsBitSet(3)` if you follow the logic of the 2 other buttons) and make a PR waiting for testers, or we can wait for people to request its implementation and then they can submit some tablet reports for us.

Satanepigone:
I tested Pro Pen 3D on PTH-660 using _Tablet Debugger_

(Pen Tip is not pressed)
|Condition|Raw Tablet Data (first 2 bytes)|Tablet Report|
|-|-|-|
|no button pressed|10 60|PenButtons:[False False]|
|1st button only|10 62|PenButtons:[True False]|
|2nd button only|10 64|PenButtons:[False True]|
|3nd button only|10 66|PenButtons:[True True]|

It seems Wacom decide to represent the 3rd button as both the 1st and the 2nd button in the raw data

ziwzih:
I would be happy to assist with any testing on this if someone walks me through it. Access to the 3d pen is a must for me, would love to see this implemented!

PR #1290, opened 2021-07-27
Implement sleep detection

X9VoiD:
Supports all platforms, supersedes #1093 and #1112.

## Fixes

- Reruns detection logic when a wake-from-sleep/hibernate event is detected.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -15,6 +15,7 @@ using OpenTabletDriver.Desktop.Reflection;
 using OpenTabletDriver.Desktop.Reflection.Metadata;
 using OpenTabletDriver.Desktop.RPC;
 using OpenTabletDriver.Devices;
+using OpenTabletDriver.Interop;
 using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Logging;
 using OpenTabletDriver.Plugin.Output;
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -44,44 +45,12 @@ public DriverDaemon()
...
-                    Log.Write("Settings", "Recovery complete");
-                    await SetSettings(settings);
-                }
-            }
-            else
+            SleepDetection = new(async () =>
             {
-                await ResetSettings();
-            }
+                Log.Write("Sleep", "Wakey wakey", LogLevel.Debug);
```
InfinityGhost, line 50: Change this lol

OpenTabletDriver/Interop/SleepDetectionThread.cs
```diff
@@ -0,0 +1,69 @@
...
+                var elapsed = stopwatch.Elapsed.TotalSeconds;
+                if (elapsed - prev > 2)
+                    action?.Invoke();
+
+                prev = elapsed;
+
+                await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken);
+            }
+            stopwatch.Stop();
+        }
```
InfinityGhost, line 61: Not huge on using a busy loop to detect if the PC went into sleep mode.

InfinityGhost (changes requested):

OpenTabletDriver.Daemon/DriverDaemon.cs
X9VoiD: Nooooo why ;-;

OpenTabletDriver/Interop/SleepDetectionThread.cs
X9VoiD: It's not a "busy loop" technically and it doesn't "take over a thread" since it's an asynchronous Task. The CPU usage on this is nil.

OpenTabletDriver.Daemon/DriverDaemon.cs
InfinityGhost: Change it.

X9VoiD: Resolved.

InfinityGhost (approved):

PR #1309, opened 2021-08-02
Added new config UC-Logic WP5540U

Sengkey:
Added new config UC-Logic WP5540U

### DIFF
--- a/OpenTabletDriver/Configurations/UC-Logic/WP5540U.json
+++ b/OpenTabletDriver/Configurations/UC-Logic/WP5540U.json
@@ -1,31 +1,37 @@
 {
     "Name": "UC-Logic WP5540U",
-    "DigitizerIdentifiers": [
-      {
-        "Width": 139.7,
-        "Height": 101.6,
-        "MaxX": 10999,
-        "MaxY": 7999,
-        "MaxPressure": 1024,
-        "ActiveReportID": {
-          "Start": 64,
-          "StartInclusive": true,
-          "End": null,
-          "EndInclusive": false
+    "Specifications": {
+        "Digitizer": {
+            "Width": 139.7,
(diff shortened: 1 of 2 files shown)


InfinityGhost (changes requested):
Our configuration format has changed since v0.5, this needs to be updated to match the new format in order to add it to our list. We also require that a newly added tablet to be put into `TABLETS.md` since that's how we track what tablets are supported.

Sengkey:
Hi @InfinityGhost ,
As requested, I have adapted to the new format. 
Note: It is NOT working in v0.5.x.x macOS version. Only the old format earlier is working.

InfinityGhost:
> Note: It is NOT working in v0.5.x.x macOS version. Only the old format earlier is working.

You'd have to use a v0.6 development build in order to use the new configuration format, older versions don't know what to do with the new format.

InfinityGhost (commented):
Everything looks good now, I just want to make sure that the configuration actually still works. If it works on macOS, its pretty safe to say it works on all platforms since macOS is the most difficult to get working.

Sengkey:
Only MacOS is supported. WinUSB implementation required. Linux is untested.

I have updated TABLETS.md

gonX:
This should be targeting the `configs` branch instead, and likely needs a rebase to the new 0.7 config standard.

Sorry for the delay in getting your change merged, I imagine it was simply an oversight that it wasn't seen.

jamesbt365 (changes requested):
This needs rebasing and linting.

jamesbt365:
Author hasn't been seen in about 3 years now and this is more than just out of date.

If anybody has this tablet so this PR can be updated, please mention it so we can update this PR and reverify.

PR #1333, opened 2021-08-12
Remove curly braces from RelativeOutputMode

david-polak:
To match the RelativeOutputMode.


InfinityGhost:
I actually would prefer the opposite of this, where RelativeOutputMode has the curly braces removed. In the past more used to go on here which is why it was there in the first place.

david-polak:
No worries, PR updated :)

gonX (approved):

gonX:
Could you please squish this into 1 commit?

To do this:
while on your branch, run `git rebase -i HEAD^^`, your editor will open - change `pick 910d1e26` to `fixup -C 910d1e26`, save and exit, and then force push the branch

PR #1343, opened 2021-08-14
Add CTH-461 Config

Kuuuube:
Verified by me personally

The touch and aux input are both currently unsupported and require a parser change. I am marking this PR as a draft for now due to that.


Kuuuube:
Unmarking this as a draft. TABLETS.md listing has been changed to Missing Features. 
Aux and touch input should be supported along with all the other wacom tablets using this type of aux and touch parsing (which should be all CTx-x6x). No need to hold back this PR to wait for that to be added.

PR #1350, opened 2021-08-16
Fix #1330

### LINKED ISSUE #1330: MTE-450 ghost clicks
## Description

When lifting the pen from the tablet, the tablet returns some Depth information in the place, where pen pressure is stored.
The Depth information doesn't get above 0x09 for me.
Note: This happens only, when approaching with the pen tip, not the eraser.

Looking at the tablet debugger, I am noticing something. 

When approaching the tablet, with my pen, the `02 XX` value changes, in order, as follows:
Note: The columns stable output/detected pen tip/pen over tablet are just the interpretation, of the raw value.

|no.|value|stable output|detected pen tip|pen over tablet|I see depth info|
|-|----|-----|----|----|----|
|1.|50|no|yes|no|yes
|2.|D0|yes|yes|yes|yes
|3.|9X|yes|no|yes|no

When approaching the tablet from outside the tablet field, the value is just 80/81, with no depth information.

To resolve this one would have to block pen clicks, if  the 'detected pen tip' flag is set. (I think)

Referencing this for information: https://github.com/OpenTabletDriver/OpenTabletDriver/pull/656#issue-556246908

## System Information:
| Name             | Value |
| ---------------- | ----- |
| Operating System         | Windows 10 and Linux 
| OpenTabletDriver Version | v.0.5.3.3 and in osu!lazer
| Tablet                   | MTE-450

C0D3-M4513R:
I am not completely sure, which bit this is. I this should be the 'detected pen tip' bit.

Also tested changes locally with the tablet debugger. Can confirm, that eraser and pen tip still work.

Fixes #1330

### DIFF
--- a/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
+++ b/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using OpenTabletDriver.Plugin.Tablet;
 using OpenTabletDriver.Tablet;
 
@@ -16,7 +17,7 @@ namespace OpenTabletDriver.Vendors.Wacom.Bamboo
 
         private IDeviceReport GetToolReport(byte[] report)
         {
-            if (report[1].IsBitSet(4))
+            if (Unsafe.ReadUnaligned<ulong>(ref report[1])>0)
                 return new BambooTabletReport(report);
 
             return new BambooAuxReport(report);
--- a/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
+++ b/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
@@ -17,7 +17,7 @@ namespace OpenTabletDriver.Vendors.Wacom.Bamboo
 
         private IDeviceReport GetToolReport(byte[] report)
         {
-            if (Unsafe.ReadUnaligned<ulong>(ref report[1])<<16>0)
+            if ( (Unsafe.ReadUnaligned<ulong>(ref report[1]) << 16) > 0 )
                 return new BambooTabletReport(report);
 
             return new BambooAuxReport(report);
--- a/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
+++ b/OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
@@ -17,7 +17,7 @@ namespace OpenTabletDriver.Vendors.Wacom.Bamboo
 
         private IDeviceReport GetToolReport(byte[] report)
         {
-            if (Unsafe.ReadUnaligned<ulong>(ref report[1])>0)
+            if (Unsafe.ReadUnaligned<ulong>(ref report[1])<<16>0)
                 return new BambooTabletReport(report);
 
             return new BambooAuxReport(report);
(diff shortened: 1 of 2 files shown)


C0D3-M4513R:
For a microscopic amount of time pen pressure is still being detected, when there is none.

C0D3-M4513R:
This should make everything work. Haven't had any issues so far, regarding this pr/issue. Played the last 5 days now, without issue.

Kuuuube:
I don't see any issues with this change but I also can't replicate the original issue at all on my MTE-450. Many other parts of the original issue also don't match up with my MTE-450's behavior.

C0D3-M4513R:
@Kuuuube
If you are testing this, make sure, that you do not have Tip Pressure high. I have it set to 0 currently, for testing purposes. 1 should also be fine.
A more exact identifier might be on the back, of the tablet. For me it is: `MTE-450/K(A)`

Kuuuube:
I am testing this in the debugger primarily and to make sure I wasn't missing it in case it went by really fast I recorded inputs. There is no way one of the bad reports would have been sent without this catching it. 

I've checked and my tablet is also labeled `MTE-450/K(A)` with the main model number being `MTE-450`. There is a version named `MTE-450/K (B)` with a main model number being `MTE-450A` and a version with `MTE-450` with no other labeling but I don't have either of those.

I've also checked a bunch of images of other people's MTE-450's and the only possible difference I see between the `MTE-450/K(A)` that would be noted on the tablet is serial numbers starting with 7 vs serial numbers starting with 8. The `MTE-450/K (B)` serial numbers all start with 9 so I think its reasonable to think on the `MTE-450/K(A)` 7 and 8 may be different revisions of something. Mine starts with a 7.

Anyways, I don't intend to block this PR even if I can't verify anything since it causes no issues.

C0D3-M4513R:
The last `06` is getting parsed into the Pen Pressure, for me. (Here it is not, since I am currently running this patch.)

I can confirm, that my serial number also starts with 7.

Example Depth information:
image

Kuuuube:
Found an issue with this change. I can get proper position reports to send when the second byte is `00` if my pen is close enough to the edge of the tablet. Which in your patch are filtered out.

My tablet also doesn't send `D0` or `50` for the second byte ever and pressure never goes up when the pen isn't touching the tablet.

It might be helpful if you join the discord server. https://discord.gg/9bcMaPkVAR

X9VoiD:
@C0D3-M4513R it would be helpful if you upload a tablet data dump using the debugger's record function, and attempt to capture the "bug" in the dump. Describe the exact steps you did to reproduce the bug during the capture.

C0D3-M4513R:
The only thing, that i am doing, is slowly approaching my tablet, but never touching it.

1. Hold the pen tip well above the tablet. Location is irrelevant.
2. Move it down. Not touching the tablet, with the pen makes it easier, to identify the reports.
3. Observe results.

[tablet-data.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/7026083/tablet-data.txt)

OpenTabletDriver/Vendors/Wacom/Bamboo/BambooReportParser.cs
```diff
@@ -17,7 +17,7 @@ public IDeviceReport Parse(byte[] report)
 
         private IDeviceReport GetToolReport(byte[] report)
         {
-            if (report.Skip(1).SkipLast(2).Any(v=>v>0))
+            if (Unsafe.ReadUnaligned<ulong>(ref report[1])>0)
```
C0D3-M4513R, line 20: This still needs a change. We need to ignore the AuxButtons here. If someone pressed a button, position would get parsed.

C0D3-M4513R:
Use #1368

PR #1364, opened 2021-08-21
HS64 V2-re

SnowLire:


InfinityGhost (changes requested):
File extension for the configuration should be `.json`, OpenTabletDriver ignores anything without it.

SnowLire:
Fixed

OpenTabletDriver/Configurations/Huion/HS64 V2.json
```diff
@@ -0,0 +1,43 @@
+{
+  "Name": "Huion HS64 V2",
```
InfinityGhost, line 2: Is this really a V2 tablet or is it another identifier for the tablet?

SnowLire: Afaik it's a different device string so I went with the safe route and make a different config.

InfinityGhost: Makes more sense to just add another identifier IMO, as long as the specifications are the same there's no reason for it to change.

SnowLire: Done.

InfinityGhost:
Migrate to the new configuration directory.

PR #1406, opened 2021-09-07
Retrieve device list through OTD.Daemon

X9VoiD:
- Fixes UX crash when attempting to generate device configuration.

### DIFF
--- a/OpenTabletDriver.UX/Windows/DeviceListDialog.cs
+++ b/OpenTabletDriver.UX/Windows/DeviceListDialog.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Eto.Drawing;
 using Eto.Forms;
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.UX/Windows/DeviceListDialog.cs
```diff
@@ -128,7 +125,14 @@ public DeviceListDialog()
             };
         }
 
-        private ListBox<IDeviceEndpoint> deviceList;
+        public async Task InitializeAsync()
+        {
+            deviceList.Source = (await App.Driver.Instance.GetDevices())
+                .Where(d => d.CanOpen)
+                .ToList();
+        }
```
InfinityGhost, line 133: This method is never invoked. Also should be async void instead, and performed on the UI thread. (`Application.Instance.AsyncInvoke`)

InfinityGhost (changes requested):

PR #1415, opened 2021-09-09
Do not setup device notification hooks on Windows 7 and below

X9VoiD:
## Changes

- Fixes daemon crash on Windows 7 and below.
- Log WinUSB warning when running on Windows 7 and below.

### DIFF
--- a/OpenTabletDriver/Devices/WinUSB/WinUSBRootHub.cs
+++ b/OpenTabletDriver/Devices/WinUSB/WinUSBRootHub.cs
@@ -35,7 +35,16 @@ namespace OpenTabletDriver.Devices.WinUSB
         {
             _callback = NotificationCallback;
             _callbackPin = GCHandle.Alloc(_callback);
-            HookDeviceNotification();
+
+            if (OperatingSystem.IsWindowsVersionAtLeast(6, 2))
+            {
+                HookDeviceNotification();
+            }
+            else
+            {
+                Log.Write(nameof(WinUSBRootHub), $"Windows 7 does not support {nameof(WinUSBRootHub)} hotplug functionality.", LogLevel.Warning);
+                Log.Write(nameof(WinUSBRootHub), $"WinUSB device connections or disconnections won't be detected automatically and requires Daemon restart.", LogLevel.Warning);
+            }
 
             _currentDevices = new List<WinUSBInterface>();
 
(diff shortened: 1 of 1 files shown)


InfinityGhost (approved):

OpenTabletDriver/Devices/WinUSB/WinUSBRootHub.cs
```diff
@@ -35,7 +35,16 @@ public unsafe WinUSBRootHub()
...
-            HookDeviceNotification();
+
+            if (OperatingSystem.IsWindowsVersionAtLeast(6, 2))
+            {
+                HookDeviceNotification();
+            }
+            else
+            {
+                Log.Write(nameof(WinUSBRootHub), $"Windows 7 does not support {nameof(WinUSBRootHub)} hotplug functionality.", LogLevel.Warning);
+                Log.Write(nameof(WinUSBRootHub), $"WinUSB device connections or disconnections won't be detected automatically and require Daemon restart.", LogLevel.Warning);
```
InfinityGhost, line 46: This is implemented in the core library, this won't fly in the osu!framework implementation.

InfinityGhost (changes requested):

OpenTabletDriver/Devices/WinUSB/WinUSBRootHub.cs
X9VoiD: Windows 7 is EOL, it shouldn't be mandatory to fully support Windows 7 anymore. If it's troubling you that o!f might not accept this quirk then talk to them rather than assuming. Anyways, the review is outside of this PR's scope as that implies that we should fix a very special edge case just to support Win7 and below, all of which are EOL.

InfinityGhost: All I want is the message changed to be more generic for implementers.

X9VoiD: Ah gotcha, Daemon name lmao. Almost thought I have to work with ancient APIs again

X9VoiD: Resolved.

InfinityGhost (approved):

Issue #1146, opened 2021-06-06
Add support for Huion 420


3045:
# Tablet Information
| Name                | Value |
| --------------------- | ------- |
| Manufacturer     | Huion
| Product Name   | 420
| Width                 | 176
| Height                | 111
| Pressure Levels | 2048
| LPI                     | 4096

## Diagnostic Information
I'm unsure how to get this information as I am using Linux. Please direct me on how to do so.

## Additional information
I am requesting support for the Huion 420 (not H420...) I figured it won't be too much to add, but you never know since as far as I'm aware, the hardware for the 420 is nearly identical to the H420 (though don't quote me on this, I do think they share the same logic board) with the difference being that the hardware buttons aren't soldered on.

Mrcubix:
good thing you're on linux cause this tablet won't work on windows rn,

as for the config file, open OTD's UI, head to `help > export diagnostic`, zip that and upload it here

Java2107:
I have attached the diagnostic logs to this reply. Sorry for the delay, I was out all day yesterday.
[otd-diag.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/6610705/otd-diag.txt)

X9VoiD:
Duplicate of #331

PR #1419, opened 2021-09-09
Add Huion 420 Configuration

### LINKED ISSUE #331: Support for HUION 420 tablet
## Description

Support for the HUION 420 tablet pls

Mrcubix:
# Verification:

https://discord.com/channels/615607687467761684/885589675333529632/885595268484055171
https://discord.com/channels/615607687467761684/885551214031409173/885580563593969724

# Note:

osu!tablet and 420 are the same tablet internally, therefore i marked both of the mas compatible since i used the same config for the both of them

As for how we determined the height, we made a guess on the lpi using the width and maxX (which is 2085 or roughly 82 lpmm)
and calculated the height using it, it may be incorrect but the advertised LPI is incorrect, tho it seem like it cover the entire tablet so we can assume that it's correct.

# Changes:

- Fixes #331 

# Tablet Owner / Contributor:

- FATKITTY#6988


JotaRata:
I think the specified width/height are wrong since there's a little margin (around 20 mm) in my 420 for which the position is clamped. As noted in #1146 the correct width and height are 176mm and 111mm respectively

jamesbt365:
The position is clamped because theres no more digitizer for the movement to be on, just because there is a signal still attached to the field the tablet generates doesn't mean  anything else actually exists to track where the pen is. The "correct width and height" you linked are the physical dimensions of the tablet itself, not the active area.

PR #1420, opened 2021-09-09
Add XP-Pen Innovator 16 support

X9VoiD:
## Changes

- Adds Innovator 16 configuration
- Adds `XP_PenTabletOverflowReport`

## Verification

https://discord.com/channels/615607687467761684/885638642469580860/885678803358216242


OpenTabletDriver/Vendors/XP_Pen/XP_PenReportParser.cs
```diff
@@ -9,9 +9,11 @@ public IDeviceReport Parse(byte[] report)
         {
             if (report[1].IsBitSet(4))
                 return new XP_PenAuxReport(report);
-            
-            if (report.Length >= 10)
+
+            if (report.Length == 10)
```
gonX, line 13: I'm not well versed on the details of every single XP Pen tablet, but this has the side effect of causing `report.Length == 11` reports to fall through to `TabletReport()` which would previously hit `XP_PenTabletReport`.

If we're 100% sure this isn't an issue, having it the way it is now seems fine for performance reasons, otherwise I'd suggest putting the new if statement prior to this one and reverting this so that it still checks for `>= 10`, (ie. `(x >= 12) {xpptOverflowReport} elseif (x >= 10) {xpptReport} else {tReport}`)

X9VoiD: Resolved

PR #1441, opened 2021-09-16
Add support for .NET 6.0

### LINKED ISSUE #1641: OTD does not build nor run with linux distros that upgraded to .NET 6
## Description

OpenTabletDriver does not have a target framework entry to support .NET 6, and the .NET 6 runtime does not have backwards compatibility with .NET 5.

## Possible fixes

- Both frameworks can be supported by specifying both `.net5.0` and `.net6.0` in `TargetFrameworks`
- Just merge #1441 
- Or, embed the runtime and enable assembly trimming for smaller binary file sizes.

X9VoiD:
and use Directory.Build.props to deduplicate some properties.

- Fixes #1641


Directory.Build.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <PropertyGroup>
+    <OTDVersion>0.6.0.0</OTDVersion>
+    <OTDFrameworkBase>net6.0</OTDFrameworkBase>
```
InfinityGhost, line 4: Can we avoid the OTD acronym within the project?

OpenTabletDriver.Desktop/Reflection/Metadata/PluginMetadata.cs
```diff
@@ -68,7 +68,7 @@ public class PluginMetadata
 
         public static string GetSHA256(Stream stream)
         {
-            using (var sha256 = SHA256Managed.Create())
+            using (var sha256 = System.Security.Cryptography.SHA256.Create())
```
InfinityGhost, line 71: Really shouldn't need the full namespace because its included in an using statement
```suggestion
            using (var sha256 = SHA256.Create())
```

X9VoiD: It conflicts with member "SHA256"

Directory.Build.props
X9VoiD: So... `VersionBase` and `FrameworkBase`?

InfinityGhost: Works for me

OpenTabletDriver.Desktop/Reflection/Metadata/PluginMetadata.cs
InfinityGhost: Good point, forgot we had our own impl

.github/workflows/dotnet.yml
```diff
@@ -49,16 +49,17 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 5.0.x
+          dotnet-version: '6.0'
+          include-prerelease: True
       - name: MacOS Publish
         env:
           runtime: osx-x64
           framework: net5
```
AbstractQbit, line 57: This is now unused here and in `Windows Publish` too. Might also move the runtime thing into options everywhere while at it

gonX:
Since #1556 was merged, you'll need to adjust `./build.sh` to use net6 as well (once you rebase, anyway)

OpenTabletDriver.Plugin/OpenTabletDriver.Plugin.csproj
```diff
@@ -1,17 +1,13 @@
...
   <PropertyGroup Label="Project Properties">
-    <TargetFrameworks>net5</TargetFrameworks>
-    <LangVersion>preview</LangVersion>
+    <TargetFramework>$(FrameworkBase)</TargetFramework>
   </PropertyGroup>
 
   <PropertyGroup Label="NuGet Package Information">
     <PackageId>OpenTabletDriver.Plugin</PackageId>
-    <Version>0.6.0.0</Version>
-    <Authors>InfinityGhost</Authors>
```
gonX, line 11: Removing these 2 lines seems unrelated to changing to .net6. What does this do?

X9VoiD: It's moved to `Directory.Build.props` so that property applies to all projects by default.

RandomExplosion:
If you guys don't mind me asking, what's the status on this pr?
From what i can tell it seems to be ready (i don't know the code base of course tho so there may well be something missing)

X9VoiD:
It's completely ready, it's just that @InfinityGhost wants to wait for osu! framework to also migrate to .NET 6 or atleast add support for it.

zsinx6:
@X9VoiD can you give more details about what is the relation between the osu! framework and this?

jamesbt365:
Our driver is integrated with osu lazer and framework, for compatibility reasons we will move to dotnet 6 when they move

X9VoiD:
Added multi-targeting to `OpenTabletDriver` project so there is no need to worry about compatibility anymore.

InfinityGhost (approved):

PR #1442, opened 2021-09-16
Re-enable self-contained publishing

X9VoiD:
and trim assemblies to cut down file size.

- Depends on #1441

Final file sizes are around 30Mb.


InfinityGhost:
I'd really prefer to not do self contained publishing. The filesize may be smaller now, but we've had multiple issues because of self contained executables.

X9VoiD:
We actually have more problems with non-self-contained assemblies (look at Discord's #help-windows and Gentoo) than self-contained. The only problem we had with self-contained publishing was big file size afaik.

InfinityGhost:
I'm really against building this self contained. I'd rather stick to our current process, the filesize is too much to justify.

X9VoiD:
Bringing this PR back as this is the recommended deployment model for .NET apps and it solves way many headaches for everybody. The file size is only around 30MB, not as big as you're making it out to be.

gonX:
Would be cool if you could squish commits a bit, so we don't get 16 commits for something that should be 1 or 2 commits (not to mention the .NET 6 change which we've already had).

Can someone explain the issues we had when we did use self-contained publishing?

X9VoiD:
No issues other than nix breaking as reported by @InfinityGhost, and file size.

We don't have to worry about type removal since all of those types are plugin types within OTD. Assemblies from OTD doesn't participate with assembly trimming because `IsTrimmable` is [false by default](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options#enable-trimming).

gonX:
Thanks. I'm still requesting a commit squish though ;)

X9VoiD:
Too much reflection on latest master to make this work again.

PR #1453, opened 2021-09-21
Update Kamvas 13 Configuration with M19f Variation

jltbb:
Found while troubleshooting, Index 201: HUION_M19f_200904


InfinityGhost (approved):

X9VoiD (changes requested):
Remove the new identifier, then change the device string into
```
"HUION_(?:M20h|M19f)_\\d{6}$"
```

X9VoiD (approved):

PR #1464, opened 2021-09-25
Decouple Driver capabilities

X9VoiD:
## Changes 
**- Implement core functionalities as required services.**
- Move report parser retrieval logic to the newly-added `IReportParserProvider`
  - Implementations: `ReportParserProvider`, `DesktopReportParserProvider`
- Add `ICompositeDeviceHub`
  - Implementations: `RootHub`
- Add `IDeviceHubsProvider`
  - Implementations: `DeviceHubsProvider`, this is mostly just a dependency of RootHub but decoupling it like this allows for CI-testing against mock devices.
- Add `IDeviceConfigurationProvider`
  - Implementations: `DeviceConfigurationProvider`, `DesktopDeviceConfigurationProvider`
  - This removes the need for loading configurations from within [osu!framework's code](https://github.com/ppy/osu-framework/blob/269c6c63625e56ec269dcf7024fb0e95baf68d36/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs#L68).
  - Also adds the possibility of delivering configuration updates as a separate assembly/nuget, completely decoupling its release cycle from OTD.Core. (for those that want more, this also makes it possible to selectively select configurations from two sources, assembly and external json files)

**- Migrate to Xunit.**

**- Removed "machine-untestable" unit tests to allow for CI testing**

Also enable nullable context on some code.


OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
```diff
@@ -8,6 +8,10 @@
     <LangVersion>preview</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0-rc.1.21451.13" />
+  </ItemGroup>
+
```
InfinityGhost, line 14: Remove this. The daemon project depends on the core project which has this dependency already, we don't want to maintain two different versions unnecessarily.

OpenTabletDriver.Tests/OpenTabletDriver.Tests.csproj
```diff
@@ -2,16 +2,23 @@
...
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="coverlet.collector" Version="3.0.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
```
InfinityGhost, line 22: What's the reason for switching to xunit?

OpenTabletDriver.Plugin/CoreComponents/IDeviceConfigurationProvider.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using OpenTabletDriver.Plugin.Tablet;
+
+namespace OpenTabletDriver.Plugin
+{
+    public interface IDeviceConfigurationProvider
+    {
+        IEnumerable<TabletConfiguration> GetTabletConfigurations();
```
InfinityGhost, line 8:
```suggestion
        IEnumerable<TabletConfiguration> TabletConfigurations { get; }
```

OpenTabletDriver/ComponentProviders/DeviceConfigurationProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+            _configurations = asm.GetManifestResourceNames()
+                .Where(path => path.Contains(".json"))
+                .Select(path => Deserialize(jsonSerializer, asm.GetManifestResourceStream(path)))
+                .ToArray();
+        }
+
+        public IEnumerable<TabletConfiguration> GetTabletConfigurations()
+        {
+            return _configurations;
+        }
```
InfinityGhost, line 28:
```suggestion
    public class DeviceConfigurationProvider : IDeviceConfigurationProvider
    {
        public DeviceConfigurationProvider()
        {
            var asm = typeof(Driver).Assembly;
            var jsonSerializer = new JsonSerializer();

            TabletConfigurations = asm.GetManifestResourceNames()
                .Where(path => path.Contains(".json"))
                .Select(path => Deserialize(jsonSerializer, asm.GetManifestResourceStream(path)))
                .ToArray();
        }

        public IEnumerable<TabletConfiguration> TabletConfigurations { get; }
```

OpenTabletDriver.Plugin/CoreComponents/IDeviceHubsProvider.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using OpenTabletDriver.Plugin.Devices;
+
+namespace OpenTabletDriver.Plugin
+{
+    public interface IDeviceHubsProvider
+    {
+        IEnumerable<IDeviceHub> GetDeviceHubs();
```
InfinityGhost, line 8:
```suggestion
        IEnumerable<IDeviceHub> DeviceHubs { get; }
```

OpenTabletDriver/ComponentProviders/DeviceHubsProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+                    && type.GetCustomAttribute<DeviceHubAttribute>() != null
+                    && (type.GetCustomAttribute<SupportedPlatformAttribute>()?.IsCurrentPlatform ?? true))
+                .Select(type => (IDeviceHub)ActivatorUtilities.CreateInstance(serviceProvider, type))
+                .ToArray();
+        }
+
+        public IEnumerable<IDeviceHub> GetDeviceHubs()
+        {
+            return _deviceHubs;
+        }
```
InfinityGhost, line 29:
```suggestion
        public DeviceHubsProvider(IServiceProvider serviceProvider)
        {
            DeviceHubs = Assembly.GetExecutingAssembly().DefinedTypes
                .Where(type => type.IsAssignableTo(typeof(IDeviceHub))
                    && type.GetCustomAttribute<DeviceHubAttribute>() != null
                    && (type.GetCustomAttribute<SupportedPlatformAttribute>()?.IsCurrentPlatform ?? true))
                .Select(type => (IDeviceHub)ActivatorUtilities.CreateInstance(serviceProvider, type))
                .ToArray();
        }

        public IEnumerable<IDeviceHub> DeviceHubs { get; }
```

OpenTabletDriver.Plugin/CoreComponents/ICompositeDeviceHub.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace OpenTabletDriver.Plugin.Devices
```
InfinityGhost, line 3: Namespace must be updated, I also want to suggest a different name as well.
```suggestion
namespace OpenTabletDriver.Plugin.Components
```

OpenTabletDriver/DriverServiceCollection.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class DriverServiceCollection : ServiceCollection
+    {
+        public static readonly ReadOnlyCollection<ServiceDescriptor> RequiredServices = new(new ServiceDescriptor[]
+        {
+            ServiceDescriptor.Singleton<IReportParserProvider, ReportParserProvider>(),
+            ServiceDescriptor.Singleton<IDeviceHubsProvider, DeviceHubsProvider>(serviceProvider => new DeviceHubsProvider(serviceProvider)),
+            ServiceDescriptor.Singleton<ICompositeDeviceHub, RootHub>(serviceProvider => RootHub.WithProvider(serviceProvider)),
+            ServiceDescriptor.Singleton<IDeviceConfigurationProvider, DeviceConfigurationProvider>()
+        });
```
InfinityGhost, line 21: This is not referenced outside of this class, this should instead just be applied at the constructor.

InfinityGhost (changes requested):

OpenTabletDriver.Tests/OpenTabletDriver.Tests.csproj
X9VoiD: For `[MemberData]`, for removal of class decoration like `[TestClass]` or something, it is to also disallow any form of state on all of the tests in the form of `[Setup]` and `[Cleanup]` etc. Any test state that may affect the run of other tests will become hard to maintain in the future, especially on keeping it CI-testable and keeping track of how a state affects a test if it's not within the test function itself.

Xunit also have helper functions like `Record` for capturing exceptions without assertion and much better `Assert` messages.

OpenTabletDriver/DriverServiceCollection.cs
X9VoiD: Turned it into a private property get.

OpenTabletDriver.Tests/OpenTabletDriver.Tests.csproj
InfinityGhost: Works for me.

PR #1468, opened 2021-09-26
Decouple configs into OpenTabletDriver.Configurations

X9VoiD:
- [ ] Depends on #1464 

## Changes

- Moved OpenTabletDriver/Configurations into a separate assembly `OpenTabletDriver.Configurations` along with `DeviceConfigurationProvider`.
- Desktop: Source configurations from both the assembly and external files. External files take precedence when existing.

## Applications

Through this, we will be able to distribute configuration updates by only updating `OpenTabletDriver.Configurations`, as long as there are no "overrides" existing.


InfinityGhost:
Not sure why this got closed on deletion, should be rebased to master now.

PR #1472, opened 2021-10-01
Add Intuos 1 mouse support (closes #1428)

### LINKED ISSUE #1428: Expand IntuosV1 parser features (fix incorrectly reported pen pressure while hovering)
## Description

IntuosV1 tablets exhibit an odd flickering behavior on the pressure byte while hovering that the current IntuosV1 parser incorrectly reads as pressure. This issue is unique to the pen side of the tablet pen as the eraser side does not exhibit this flickering behavior (as seen in linked video). Hardware issues can be ruled out as the hardware pen status light on the tablet only turns green when actually putting pressure on the pen. 

## Attachments
[Video demonstration of bug](https://youtu.be/DUs7BDtjDU8)
[Tablet data recording](https://github.com/OpenTabletDriver/OpenTabletDriver/files/7147329/gd-0912-u.tablet-data.txt)

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Windows 10 (20H2)
| OpenTabletDriver Version | v0.6.0 pre-release
| Tablet                   | Wacom GD-0912-U

EriDere:
**Summary of changes:**
- Adds a basic mouse parser for Intuos 1 tablets, based on the existing Intuos 3 mouse parser. 
- Updates Intuos 1 tablet configs to add mouse options.

**Current issues/limitations:**
- There is a small chance that this parser change may cause unintended behaviour for other tablets that use the IntuosV1 parser due to the sheer number of tablets that share the same parser. I have no way of testing other models myself, so this should be checked by those who are able.


OpenTabletDriver/Vendors/Wacom/IntuosV1/IntuosV1MouseReport.cs
```diff
@@ -0,0 +1,39 @@
...
+                report[8].IsBitSet(0), // primary
+                report[8].IsBitSet(2), // secondary
+                report[8].IsBitSet(1), // middle
+                report[8].IsBitSet(5), // forward
+                report[8].IsBitSet(4), // backward
+            };
+
+            Scroll = new Vector2
+            {
+                    Y = report[8].IsBitSet(3) && report[6] > 0 ? 1 : report[6] > 0 ? -1 : 0 // ignores analogue mouse wheel info and simply reports whether the mouse wheel is spun upwards or downwards. should eventually be replaced.
```
X9VoiD, line 30:
```suggestion
            var scrollAmount = (report[6] << 2) | ((report[7] >> 6) & 0b11);
            Scroll = new Vector2
            {
                    Y = report[8].IsBitSet(3) ? scrollAmount : -scrollAmount
```

OpenTabletDriver/Vendors/Wacom/IntuosV1/IntuosV1ReportParser.cs
```diff
@@ -22,6 +22,10 @@ private IDeviceReport GetToolReport(byte[] report)
         {
             if (report[0] == 0x10 && report[1] == 0x20)
                 return new DeviceReport(report);
+            if (report[1] == 0xE8 || report[1] == 0xA8)
+                return new IntuosV1MouseReport(report);
+            if (report[1] == 0xEA || report[1] == 0xAA) // ignore secondary mouse reports
+                return null;
```
X9VoiD, line 28: See if this works better.
```suggestion
            if (((report[1] >> 1) & 0x0F) == 0x04)
                return new IntuosV1MouseReport(report);
```

X9VoiD (changes requested):

OpenTabletDriver/Vendors/Wacom/IntuosV1/IntuosV1MouseReport.cs
```diff
@@ -0,0 +1,40 @@
...
+                report[8].IsBitSet(1), // middle
+                report[8].IsBitSet(5), // forward
+                report[8].IsBitSet(4), // backward
+            };
+
+            var scrollAmount = (report[6] << 2) | ((report[7] >> 6) & 0b11);
+            Scroll = new Vector2
+            {
+                    Y = report[8].IsBitSet(3) ? scrollAmount : -scrollAmount
+            };
```
InfinityGhost, line 32: Fix formatting
```suggestion
            Scroll = new Vector2
            {
                Y = report[8].IsBitSet(3) ? scrollAmount : -scrollAmount
            };
```

InfinityGhost (changes requested):

OpenTabletDriver/Vendors/Wacom/IntuosV1/IntuosV1ReportParser.cs
EriDere: Discussed this on Discord the other day, but figured I'd note the same things here for the sake of organization.

This parser change works aside from the mouse sometimes immediately sending and holding whatever left click is bound to when it enters range of the tablet, particularly from the edges of the tablet area. From the Discord conversation, it sounded as if this was a known issue with all tablet tools and is something that is beyond the scope of this particular parser, so if that is confirmed to be the case then this change can be committed.

EriDere:
Went ahead and committed the parser change suggestion I'd been holding out on. This change does still have the issue of sometimes randomly "locking" the pen tip binding while using the mouse (particularly upon entry and exit of the mouse into range of the tablet), but it sounds as if this issue is due to the way OTD handles out of range reports and is entirely out of the scope of this parser unless reports are manually nullified like I originally had in my parser solution (which certainly isn't a clean or optimal implementation). 

But assuming the randomly sticking pen tip binding is due to the suspected out of range report issue, this parser works on all other accounts. It is just a bit buggy and annoying to use with its current functionality when you have pen tip bound to something.

gonX:
Is your last comment perhaps addressed by #1676 ?

EriDere:
It may help solve some issues if functional beyond Artist Mode on Linux (all my testing with this parser has been on Windows), but it in all likelihood won't solve the specific issue here. The parser is currently hard-coded in IntuosV1ReportParser based on certain bytes the tablet gives while using the mouse rather than using the tablet tool system like it likely should be. This is both due to the current Intuos 3 mouse parser using a similar system as well as tablet tools not working properly (or at all) on Windows in general (which may have been fixed since last time this was brought up on Discord; need to check master). But if the tool handling on Windows is fixed and out of range reports are made more resilient against some of the weird states that can occur, I think a rewrite of this (and possibly also the Intuos 3 mouse parser) to use the tablet tool system would both fix all major issues and make the mouse parser more consistent with the parsing of other tablet tools. 

So TLDR, it should help but I think a future rewrite of this parser is the best solution.

EriDere:
I'll leave this pr as-is for the time being since it does match the current Intuos 3 mouse parser, but may switch to draft if Windows tool reports and out of range stuff are confirmed to be fixed soon since I don't think this is the optimal parser solution if those two things get fixed.

gonX:
Any updates on this PR? I'm not sure if we have anything on our end that is being held up here

EriDere:
Closing, as the way this parser is handled isn't optimal, and parser improvements such as #2260 will supercede functionality if expanded to include mouse/tool support.

PR #1519, opened 2021-10-13
Change "Pressure" to "Threshold" on the bindings tab

jamesbt365:
Changed pressure to deadzone in the bindings tab in otd to reduce confusion.

### DIFF
--- a/OpenTabletDriver.Console/Program.Commands.cs
+++ b/OpenTabletDriver.Console/Program.Commands.cs
@@ -83,7 +83,7 @@ namespace OpenTabletDriver.Console
                 var tipBinding = AppInfo.PluginManager.ConstructObject<IBinding>(name);
 
                 p.BindingSettings.TipButton = new PluginSettingStore(tipBinding);
-                p.BindingSettings.TipActivationPressure = threshold;
+                p.BindingSettings.TipActivationDeadzone = threshold;
             });
         }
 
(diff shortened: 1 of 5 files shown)


gonX:
I agree that pressure isn't the best wording, however I think we can do better than deadzone...

Maybe just use the wording "pressure threshold" or simply "threshold" instead of deadzone?

jamesbt365:
Converted to Threshold.

InfinityGhost:
This patch is all that is needed for the migration to work as intended. Existing settings (on master) will not be migrated, however as a development version I don't care that it won't be.
```diff
diff --git a/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs b/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
index 0137a87c..4ae4aa24 100644
--- a/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
+++ b/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
@@ -62,6 +62,7 @@ namespace OpenTabletDriver.Desktop.Migration
                         }
                     )
                 );
+                profile.BindingSettings.TipActivationThreshold = old.TipActivationPressure;
                 profile.BindingSettings.PenButtons = SafeMigrateCollection(old.PenButtons);
                 profile.BindingSettings.AuxButtons = SafeMigrateCollection(old.AuxButtons);
             }
```

jamesbt365:
should be good to go.

jamesbt365:
should be good to go now.

gonX:
> Existing settings (on master) will not be migrated, however as a development version I don't care that it won't be.
What would be needed to make this possible on master? Alternatively, is the user informed of settings being dropped?

InfinityGhost:
> What would be needed to make this possible on master?

We would need some sort of versioning system for settings, which quickly becomes messy.

> Alternatively, is the user informed of settings being dropped?

Nope, it should just be filled with defaults instead. Might be worth doing in another PR.

InfinityGhost (approved):

PR #1568, opened 2021-10-29
Colorize entries in console tab based on LogLevel

Sublimelime:
A small enhancement PR to add code to colorize entries in the Console tab based on their LogLevel severity. Creates greater visual clarity (easy to spot errors quickly) and uses black text on light pastel colors to ensure readability with the exception of Fatal errors, which use white text on a dark red to really stand out.

An image of how it looks with different types of LogLevel messages printed out manually is attached. Open to feedback on the colors, and whether or not the Debug level should be colorized or remain black on white like Info still is.

Console coloration

### DIFF
--- a/OpenTabletDriver.UX/Controls/LogView.cs
+++ b/OpenTabletDriver.UX/Controls/LogView.cs
@@ -77,7 +77,7 @@ namespace OpenTabletDriver.UX.Controls
             };
 
             //Console coloring causes non-accelerated drawing on MacOS, cannot be done without performance hit
-            if (OpenTabletDriver.Interop.SystemInterop.CurrentPlatform != PluginPlatform.MacOS)
+            if (Interop.SystemInterop.CurrentPlatform != PluginPlatform.MacOS)
             {
                 messageList.CellFormatting += (_, entry) =>
                 {
--- a/OpenTabletDriver.UX/Controls/LogView.cs
+++ b/OpenTabletDriver.UX/Controls/LogView.cs
@@ -130,6 +130,7 @@ namespace OpenTabletDriver.UX.Controls
         private readonly GridView<LogMessage> messageList = new GridView<LogMessage>
         {
             AllowMultipleSelection = true,
+            GridLines = GridLines.None,
             Columns =
             {
                 new GridColumn
--- a/OpenTabletDriver.UX/Controls/LogView.cs
+++ b/OpenTabletDriver.UX/Controls/LogView.cs
@@ -99,8 +99,8 @@ namespace OpenTabletDriver.UX.Controls
                         entry.BackgroundColor = Colors.DarkRed;
                         break;
                     default:
-                        entry.ForegroundColor = Colors.Black;
-                        entry.BackgroundColor = Colors.White;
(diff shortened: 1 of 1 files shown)


jamesbt365:
How does this handle with gtk themes?

jamesbt365:
Turns out, they don't.
image

Sublimelime:
Converting to draft until we can make sure this change plays nice with themes and all 3 platforms.

OpenTabletDriver.UX/Controls/LogView.cs
```diff
@@ -76,6 +76,35 @@ public LogView()
...
+                        entry.BackgroundColor = Colors.Pink;
+                        break;
+                    case LogLevel.Fatal:
+                        entry.ForegroundColor = Colors.White;
+                        entry.BackgroundColor = Colors.DarkRed;
+                        break;
+                    default:
+                        entry.ForegroundColor = Colors.Black;
+                        entry.BackgroundColor = Colors.White;
+                        break;
```
InfinityGhost, line 104: Remove this so it will conform with the native theme instead.

InfinityGhost (changes requested):

OpenTabletDriver.UX/Controls/LogView.cs
InfinityGhost, line 104:
```suggestion
                    default:
                        entry.ForegroundColor = SystemColors.ControlText;
                        entry.BackgroundColor = SystemColors.ControlBackground;
                        break;
```

OpenTabletDriver.UX/Controls/LogView.cs
```diff
@@ -76,6 +76,39 @@ public LogView()
                 });
             };
 
+            //Console coloring causes non-accelerated drawing on MacOS, cannot be done without performance hit
+            if (OpenTabletDriver.Interop.SystemInterop.CurrentPlatform != PluginPlatform.MacOS)
```
InfinityGhost, line 80:
```suggestion
            if (Interop.SystemInterop.CurrentPlatform != PluginPlatform.MacOS)
```

Sublimelime: Hah, still getting used to C#'s access and stuff. Didn't know Interop was directly accessible, sorry for making you fix something so simple... ;P

Sublimelime:
WPF weird log borders

@Kuuuube I may have a fix for this if it's still happening after 2afdc47, I'm gonna do some testing today on my windows machine (as I don't get this on linux at all -- seems GTK doesn't show those borders) and see if I can make it look clean on windows too (like the screenshot in the PR message).

Sublimelime:
Drafting once again as there's still unresolved minor issues with windows (though GTK works fine). Current appearance on Windows: 

Capture

Notable problems:

- Selecting rows only highlights the borders in blue, instead of inverting the entire row to white on blue
- Ugly white borders around each row and column, especially noticeable on colored rows

Sublimelime:
Closing this due to unresolved issues, and impending gui rewrite in 0.7.0. Will revisit once rewrite to MAUI is complete, it may be easier or doable with icons instead of changing the background color.

PR #1577, opened 2021-10-31
BindingState: Only invoke IInterruptBinding handler when state has changed

gonX:
This caused some headaches in attempting to implement passthrough aux keys, as the binding handler seems to iterate over all aux keys bindings when any key is pressed, resulting in `IInterruptBinding`s being invoked when they weren't actually intended to be.

Pre-PR:
`IInterruptBinding`s are invoked if any key in the matrix is activated, even if that key itself isn't bound to it; ie if aux binding 1 (b1) has pen passthrough but aux binding 3 (b3) does not - clicking b3 still activated the pen passthrough feature.

Post-PR
`IInterruptBinding`s are only invoked when state has changed, likely resulting in more accurate results.

As a side note, if you're using a tablet which supports multiple held keys simultaneously, it is possible to get non-Passthrough buttons "stuck" in an activated state (using the example above, holding b3 while pressing b1, and then keeping b3 held while releasing b1 will result in the system thinking b3 is active until another pen passthrough aux report is parsed again)

This can also happen pre-PR on non-Aux `IInterruptBinding`s so this PR brings it more in line with `IStateBinding`s.

### DIFF
--- a/OpenTabletDriver.Desktop/Binding/BindingState.cs
+++ b/OpenTabletDriver.Desktop/Binding/BindingState.cs
@@ -23,7 +23,8 @@ namespace OpenTabletDriver.Desktop.Binding
 
             if (Binding is IInterruptBinding interruptBinding)
             {
-                interruptBinding.Invoke(tablet, report);
+                if ((newState && !PreviousState) || (!newState && PreviousState))
+                    interruptBinding.Invoke(tablet, report);
             }
 
             PreviousState = newState;
--- a/OpenTabletDriver.Desktop/Binding/BindingState.cs
+++ b/OpenTabletDriver.Desktop/Binding/BindingState.cs
@@ -23,7 +23,7 @@ namespace OpenTabletDriver.Desktop.Binding
 
             if (Binding is IInterruptBinding interruptBinding)
             {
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Desktop/Binding/BindingState.cs
```diff
@@ -23,7 +23,8 @@ public virtual void Invoke(TabletReference tablet, IDeviceReport report, bool ne
 
             if (Binding is IInterruptBinding interruptBinding)
             {
-                interruptBinding.Invoke(tablet, report);
+                if ((newState && !PreviousState) || (!newState && PreviousState))
```
X9VoiD, line 26: Won't something like this be enough?
```suggestion
                if (newState != PreviousState)
```

X9VoiD (changes requested):

gonX:
Actually, one breaking change I noticed is that this breaks pen passthrough tip bindings resulting in pressure not being continually updated.

gonX:
Closing for now since that is such a breaking change for preparation that might not even be necessary.

gonX:
Something something to link #1578 to this

PR #1587, opened 2021-11-01
Always show most of UI even with no tablets detected

### LINKED ISSUE #1405: Console not visible when no tablet is detected
## Description
Due to the way tablet selector is nested inside the UI pages, it becomes hidden when no tablet is detected. This makes it *hard* for users to determine why there is no tablet detected.

Sublimelime:
This PR changes the structure of the UI to allow for showing the tab bar even when no tablets are connected.

Previously, the entire UI save for the "File, Tablets, etc" menubar would get blocked out by a placeholder, preventing the user from accessing the console tab or plugin tabs. This made it pretty hard to determine why, having to open up daemon.log manually and all that.

Combined with #1564, this allows us to show the user errors in the console tab even if we can't detect a tablet, this lets error messages like "OEM drivers are installed" and the like to be visible. No more having to pop up a message box. With #1564 it even will flip the tab to console upon such an error, so everything works smoothly.

This is how it looks now:

New-No-Tablets-Screen

As you can see, only the Output tab's content itself is placeholdered. The aux, pen and mouse binding tabs are also hidden when no tablets are detected, as expected. Everything else, the plugin tabs (can still un/install plugins or configure them as that doesn't require a tablet), the profile switcher, the console tab, etc is all still visible and usable.

As soon as a tablet is successfully detected, the output tab is returned to normal and the aforementioned binding tabs are shown as fitting for the tablet (no change from current behavior).

Testing - Tested on GTK (Linux) by starting the UI without a tablet plugged in, plugging in a tablet, then unplugging the tablet. All works. 

Fixes #1405

### DIFF
--- a/OpenTabletDriver.UX/Controls/ControlPanel.cs
+++ b/OpenTabletDriver.UX/Controls/ControlPanel.cs
@@ -62,11 +62,6 @@ namespace OpenTabletDriver.UX.Controls
                 }
             };
 
-            placeholder = new Placeholder
-            {
-                Text = "No tablets detected."
-            };
-
             outputModeEditor.ProfileBinding.Bind(ProfileBinding);
             penBindingEditor.ProfileBinding.Bind(ProfileBinding);
             auxBindingEditor.ProfileBinding.Bind(ProfileBinding);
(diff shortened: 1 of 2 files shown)


InfinityGhost:
Closing as #1625 covers this in scope while also covering the case that the other tabs could potentially still be configured when no tablet is selected. Unintentionally made my changes without realizing this PR was already here, thanks for the contribution nonetheless!

PR #1599, opened 2021-11-03
Fix crash on rapid redetect/change how 'searching' detect logs are logged

### LINKED ISSUE #1498: Invoking detection repeatedly causes UI deadlock
## Description

If you make otd detect too quickly through either the menu or by spamming control + d then otd freezes and crashes, upon reloading the ux doesn't respond at all, no clicks are registered even though the ux hasn't crashed.

otd

It gets stuck on the select tablet and mode but otd doesn't work even if clicked.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System   Linux      |
| OpenTabletDriver Version 0.6 pre release (arch aur the day before the config folder move)|
| Tablet         XP PEN Star G640 Rev A          |

Sublimelime:
Idea/theory PR to change how "searching for tablet" messages are logged, in order to make them less spammy. Opened as draft for now to gather feedback on both the code and the "fix path" (do we wanna fix #1498 this way?).

From my testing, this change dramatically cuts down the amount of log messages necessary when searching for tablets, (which appears to be the cause of #1498), which, from my testing involving spamming Ctrl-D to try and cause the crash outlined in #1498, means this PR: 

Fixes #1498 

At the cost of a bit of a loss in info. Now instead of listing each tablet it just lists by manufacturer (which is pulled from the first word of the config's Name). This is overall cleaner IMO, and the driver still indicates the exact tablet(s) it's found when it finds them, or logs "no tablets detected" if none are found.

```
[Detect:Info]	Searching for Artisul tablets...
[Detect:Info]	Searching for Gaomon tablets...
[Detect:Info]	Searching for Genius tablets...
[Detect:Info]	Searching for Huion tablets...
[Detect:Info]	Searching for Parblo tablets...
[Detect:Info]	Searching for UC-Logic tablets...
[Detect:Info]	Searching for UGTABLET tablets...
[Detect:Info]	Searching for VEIKK tablets...
[Detect:Info]	Searching for Wacom tablets...
[Detect:Info]	Searching for XenceLabs tablets...
[Detect:Info]	Searching for XP-Pen tablets...
[Detect:Info]	No tablets found.
```

And that's it. Simple, clean, will scale well as we add more tablets, and doesn't overwhelm the logger.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -128,7 +128,8 @@ namespace OpenTabletDriver.Daemon
 
         public async Task<IEnumerable<TabletReference>> DetectTablets()
         {
-            Driver.Detect();
+            if (!Driver.Detect())
+                Log.Write("Detect", "No tablets found.");
 
             foreach (var tablet in Driver.InputDevices)
             {
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -39,10 +39,18 @@ namespace OpenTabletDriver
         public virtual bool Detect()
         {
             bool success = false;
+            string lastManufacturer = "";
 
             InputDevices.Clear();
             foreach (var config in _deviceConfigurationProvider.TabletConfigurations)
             {
+                string currentManufacturer = config.Name.Substring(0, config.Name.IndexOf(' '));
+                if (lastManufacturer != currentManufacturer)
+                {
+                    lastManufacturer = currentManufacturer;
+                    Log.Write("Detect", $"Searching for {currentManufacturer} tablets...");
+                }
+
                 if (Match(config) is InputDeviceTree tree)
                 {
                     success = true;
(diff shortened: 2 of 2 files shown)


jamesbt365:
The old logging method is still needed as a way to see which configs are being loaded, imply saying what brand isn't enough.

Sublimelime:
This is a concern (and what I mean by a reduction to the amount of info available), but the way the configs are loaded means that if it loads a config from one manufacturer, it should load all tablets from that same manufacturer. As long as all manufacturers are listed, things are running normally.

The only instance I could see where this could be an issue is missing a config for a specific tablet, which shouldn't happen so long as it's been compiled properly.

jamesbt365:
Giving people configs to add and them adding then and seeing if they load is important, so this approach isn't exactly perfect and reduces the amount of information that is nice to have by a lot.

Sublimelime:
Closing as this solution isn't viable - Breaks giving configs to people and ensuring they load (log messages printed to exported diags). Still, here for posterity.

PR #1601, opened 2021-11-04
Add unsupported features on Administator warning on startup

jamesbt365:
Added a short description on what won't work when starting OpenTabletDriver when running with administrative permissions.


Sublimelime:
This should be done for Linux (if run with sudo) and for MacOS too if possible.

InfinityGhost:
> This should be done for Linux (if run with sudo) and for MacOS too if possible.

Just doesn't work as root on linux unless you're using a desktop session as root (which of course you should never do)

gonX:
> Just doesn't work as root on linux unless you're using a desktop session as root (which of course you should never do)

Even then, it should still work and I don't really think we should go out of our way to stop people from running OTD as root, if someone for some reason has a niche use case of a single-user embedded OS or something along those lines.

gonX:
PR LGTM BTW

InfinityGhost (approved):

PR #1610, opened 2021-11-06
Adjust sizes of various UX elements

jamesbt365:
Adjusted the size of the Main OpenTabletDriver window making it slightly larger.
Adjusted the size of the plugin manager so that Plugin names don't get cut off.
Adjusted the size of the Tablet Debugger so that the Tablet Visualizer is larger.

### DIFF
--- a/OpenTabletDriver.UX/DesktopForm.cs
+++ b/OpenTabletDriver.UX/DesktopForm.cs
@@ -22,7 +22,7 @@ namespace OpenTabletDriver.UX
         }
 
         private bool platformInit;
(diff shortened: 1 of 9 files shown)


X9VoiD (changes requested):
These default values are beyond what 1366x768 could display.

jamesbt365:
> These default values are beyond what 1366x768 could display.

I have restored the height change back to 760 to stop cutoff on 1366x768 resolution displays.

jamesbt365:
I FUCEK ID UP AGAIN

PR #1619, opened 2021-11-09
Fix logging UI deadlock

InfinityGhost:
The blocking collection causes the UI to freeze when logging from separate threads, this returns it to a concurrent collection to mitigate this.

### DIFF
--- a/OpenTabletDriver.UX/Controls/LogView.cs
+++ b/OpenTabletDriver.UX/Controls/LogView.cs
@@ -88,7 +88,7 @@ namespace OpenTabletDriver.UX.Controls
                 where message is LogMessage
                 select message;
 
-            foreach (var message in currentMessages)
+            foreach (var message in currentMessages.Take(150))
                 AddMessage(message);
 
             App.Driver.AddConnectionHook(i => i.Message += (sender, message) => AddMessage(message));
@@ -104,7 +104,7 @@ namespace OpenTabletDriver.UX.Controls
                     HeaderText = "Time",
                     DataCell = new TextBoxCell
                     {
-                        Binding = Eto.Forms.Binding.Property<LogMessage, string>(m => m.Time.ToLongTimeString())
+                        Binding = Binding.Property<LogMessage, string>(m => m.Time.ToLongTimeString())
                     }
                 },
                 new GridColumn
@@ -112,7 +112,7 @@ namespace OpenTabletDriver.UX.Controls
                     HeaderText = "Level",
                     DataCell = new TextBoxCell
                     {
-                        Binding = Eto.Forms.Binding.Property<LogMessage, string>(m => m.Level.GetName())
+                        Binding = Binding.Property<LogMessage, string>(m => m.Level.GetName())
                     }
                 },
                 new GridColumn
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.UX/Controls/LogView.cs
```diff
@@ -88,7 +88,7 @@ private async Task InitializeAsync()
                 where message is LogMessage
                 select message;
 
-            foreach (var message in currentMessages)
+            foreach (var message in currentMessages.Take(150))
```
X9VoiD, line 91: This creates a bug where *only the first 150 logs* will be taken from Daemon, causing inconsistent logging. It will look as if there's a hole in the logs after that first 150.

OpenTabletDriver.UX/Tools/LogDataStore.cs
```diff
@@ -10,7 +10,7 @@ namespace OpenTabletDriver.UX.Tools
 {
     public class LogDataStore : INotifyCollectionChanged, IEnumerable<LogMessage>
     {
-        private readonly BlockingCollection<LogMessage> messages = new BlockingCollection<LogMessage>(500);
+        private readonly ConcurrentQueue<LogMessage> messages = new ConcurrentQueue<LogMessage>();
```
X9VoiD, line 13: It's not trimming the logs so this will grow indefinitely, causing memory hogging problems.

X9VoiD (changes requested):
Currently there are two copies of the same log, one in Daemon, the other in UX. Eto.Forms also allocates separately for displaying the logs itself. All of this is *never trimmed*, and will grow indefinitely until host memory is exhausted. Should do something about that.

OpenTabletDriver.UX/Tools/LogDataStore.cs
InfinityGhost: Fixed.

OpenTabletDriver.UX/Controls/LogView.cs
InfinityGhost: Resolved.

OpenTabletDriver.UX/Tools/LogDataStore.cs
```diff
@@ -10,7 +11,20 @@ namespace OpenTabletDriver.UX.Tools
...
-        private readonly BlockingCollection<LogMessage> messages = new BlockingCollection<LogMessage>(500);
+        public LogDataStore(IEnumerable<LogMessage> currentMessages)
+        {
+            var messageList = currentMessages.ToArray();
+            if (messageList.Length > MAX_NUM_MESSAGES)
+            {
+                var startIndex = messageList.Length - MAX_NUM_MESSAGES;
+                messageList = messageList[startIndex..^0];
+            }
+            messages = new ConcurrentQueue<LogMessage>(messageList);
```
X9VoiD, line 22:
```suggestion
            messages = new ConcurrentQueue<LogMessage>(currentMessages.TakeLast(MAX_NUM_MESSAGES));
```

X9VoiD (changes requested):

OpenTabletDriver.UX/Tools/LogDataStore.cs
InfinityGhost: Honestly no idea how I forgot that function existed

Sublimelime:
This only effects the display of the log in the console tab, right? Exporting diags and viewing daemon.log to see the full console output still works fine?

X9VoiD:
Yes

X9VoiD (approved):

PR #1626, opened 2021-11-11
Tablet detection logging improvements

InfinityGhost:
- Make tablet search messages verbose
- Log tablet search begin and end (when no tablets are found)

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -63,7 +63,7 @@ namespace OpenTabletDriver
 
         protected virtual InputDeviceTree? Match(TabletConfiguration config)
         {
-            Log.Write("Detect", $"Searching for tablet '{config.Name}'");
+            Log.Debug("Detect", $"Searching for tablet '{config.Name}'");
             try
             {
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -58,12 +60,17 @@ public virtual bool Detect()
 
             TabletsChanged?.Invoke(this, Tablets);
 
+            if (!success)
+            {
+                Log.Write("Detect", "No tablets were detected.", LogLevel.Error);
```
Sublimelime, line 65: This shouldn't be Error as that'll make OTD switch to the console tab now - Unnecessary as we already show that no tablets were detected in other parts of the UI. (Window header, the output tab itself, etc) 

Error should be reserved for actual problems, we don't wanna error every time someone unplugs their tablet.

InfinityGhost: Info instead then? Reasonable to me.

Sublimelime: Yeah, afaik it is/was info at some point.

gonX (approved):

PR #1629, opened 2021-11-11
Add Built-in Self-Updater for Windows and MacOS

### LINKED ISSUE #1400: Automatic Updates via UI
## Description

Automatically prompt for updates on specific platform.

InfinityGhost:
# Changes

- [x] Added an updater service
- [x] Added unit tests for the updater implementations
- [x] Added UI for updating
- [x] Added prompt for updates via notification

## Issues

- Closes #1400


X9VoiD (changes requested):
The updater itself should not depend on Octokit for retrieving releases. That should be the responsibility of the implementation, so it should look something like:
```cs
public interface IUpdateProvider
{
    Version GetCurrentVersion(); // can be removed, just here to explicitly state that getting current version is the responsibility of IUpdateProvider
    // bool HasUpdate();
    bool HasUpdate(Version currentVersion);
    Stream<byte> GetLatestBinary(); // Binary retrieval is up to the provider
    void InstallBinary(Stream<byte> binary);
    
}
```
to allow mocking of updates with something like:
```cs
public class MockUpdateProvider
{
    public static readonly Version UpdatedVersion = new Version("1.0.0");
    public void SetupMockEnvironment()
    {
        // Put mock binary, or just text specifying mock version.
    }
    public Version GetCurrentVersion()
    {
        // Read mock binary, or text.
    }
    public bool HasUpdate(Version currentVersion) => currentVersion < UpdatedVersion;
    public Stream<byte> GetLatestBinary()
    {
        // Since we know how the mock binary/text is made
        // it allows us to also mock the updated binary/text.
    }
    ...
}
```
which makes it possible to unit test everything about the update process in a controlled manner (like testing integration with UX once UX is reworked for testability).

I'd also like to see the unit test refactored to follow [Unit-testing best practices](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices), specifically the article about [arranging tests](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#arranging-your-tests) and [avoiding branches within unit tests](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests). 

Current unit test performs three separate processes: download; extract; and install; none of which were asserted explicitly. It assumes that
> As long as the test does not raise an exception then it's correct.

which is dangerous.

You may always just use an `abstract class OctokitUpdateProvider : IUpdateProvider` to make an update provider that uses octokit for the base implementation.

InfinityGhost:
> The updater itself should not depend on Octokit for retrieving releases. That should be the responsibility of the implementation, so it should look something like:

I'm not wasting time wrapping the entirety of the Release and ReleaseAsset class for something that will never be changed, nor is it something that will be the source of error. We're testing the actual install process here, not the download.

X9VoiD (approved):

OpenTabletDriver.Desktop/Updater/IUpdater.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading.Tasks;
+using Octokit;
+
+namespace OpenTabletDriver.Desktop.Updater
+{
+    public interface IUpdater
+    {
+        Task<bool> CheckForUpdates();
+        Task<Release> GetRelease();
```
X9VoiD, line 9: There's no need for the hardcoded `octokit` dependency in this interface if you only needed the tag. This could just be `Version` which is sufficient for your usage of this function, or put it as an optional `out` param of `CheckForUpdates`.

X9VoiD (changes requested):

OpenTabletDriver.Desktop/Updater/IUpdater.cs
InfinityGhost: I thought this was made clear, it needs the tag, the download path, eventually the release contents, its just easier to use the Release class that already exists rather than wrapping it.

PR #1639, opened 2021-11-13
build.sh: Do not continue building if an error occurs

gonX:
This can save the user some time if something errors out early, and it also has the nice side effect of not scrolling build errors out of view.

Pre-PR:
`./build.sh` builds everything unconditionally, even if an earlier project failed

Post-PR:
`./build.sh` exits ASAP with an error code if any of the `dotnet publish` commands fail


gonX:
This doesn't work as expected. Nothing to see here.

gonX:
Superseded by #1677

PR #1668, opened 2021-11-26
Overhaul version display, make build date available in diag and about screen

### LINKED ISSUE #1221: Action builds should use a more accurate version than last tag
Currently, action builds display the last tagged version (currently 0.5.3.1), which confuses users (and helpers).

The AUR package gets around this by specifying a version prefix and suffix (trimmed for conciseness):
```sh
    PREFIX=$(git describe --long --tags | sed 's/-.*//;s/v//')
    SUFFIX=$(git describe --long --tags | sed 's/^[^-]*-//;s/\([^-]*-g\)/r\1/;s/-/./g')

    dotnet publish        OpenTabletDriver.Daemon   \
        /p:VersionPrefix="$PREFIX"                  

    dotnet publish        OpenTabletDriver.Console  \
        --version-suffix  "$SUFFIX"                 \
        /p:VersionPrefix="$PREFIX"                  

    dotnet publish        OpenTabletDriver.UX.Gtk   \
        --version-suffix  "$SUFFIX"                 \
        /p:VersionPrefix="$PREFIX"                  
```

Sublimelime:
This is a small PR to change how we handle OTD's version. It adds a build date attribute to OpenTabletDriver.Plugin, which can be accessed in-code through Plugin's assembly to obtain the date the binary was built.

This has been added in two places, the about screen and the export of a diagnostic.

Note that the version display in the window header, "OpenTabletDriver v0.6.0.0", hasn't changed.

The exported diag now looks like this....

```
{
  "App Version": "OpenTabletDriver v0.6.0.0, built on 11/27/2021",
  "Operating System": {
    "Platform": 4,
    "ServicePack": "",

.....
```

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup>
-    <VersionBase>0.6.0.0</VersionBase>
+    <VersionPrefix>0.6.0.0</VersionPrefix>
+    <VersionSuffix>dev</VersionSuffix>
+    <SourceRevisionId>Built on $([System.DateTime]::UtcNow.ToString("d"))</SourceRevisionId>
     <FrameworkBase>net6.0</FrameworkBase>
-    <LangVersion>preview</LangVersion>
-    <VersionPrefix>$(VersionBase)</VersionPrefix>
   </PropertyGroup>
 
   <PropertyGroup Label="Package Information">
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,9 +1,9 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>0.6.0.0</VersionPrefix>
-    <VersionSuffix>dev</VersionSuffix>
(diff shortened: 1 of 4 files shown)


Sublimelime:
I *cannot* figure this out... I might be missing something but for some reason this works except for the Native and Plugin sub-projects.... which fail compilation with:

```
/OpenTabletDriver/OpenTabletDriver.Plugin/obj/Release/net6.0/OpenTabletDriver.Plugin.AssemblyInfo.cs(13,12): error CS0246: The type or namespace name 'BuildDateAttributeAttribute' could not be found (are you missing a using directive or an assembly reference?) 
```

Same for OpenTabletDriver.Native. I've pushed this for now despite the failing build to get feedback/guidance (I will squash and fix this later, don't worry). I may be doing this entirely wrong (as I've not worked with attributes much, and I'm just going off of @X9VoiD's suggestion) but I feel this shouldn't be erroring.

X9VoiD:
Seems like MSBuild is already appending `Attribute` for us so...

```
<AssemblyAttribute Include="BuildDate">
```

PR #1671, opened 2021-11-28
Improve Wayland/swaywm compatibility

gonX:
This PR introduces a couple of commits, all of which try to standardize OTD's Linux Artist mode behavior to behave more like an upstreamed kernel tablet driver.

## 87446a9:
This batches the different types of reports attached to a report (position, pressure, tilt, etc) and only sends a final `Sync()` when all attributes have been resolved - this is done by adding a `Finish()` function to `IVirtualTablet`.

Most (all?) tablets batch all properties in the same report anyway, so this is likely the desired behavior that clients expect.

## 199ab4b:
This drops the square 1:1 aspect ratio tablet and instead uses a virtual tablet sized similarly to `EvdevAbsolutePointer`, ie. the same as the virtual desktop, but with a bump in resolution to keep subpixel accuracy intact. The change makes lpmm symmetric on both axises, assuming the display and tablet aspect ratios are matched.

## bf86cce:
Wayland (or at least Sway) had some serious cursor spazzing going on in non-Xwayland windows while in Artist mode. This commit significantly increases the reported lpmm of the virtual tablet, which in turn makes the cursor significantly more stable. E.g. my desktop of 4480x1440 results in a virtual tablet size of 44.8x14.4mm, rather than a size that is a few dozen meters on each axis.

Without this specific commit, even with the other changes, the cursor is uncontrollably jittery with sway.

# Summed up:
**Pre-PR**:
- Artist mode sends a `SYN_REPORT` _per property_ which likely introduces overhead, and also might confuse applications
- Artist mode virtual tablet is same size on both axises, resulting in an asymmetric lpmm.
- Artist mode is unusable in non-Xwayland windows with Sway, resulting in the cursor going literally everywhere.

**Post-PR**:
- Artist mode only sends 1 `SYN_REPORT` per tablet report handled.
- Artist mode virtual tablet dimensions are instead sized based off OTD's virtual screen coordinates, which should keep lpmm symmetric in most instances.
- Artist mode's behavior is significantly more stable with Sway, and likely other Wayland compositors - however the latter has been untested.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -10,7 +10,6 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 {
     public class EvdevVirtualTablet : EvdevVirtualMouse, IAbsolutePointer, IVirtualTablet
     {
-        private const int Max = 1 << 28;
         private static readonly EventCode[] BUTTONS =
         {
             EventCode.BTN_STYLUS,
@@ -19,6 +18,7 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
         };
 
         private Vector2 ScreenScale = new Vector2(DesktopInterop.VirtualScreen.Width, DesktopInterop.VirtualScreen.Height);
+        private int Resolution = 1000; // subpixels per screen pixel
         private bool IsEraser = false;
         private bool Proximity = true;
 
@@ -31,7 +31,7 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
             var xAbs = new input_absinfo
             {
-                maximum = Max,
+                maximum = (int)(DesktopInterop.VirtualScreen.Width * Resolution),
                 resolution = 100000
             };
             input_absinfo* xPtr = &xAbs;
@@ -39,7 +39,7 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
             var yAbs = new input_absinfo
             {
-                maximum = Max,
+                maximum = (int)(DesktopInterop.VirtualScreen.Height * Resolution),
                 resolution = 100000
             };
             input_absinfo* yPtr = &yAbs;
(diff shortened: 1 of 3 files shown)


OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
```diff
@@ -19,6 +18,7 @@ public class EvdevVirtualTablet : EvdevVirtualMouse, IAbsolutePointer, IVirtualT
         };
 
         private Vector2 ScreenScale = new Vector2(DesktopInterop.VirtualScreen.Width, DesktopInterop.VirtualScreen.Height);
+        private int Resolution = 1000; // subpixels per screen pixel
```
Sublimelime, line 21:
```suggestion
        private const int Resolution = 1000; // subpixels per screen pixel
```

OpenTabletDriver.Plugin/Platform/Pointer/IVirtualTablet.cs
```diff
@@ -9,5 +9,6 @@ public interface IVirtualTablet : IAbsolutePointer
         void SetButtonState(uint button, bool active);
         void SetEraser(bool isEraser);
         void SetProximity(bool proximity, uint distance);
+        void Finish();
```
InfinityGhost, line 12: I would prefer to continue to use sync to describe this functionality.
```suggestion
        void Sync();
```

OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
```diff
@@ -132,6 +127,10 @@ public void SetProximity(bool proximity, uint distance)
         {
             Proximity = proximity;
             Device.Write(EventType.EV_ABS, EventCode.ABS_DISTANCE, (int)distance);
+        }
+
+        public void Finish()
```
Sublimelime, line 132:
```suggestion
        public void Finish() => Device.Sync();
```

Cleaner to use => if it's just a one liner

(annoying that you can't do a multi-line suggest, of course remove the curly brackets below)

InfinityGhost (changes requested):
Everything looks good, just terminology changes.

Sublimelime (changes requested):

OpenTabletDriver.Plugin/Platform/Pointer/IVirtualTablet.cs
gonX: Ok

OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
gonX: Ok and made `Resolution` uppercase since its now a const

OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
gonX: I forgor :skull: Ty and added

Sublimelime (approved):

PR #1672, opened 2021-11-28
Add BuildDate attribute to track date of build

### LINKED ISSUE #1221: Action builds should use a more accurate version than last tag

Sublimelime:
This is a small PR to add a build date attribute to OpenTabletDriver.Plugin, which can be accessed in-code through Plugin's assembly to obtain the date the binary was built. 

Fixes #1221 

This has been added in two places, the about screen and the export of a diagnostic.

Note that the version display in the window header, "OpenTabletDriver v0.6.0.0", hasn't changed.

The exported diag now looks like this....

```
{
  "App Version": "OpenTabletDriver v0.6.0.0, built on 11/27/2021",
  "Operating System": {
    "Platform": 4,
    "ServicePack": "",

.....
```


InfinityGhost:
This doesn't necessarily close #1221 however, we really want the git tag to be able to pinpoint a build. The method described in that issue is probably the best way to handle that issue, but this is a welcome addition anyway.

Sublimelime:
([Relevant discord convo](https://discord.com/channels/615607687467761684/628651971267788800/914359719957766145), putting this here for record keeping)

TLDR: PR is good but we should also include the git commit SHA in the InformationalVersion, to be done with build scripts, something like:

`/p:SourceRevisionId=$(git rev-parse --short HEAD)`

InfinityGhost (changes requested):
Add the revision hash to all action builds.

.github/workflows/dotnet.yml
```diff
@@ -19,8 +19,8 @@ jobs:
       - name: Linux Publish
         env:
           runtime: linux-x64
-          options: --configuration Release -p:DebugType=embedded --no-self-contained -p:SourceRevisionId=$(git rev-parse --short HEAD)
         run: |
+          $options=@('--configuration Release', '-p:DebugType=embedded', '--no-self-contained', '-p:SourceRevisionId=$(git rev-parse --short HEAD)')
```
X9VoiD: Revert all the future changes to just this.

and

```
options="--configuration Release -p:DebugType=embedded --no-self-contained -p:SourceRevisionId=$(git rev-parse --short HEAD)"
```

.github/workflows/dotnet.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       - name: MacOS Publish
         env:
           runtime: osx-x64
-          options: --configuration Release -p:DebugType=embedded -p:PublishTrimmed=false --no-self-contained -p:SourceRevisionId=$(git rev-parse --short HEAD)
         run: |
+          $options=@('--configuration Release', '-p:DebugType=embedded', '-p:PublishTrimmed=false', '--no-self-contained', '-p:SourceRevisionId=$(git rev-parse --short HEAD)')
```
X9VoiD: Same for here

X9VoiD (changes requested):
Shell expansion won't work with single quotes, shell expansion also don't work within github's `env`, and using arrays when we're not addressing them as an array just makes it look complicated than it should be.

OpenTabletDriver.UX/App.cs
```diff
@@ -83,7 +84,7 @@ public Settings Settings
             ProgramDescription = "Open source, cross-platform tablet configurator",
             WebsiteLabel = "OpenTabletDriver GitHub Repository",
             Website = new Uri(@"https://github.com/OpenTabletDriver/OpenTabletDriver"),
-            Version = $"v{Version}",
+            Version = $"v{Version}, Built on {BuildDate}",
```
X9VoiD, line 87:
```suggestion
            Version = $"v{Version}, built on {BuildDate}",
```
But this in itself is a weird change for me, build date is not part of a version. The date that a software is built does not indicate the version, nor change its behaviour. We can have the same "version" of a software built on different dates.

This should be moved out to another entry in diagnostics, and removed from UX.

X9VoiD (changes requested):

OpenTabletDriver.UX/App.cs
Sublimelime: > This should be moved out to another entry in diagnostics, and removed from UX.

You're right, as we've added the git hash to the version we don't really need to have the build date in the UX, just the exported diag. I'll do that.

X9VoiD (approved):

build.sh
```diff
@@ -7,7 +7,8 @@
 runtime=${1:-linux-x64}
 shift
 
-options=(--configuration='Release' --framework='net6.0' --self-contained='false' --output='./bin' /p:SuppressNETCoreSdkPreviewMessage=true /p:PublishTrimmed=false --runtime=$runtime)
+options=(--configuration='Release' --framework='net6.0' --self-contained='false' --output='./bin' \
+/p:SuppressNETCoreSdkPreviewMessage=true /p:PublishTrimmed=false --runtime=$runtime -p:SourceRevisionId=$(git rev-parse --short HEAD))
```
gonX, line 11:
```suggestion
    /p:SuppressNETCoreSdkPreviewMessage=true /p:PublishTrimmed=false --runtime=$runtime -p:SourceRevisionId=$(git rev-parse --short HEAD))
```
It's more obvious its continuing last line this way.

OpenTabletDriver.UX/App.cs
```diff
@@ -65,7 +66,6 @@ public static void Run(string platform, string[] args)
 
         public const string FaqUrl = "https://opentabletdriver.net/Wiki";
         public static readonly string Version = Assembly.GetEntryAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
-
```
gonX, line 68: Does this try to accommodate some sort of styling preference we have or is this a mistake?

gonX (changes requested):
Just that one thing in `build.sh` then it's okay.

OpenTabletDriver.UX/App.cs
Sublimelime: It was a mistake, likely caused by the several failed commits that I squashed. I'll undo the changes to App.cs with the commit that fixes the indentation in build.sh, one moment....

gonX (approved):

InfinityGhost (approved):

PR #1675, opened 2021-11-28
Improve unloadable plugin exception logging

gonX:
Plugin development/debugging is harder than it needs to be, especially when porting to the new API.
This PR increases the verbosity when plugins are unable to be loaded, which will make it more obvious why plugins aren't loading.

## Pre-PR:
- Console only tells us that the plugin could not be loaded, and that it is likely out of date

## Post-PR:
- Console now also tells us the `HRESULT` as well giving us the exception message and stacktrace at the debug info level.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -82,7 +82,7 @@ namespace OpenTabletDriver.Daemon
 
         public Task WriteMessage(LogMessage message)
         {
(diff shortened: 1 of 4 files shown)


X9VoiD (approved):
Windows always logs something about failing to load Eto, but it's outside of the scope of this PR.

InfinityGhost (approved):

gonX:
New changes look fine if it effectively displays the same thing.

Issue #1579, opened 2021-10-31
Linux Artist Mode incorrectly tells libinput about tool proximity on tablets that don't implement IProximityReport


gonX:
## Description
How to test:

Firstly make sure you're running in Linux Artist Mode and you're using a tablet that does not implement `IProximityReport`s (check your parser)

1. Run `libinput debug-tablet`
2. Bring a pen in range, then take it out of range again.
3. Observe that `BTN_TOOL_PEN` or `BTN_TOOL_RUBBER` buttons remains active despite being out of range.

If no buttons show up, restart OTD and make sure `libinput debug-tablet` is running before bringing the pen in range.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Linux
| OpenTabletDriver Version | 267a322
| Tablet                   | XP-Pen Star G960S Plus

Sublimelime:
Seems related to issue #1533

gonX:
It's not the same, but it is similar in that they're hardware detects that we have to work around.

XP Pen does have out of range detection (look for 0xC0 in `report[1]`), so we likely just need to reimplement proximity and/or split `IProximityReport` into 2 separate things, `IInRangeReport` and `IDistanceReport`  (or similar naming), so we can set InProximity as long as we're not encountering 0xC0.

gonX:
#1676 handles this for the tablets that are covered in the PR (currently intuosv1 and xppen parser). All parsers needs to implement `OutOfRangeReport` (assuming the PR is accepted) to close this bug successfully, or we need to synthesize an `OutOfRangeReport` for tablets that do not have `OutOfRangeReport`s implemented - it's best we don't synthesize them on tablets that have them.

gonX:
If this can't be solved properly in time for 0.6.x I am also willing to look at alternatives where we simply add a timer to Linux Artist Mode which does the reset when report rate drops below a certain threshold.

Ideally the output mode should be aware of whether a tablet/parser can report this in the first place - if this isn't accounted for we at worst lift buttons (and pressure) when we don't intend to.

InfinityGhost:
> If this can't be solved properly in time for 0.6.x I am also willing to look at alternatives where we simply add a timer to Linux Artist Mode which does the reset when report rate drops below a certain threshold.

[...]

Issue #1533, opened 2021-10-18
Mouse pointer no longer can interact with the computer if you lift the stylus away from the surface of the tablet while holding down a button on the stylus


K4LCIFER:
## Description
If you touch the stylus to the surface of the tablet, hold down one of the stylus's buttons, and then lift the stylus away from the surface of the tablet (while keeping the button pressed down), until the stylus is out of range of the tablet, you can observe that the mouse pointer will no longer work (as in it will move around, but it will not be able to click on anything/interact with anything) until the stylus is brought back into range of the tablet without the button pressed down.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Arch Linux |
| OpenTabletDriver Version | 0.5.3.3-1|
| Tablet                   | XP-Pen Artist Pro 12|

Sublimelime:
For additional clarification:

- What happens if you bring the stylus within range again while still holding the pen button? Does it still work fine? What if you keep holding as you bring it closer, then release the pen button while in range?
- Does unplugging the tablet after causing the bug allow the mouse to work again?
- Does re-detecting the tablet via the menu allow the mouse to work again?

K4LCIFER:
@Sublimelime unfortunately, `<ctrl> + d` interferes with other keybindings for me.

K4LCIFER:
> this has somehow a deep smell of hardware issue, but just to make sure, check if the second byte change in tablet debugger when your pen goes out of range with one of the buttons held

The second byte changes from `A2` to `C0` when the pen moves out of range.

gonX:
#1676 should handle this for your tablet if you use Linux Artist Mode.

PR #1676, opened 2021-11-29
Handle out of range reports (in Linux Artist Mode)

gonX:
This PR addresses a few weak points on some edge cases with Linux' Artist mode, like buttons remaining clicked when pen leaves the area.

We have a thread on Out of Range reports on the Discord ([thread link](https://discord.com/channels/615607687467761684/907410063394758696/907410064720158740)) and the general consensus that this was desirable.

This is an initial implementation which as least handles Out of Range reports for Linux' Artist mode on a few tablets, but seems reasonable to extend to the rest of the project eventually.

There is a chance it might be interesting to also have this as an interface (somehow), so that plugins (or internals) can know whether a parser will send out of range reports via this method or not. Guidance is appreciated.

## 9260d28d3c9e229fe1ac6bc04bb5a17ae16c447f:
This adds and implements `OutOfRange` reports in the IntuosV1 parser (tested working on a GD-0608) **There is a chance this might not work on other IntuosV1-parsed tablet configs**

## 1baecbfa00a400cc00c177841786fa569e644633:
This adds and implements `ResetButtons()` which is called whenever a report is an `OutOfRange` report. `ResetButtons()` resets pressure, touch, tool state (eraser/pen), tip and barrel buttons.

---

Pre-PR:
- States (buttons, tool mode) could often be left hanging when the pen loses connection to the tablet unexpectedly, such as the pen laying down on the tablet, resulting in stuff like both eraser and pen ends being detected simultaneously.

Post-PR:
- IntuosV1-parsed tablets seems far more resilient against weird internal OTD/libinput states from happening
- Other parsers can easily implement OutOfRange reporting with typically 2 lines of code, enabling them to be more resilient as well.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Wacom/IntuosV1/IntuosV1ReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Wacom/IntuosV1/IntuosV1ReportParser.cs
@@ -21,6 +21,8 @@ namespace OpenTabletDriver.Configurations.Parsers.Wacom.IntuosV1
         {
             if (report[0] == 0x10 && report[1] == 0x20)
                 return new DeviceReport(report);
+            if (report[1] == 0x80)
+                return new OutOfRangeReport(report);
             if (report[1].IsBitSet(5))
                 return new IntuosV1TabletReport(report);
             else if (report[1] == 0xC2)
--- a/OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenReportParser.cs
@@ -6,6 +6,9 @@ namespace OpenTabletDriver.Configurations.Parsers.XP_Pen
     {
         public IDeviceReport Parse(byte[] report)
         {
+            if (report[1] == 0xC0)
+                return new OutOfRangeReport(report);
+
             if (report[1].IsBitSet(4))
                 return new XP_PenAuxReport(report);
 
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -129,6 +129,21 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
             Device.Write(EventType.EV_ABS, EventCode.ABS_DISTANCE, (int)distance);
         }
 
+        public void ResetButtons()
+        {
+            // Zero out everything except position and tilt
+            Device.Write(EventType.EV_KEY, EventCode.BTN_TOOL_RUBBER, 0);
+            Device.Write(EventType.EV_KEY, EventCode.BTN_TOOL_PEN, 0);
+            Device.Write(EventType.EV_KEY, EventCode.BTN_TOUCH, 0);
+            Device.Write(EventType.EV_ABS, EventCode.ABS_PRESSURE, 0);
+            Device.Write(EventType.EV_KEY, EventCode.BTN_STYLUS, 0);
+            Device.Write(EventType.EV_KEY, EventCode.BTN_STYLUS2, 0);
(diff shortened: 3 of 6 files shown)


Sublimelime:
This is to fix #1579, right? According [to infinity](https://github.com/OpenTabletDriver/OpenTabletDriver/issues/1533#issuecomment-946304634) Wacom tablets could benefit from this as well as XP-Pen ones (#1533).

In fact if this works we should see if any other tablets could benefit from this.

AbstractQbit:
The point of `IProximityReport` was to expose data related to proximity in regualar reports, for things like limiting hover height (useful for relative mode). Many somewhat recent Wacom tablets still report position past `NearProximity` range, so this was not meant as an out of range indicator. I'd suggest leaving `IProximityReport` as it is and instead making an `IOutOfRangeReport` for actual out of range reports sent by the tablet.

gonX:
Force pushed rebase from master. Is there a specific reason this is targeting configs now? It depends on functionality from `master`.

gonX:
I've added OutOfRange reports for XP-Pen as well. This makes multi tablet behavior a bit more sane.

gonX:
I've pushed a change that moves `OutOfRange` into `Plugin.Tablet` instead.

gonX:
I've removed the `IProximityReport` change so that the change is effectively only adding `OutOfRangeReport` now.

gonX:
Is there a particular use case for `IOutOfRangeReport` with a `bool OutOfRange` or is just `OutOfRangeReport` good enough? Example:

```diff
diff --git a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
index b90e2469..d1627f28 100644
--- a/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
+++ b/OpenTabletDriver.Desktop/Output/LinuxArtistMode.cs
@@ -33,7 +33,7 @@ namespace OpenTabletDriver.Desktop.Output
             if (report is IProximityReport proximityReport)
                 VirtualTablet.SetProximity(proximityReport.NearProximity, proximityReport.HoverDistance);
 
-            if (report is OutOfRangeReport)
+            if (report is IOutOfRangeReport oorReport && oorReport.OutOfRange)
                 VirtualTablet.ResetButtons();
 
             VirtualTablet.Sync();
diff --git a/OpenTabletDriver.Plugin/Tablet/IOutOfRangeReport.cs b/OpenTabletDriver.Plugin/Tablet/IOutOfRangeReport.cs
new file mode 100644
index 00000000..a5762098
--- /dev/null
+++ b/OpenTabletDriver.Plugin/Tablet/IOutOfRangeReport.cs
@@ -0,0 +1,7 @@
+namespace OpenTabletDriver.Plugin.Tablet
+{
+    public interface IOutOfRangeReport : IDeviceReport
+    {
+        bool OutOfRange { set; get; }
+    }
+}
diff --git a/OpenTabletDriver.Plugin/Tablet/OutOfRangeReport.cs b/OpenTabletDriver.Plugin/Tablet/OutOfRangeReport.cs
index 2a94bed2..689a5fff 100644
--- a/OpenTabletDriver.Plugin/Tablet/OutOfRangeReport.cs
+++ b/OpenTabletDriver.Plugin/Tablet/OutOfRangeReport.cs
@@ -2,13 +2,15 @@ using OpenTabletDriver.Plugin.Tablet;
 
 namespace OpenTabletDriver.Plugin.Tablet
 {
-    public struct OutOfRangeReport : IDeviceReport
+    public struct OutOfRangeReport : IOutOfRangeReport, IDeviceReport
     {
         public OutOfRangeReport(byte[] report)
         {
             Raw = report;
+            OutOfRange = true;
         }
 
         public byte[] Raw { set; get; }
+        public bool OutOfRange { set; get; }
     }
 }
```

gonX:
Rebased

gonX:
Also removed breaking label since the PR is no longer changing `IProximityReport`. I don't think adding a function to `IVirtualTablet` qualifies as breaking.

InfinityGhost (approved):

PR #1677, opened 2021-11-29
build.sh: Exit immediately if a compilation error occurs

gonX:
This PR makes the build script exit immediately if an error occurs during compilation, so that it e.g. doesn't uselessly compile console and UX if Daemon can't compile.

Pre-PR:
- The entire build script would run regardless of whether any issues occured, resulting in confusing output, and wasting the user/developers time.

Post-PR:
- The build script does not emit 'Build finished' unless all 3 `dotnet publish` commands succeeded, and it stops as soon as possible.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,12 +13,12 @@ echo "Building OpenTabletDriver with runtime $runtime."
 mkdir -p ./bin
 
 echo -e "\nBuilding Daemon...\n"
-dotnet publish OpenTabletDriver.Daemon ${options[@]} $@
+dotnet publish OpenTabletDriver.Daemon ${options[@]} $@ || exit 1
 
 echo -e "\nBuilding Console...\n"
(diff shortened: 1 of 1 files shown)


Sublimelime:
Generally to follow convention ([see here](https://tldp.org/LDP/abs/html/exitcodes.html)), we *should* use exit codes 3-6 for our stuff (1 is general errors, and 2 is for misuse of shell builtins).

Though this is a turbo-nitpick and almost certainly won't matter.

gonX:
> Generally to follow convention ([see here](https://tldp.org/LDP/abs/html/exitcodes.html)), we _should_ use exit codes 3-6 for our stuff (1 is general errors, and 2 is for misuse of shell builtins).

I thought about this and if there is more push for it I'll do it. From experience I've never heard anyone actually obey exit code 2, just that you should use exit codes between 1-63 inclusive.

InfinityGhost (approved):

Issue #1627, opened 2021-11-11
Tablet replugging throws ArgumentOutOfRangeException


InfinityGhost:
## Description

Replugging or unplugging a device can cause this exception to be thrown.
```
Exception in Gtk# callback delegate
  Note: Applications can use GLib.ExceptionManager.UnhandledException to handle the exception.
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
   at System.Linq.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
   at System.Linq.Enumerable.ElementAt[TSource](IEnumerable`1 source, Int32 index)
   at Eto.EnumerableChangedHandler`2.InternalElementAt(Int32 index)
   at Eto.CollectionChangedHandler`2.ElementAt(Int32 index)
   at Eto.GtkSharp.Forms.Controls.GtkEnumerableModel`1.GetValue(TreeIter iter, Int32 col, Value& val)
   at Gtk.TreeModelAdapter.GetValue_cb(IntPtr inst, IntPtr iter, Int32 column, IntPtr value)
   at GLib.ExceptionManager.RaiseUnhandledException(Exception e, Boolean is_terminal)
   at Gtk.TreeModelAdapter.GetValue_cb(IntPtr inst, IntPtr iter, Int32 column, IntPtr value)
   at Gtk.Application.Run()
   at Eto.GtkSharp.Forms.ApplicationHandler.Run()
   at Eto.Forms.Application.Run(Form mainForm)
   at OpenTabletDriver.UX.App.Run(String platform, String[] args) in /repos/OpenTabletDriver/OpenTabletDriver.UX/App.cs:line 61
   at OpenTabletDriver.UX.Gtk.Program.Main(String[] args) in /repos/OpenTabletDriver/OpenTabletDriver.UX.Gtk/Program.cs:line 10
```

gonX:
Can reproduce simply by unplugging the tablet if it was plugged in during that UI session.

Example script that guarantees a crash:
```sh
echo 0 | sudo tee /sys/bus/usb/devices/5-4/authorized
echo 1 | sudo tee /sys/bus/usb/devices/5-4/authorized 
echo 0 | sudo tee /sys/bus/usb/devices/5-4/authorized
```
replace `5-4` with your endpoint (check `lsusb`)

Issue #1422, opened 2021-09-10
Filter settings on UX are not updated when changing tablets


X9VoiD:
## Description
Settings of filters as displayed on UX are not updated on following scenarios:

- Unplugging current tablet and plugging in a different one.
- Normally switching tablet focus.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| OpenTabletDriver Version | 0.6.0-prerelease

PR #1678, opened 2021-11-29
Synchronize all daemon events to UI task scheduler at event source

InfinityGhost:


gonX:
Would this affect #1422?

InfinityGhost:
> Would this affect #1422?

It shouldn't, that's a binding issue

InfinityGhost:
Unlinking this PR from #1627, that appears to be an upstream issue.

PR #1682, opened 2021-12-01
Expose tool serial and tool ID to libinput 

gonX:
This allows libinput (and clients) to handle inputs on a per-tool basis, giving a similar experience to the stock Wacom driver.

Relevant reading: [libinput docs on Tracking unique tools](https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#tracking-unique-tools)

Pre-PR:
- libinput sees tool- and serial ID `0` for _any_ tool

Post-PR:
- libinput sees actual tool- and serial ID of the tool, if the tablet supports it. Otherwise it falls back to reporting 0.

image

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -72,8 +72,9 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
             input_absinfo* yTiltPtr = &yTilt;
             Device.EnableCustomCode(EventType.EV_ABS, EventCode.ABS_TILT_Y, (IntPtr)yTiltPtr);
 
+            var toolId = new input_absinfo();
             Device.EnableType(EventType.EV_MSC);
-            Device.EnableCustomCode(EventType.EV_ABS, EventCode.ABS_MISC, (IntPtr)yTiltPtr); // tool ID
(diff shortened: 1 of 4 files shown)


OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
```diff
@@ -70,6 +72,10 @@ public unsafe EvdevVirtualTablet()
             input_absinfo* yTiltPtr = &yTilt;
             Device.EnableCustomCode(EventType.EV_ABS, EventCode.ABS_TILT_Y, (IntPtr)yTiltPtr);
 
+            Device.EnableType(EventType.EV_MSC);
+            Device.EnableCustomCode(EventType.EV_ABS, EventCode.ABS_MISC, (IntPtr)yTiltPtr); // tool ID
```
InfinityGhost, line 76: This is pointing at the tilt pointer, is that intentional?

InfinityGhost: Would like this at least to be commented for clarity, it seems pointless otherwise.

gonX: I'll just give it a new struct, that should be more obvious

gonX:
This will be superseded in an upcoming PR

PR #1685, opened 2021-12-03
[GTK] Show a message box for unhandled exceptions

InfinityGhost:


Sublimelime (approved):
Could this be done for Windows and MacOS as well?

jamesbt365:
This is more to show in linux which error has happened. Don't really see how this specifically matters for other platforms. Maybe something in another pr if you can think of anything?

InfinityGhost:
It's certainly doable (via `AppDomain.CurrentDomain.UnhandledException`) however on GTK it has to be done this way.

Sublimelime:
Would it even be worth doing though, I haven't done a ton of work with the windows/mac side of things. Do those two crash "cleaner"?

InfinityGhost:
There's really no downside to adding it. GTKSharp has its own exception manager that bypasses CurrentDomain somehow, I'm not sure if WPF or mac does this.

PR #1692, opened 2021-12-05
Introduce less intrusive environment variables in diagnostics

### LINKED ISSUE #1635: Amount of exported environment variables in diagnostics might be too extreme
## Description
OTD diagnostics reports every single environment variable set on the system, which can be problematic privacy-wise, or at worst leak a users password or an API key.

Ideally we should only report the necessary environment variables, or at least make exporting of environment variables optional, since most issues can be resolved without having to see the users environment variables.

This issue can be used to track which environment variables we're definitely interested in. For now, I can think of 2:
- `DISPLAY` 
- `WAYLAND_DISPLAY`

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Any
| OpenTabletDriver Version | 35ae3e6

InfinityGhost:
Closes #1635

### DIFF
--- a/OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
+++ b/OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+using OpenTabletDriver.Desktop.Interop;
+using OpenTabletDriver.Plugin;
+
+namespace OpenTabletDriver.Desktop.Diagnostics
+{
+    public class EnvironmentDictionary : Dictionary<string, string>
+    {
+        public EnvironmentDictionary()
+        {
+            AddVariable("USER");
+            AddVariable("TEMP");
+            AddVariable("TMP");
+            AddVariable("TMPDIR");
+            switch (DesktopInterop.CurrentPlatform)
+            {
+                case PluginPlatform.Linux:
+                    AddVariable("DISPLAY");
+                    AddVariable("WAYLAND_DISPLAY");
+                    AddVariable("PWD");
+                    AddVariable("PATH");
+                    break;
+                case PluginPlatform.Windows:
+                    AddVariable("USERPROFILE");
+                    break;
+            }
+        }
+
+        private void AddVariable(string variable)
+        {
+            var value = Environment.GetEnvironmentVariable(variable);
+            base.Add(variable, value);
+        }
+    }
+}
--- a/OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
+++ b/OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
@@ -9,11 +9,11 @@ namespace OpenTabletDriver.Desktop.Diagnostics
     {
         public EnvironmentDictionary()
         {
-            AddVariables("USER", "TEMP", "TMP", "TMPDIR");
+            AddVariable("USER", "TEMP", "TMP", "TMPDIR");
             switch (DesktopInterop.CurrentPlatform)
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace OpenTabletDriver.Desktop.Diagnostics
+{
+    public class EnvironmentDictionary : Dictionary<string, string>
+    {
+        public EnvironmentDictionary()
+        {
+            AddVariable("USER");
+            AddVariable("TEMP");
+            AddVariable("TMP");
+            AddVariable("TMPDIR");
```
Sublimelime, line 15: Probably better to do this as an array of strings or Collection then for loop it - if in the future we need to add more vars this might get a bit spammy to just list AddVariable calls.

Something like

```
var variables = new List<string>{"USER", "TEMP", ...}
for(v in variables) 
    AddVariable(v);
```

Just my $0.02.

InfinityGhost: Could just pass `params string[] variables` instead and it would allow to just call it as 
```cs
AddVariables("USER", "TEMP", ...);
```

OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
```diff
@@ -0,0 +1,36 @@
...
+                    break;
+                case PluginPlatform.Windows:
+                    AddVariable("USERPROFILE");
+                    break;
+            }
+        }
+
+        private void AddVariable(string variable)
+        {
+            var value = Environment.GetEnvironmentVariable(variable);
```
Sublimelime, line 32: This breaks if the specified variable isn't found - might be worth just forgoing the above switch, putting this in a try catch, and just having one big list of all envs for all platforms - wrong platform? Just throws, gets caught and ignored, and everything works out. No need for the above switch or remembering which platform has which variable.

OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
Sublimelime: Yes, that would also work.

OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
X9VoiD: It just returns `null` if not found.

InfinityGhost: `Environment.GetEnvironmentVariable()` returns null when the variable isn't found. This is desired.

OpenTabletDriver.Desktop/Diagnostics/EnvironmentDictionary.cs
```diff
@@ -9,24 +9,24 @@ public class EnvironmentDictionary : Dictionary<string, string>
...
                 case PluginPlatform.Windows:
                     AddVariable("USERPROFILE");
                     break;
             }
         }
 
+        private void AddVariables(params string[] variables)
+        {
+            foreach (var variable in variables)
+                AddVariable(variable);
```
Sublimelime, line 27: Why do this instead of just doing the code you have inside of AddVariable, in this for loop lol

X9VoiD: To not change both if AddVariable changed.

gonX (approved):

PR #1696, opened 2021-12-05
Add .editorconfig

InfinityGhost:


.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+trim_trailing_whitespace = true
+indent_style = space
+
+[*.{cs,csproj,json,yml}]
+tab_width = 4
```
X9VoiD, line 8: This will require changes to all of our .csproj and .json files, you probably meant 2 spaces for them.

X9VoiD (changes requested):
We might want to add more rules, like how to name variables, private fields etc.

.editorconfig has so much more stuff it can do, but this is a good start.

.editorconfig
InfinityGhost: Oops good catch

PR #1042, opened 2021-04-23
Improve diagnostics serialization behaviour

X9VoiD:
Make a "best-effort" diagnostics output in case of serialization errors. The title is also changed to "prompt" the user better that they are supposed to choose a location for saving the diagnostics.

## Changes

- "Exporting diagnostic information" -> "Save diagnostic information to..."
- Handle serialization exceptions

PR #1700, opened 2021-12-06
Use saner default file dialog locations in UX

gonX:
Currently, we either use nothing, or working directory (typically install location)

This is annoying for Linux users typically because of the lack of write access, and likely unwanted for Windows users as well.

This PR changes it to Documents which is very likely to be what the user wants.


X9VoiD:
Looks like it duplicates a part of #1042. Will rebase that PR so it doesn't touch the folders anymore.

X9VoiD (approved):

InfinityGhost (approved):

PR #1716, opened 2021-12-11
Console command for presets

vedattt:
Continuing from #1715:
Added basic support for presets in `OTD.Console`.
Tested on Windows, seems to work as intended.


gonX:
I'd prefer this to just be included with #1715. The extra code you're doing here looks okay at a quick glance btw.

vedattt:
Will do.

Issue #1719, opened 2021-12-11
NRE on any form of input


X9VoiD:
## Description

Reproduction:
- Do anything with the tablet.

```
[NullReferenceException:Error]	Object reference not set to an instance of an object.
   at OpenTabletDriver.Devices.DeviceReader`1.Main() in (schtuff)/OpenTabletDriver/Devices/DeviceReader.cs:line 106
```

gonX:
Can't reproduce. Commit id and OS?

andigandi:
Is this related to #1703 ? Are you using dotnet 6?

X9VoiD:
> Can't reproduce on [4b9e63f](https://github.com/OpenTabletDriver/OpenTabletDriver/commit/4b9e63f58ba5c1a271daa995acdf51eb2fb623de). What your commit id and OS?

4b9e63f and Arch Linux

> Is this related to #1703 ? Are you using dotnet 6?

Not related, this is from a clean build with zero settings. And OTD.Daemon/UX only builds with .NET 6.

X9VoiD:
It's an issue with the build script not clearing out the ./bin folder.

PR #1720, opened 2021-12-12
build.sh: Clean old artifacts before building

gonX:
`dotnet publish` doesn't seem to handle consecutive builds well if code contents have changed in the mean time

This was an issue in OpenTabletDriver/OpenTabletDriver#1719

**Pre-PR**:
Old `./bin` folder is left as-is, resulting in changed files not being rebuilt, seemingly.

**Post-PR**:
We effectively `rm -r ./bin` before the build process starts.
We also run `dotnet clean` for extra insurance.
It also moves the `./bin` folder to a temporary spot (`./bin_`) in case the user runs the script in the wrong location - this can be changed, since I'm unsure of what best practice is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Build Artifacts
 obj/
 bin/
+bin_/
 nupkg/
 [Bb]uild/
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 # Build Artifacts
 obj/
 bin/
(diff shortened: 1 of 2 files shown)


InfinityGhost (approved):

gonX:
Forgot to update the Post-PR, but for the record, it instead fixes the current working directory to the script location and verifies that the `OpenTabletDriver` directory exists next to the script

PR #1725, opened 2021-12-13
Show any unhandled exception in a message box on all platforms

InfinityGhost:


gonX:
Did you test this code? I'm getting an NRE from this

PR #1734, opened 2021-12-15
Fix device interface matching  for uc-logic/tablet 1060n on macos

AkiSakurai:
There are three interfaces on this tablet.
On macos interfaces are enumerated non-deterministically, so sometimes invalid interface got chosen.


X9VoiD (approved):

X9VoiD:
Retarget the PR to `configs` branch.

PR #1736, opened 2021-12-15
Centralize pointer input injection

X9VoiD:
### Removed `IInteruptBinding`

- The interface allows for non-isolated changes and undesired effects. The only current use-case for this interface is `PenPassthroughBinding` to extend `AbsoluteOutputMode` or `RelativeOutputMode`'s `OnOutput`, but this introduces the so-called "undesired effects". The binding may be assigned to any bindable, and it will inject Pressure, and manage Pen Button 1, and 2, even if there's no binding set in those places.

### Centralized pointer input injection

- As a cleaner alternative, all pointer input injection is moved to the output modes as where they should have really been.

### Misc changes

- Renamed `IRelativePointer.Translate(Vector2)` to `IRelativePointer.SetPosition(Vector2)` for uniformity, the interface and parameter name are enough hints to say that this is *relative*.
- Added `LinuxArtistModeButtonBinding` to manually control its bindings, which was not possible before due to the side-effects of `PenPassthroughBinding`.
- Separated `IVirtualTablet` into cleaner interfaces, as not all pointers support all of those features.
- Added `ISynchronousPointer` to centralize "synchronous" functionality like `Reset()` and the now-removed `IVirtualTablet.Sync()`.
- All pointer-related input injection are now moved to the output modes, rather than scattered somewhere else.

## Notes

- There is no pen tip binding for `LinuxArtistMode` as this requires moving the binding handler to before the last element of the output mode for it to work properly (needs to block pressure).


gonX:
Looks cool at a brief glance, will test and review later.

OpenTabletDriver.Plugin/Platform/Pointer/ISynchronousPointer.cs
```diff
@@ -0,0 +1,8 @@
+namespace OpenTabletDriver.Plugin.Platform.Pointer
+{
+    public interface ISynchronousPointer
+    {
+        public void Reset();
+        public void Flush();
```
InfinityGhost, line 6:
```suggestion
        void Reset();
        void Flush();
```

InfinityGhost (changes requested):

InfinityGhost (approved):

gonX:
needs rebase btw

gonX:
I'm getting a daemon crash when my pen goes in range. It doesn't do that if I rebase to master, so it's probably nothing.

However, it seems like Tip Threshold isn't being obeyed. This should be fixed before merging.

gonX:
Actually, because of the way libinput works we can theoretically merge this as-is. While users will not be able to remap pressure with the pen settings UI with this change, a plugin would likely be able to do it, meaning that we don't need to worry about this being unfit for 0.6.x.

I would however prefer if `BTN_TOUCH` would only be sent on tip down. Can this add another binding (with special treatment) for pen tip that actuates the `BTN_TOUCH` event? I'll try to see if I can make sense of your code and add that behavior myself. Basically `BTN_TOUCH` should not be emitted by artist mode, but rather by the binding.

gonX (changes requested):
Add my PR and I'll accept it. Emitting `BTN_TOUCH` when the tip isn't actually touching seems to be incorrect behavior.

X9VoiD:
See Notes, some environments, like Gnome, still accept any form of pressure as a touch. Adding a pen tip binding will make it seem like we have control over it, but we don't for now.

gonX:
Ok, that kind of leaves it not mergeable in time for 0.6.x if you ask me. That or there needs to be a pressure rewrite earlier in the stack that foregoes this issue.

We should be using pen pressure threshold to rewrite pressure as early as we can anyway - preferably at parser level, or first step after it.

X9VoiD:
Cons of not merging:
- As long as pen passthrough binding is binded to anything, pen tip, button 1 and button 2 will all be binded.
- `IVirtualTablet` is used only in pen passthrough binding and is not an actual OTD plugin interface, the only object that could make use of it is [`EvdevVirtualTablet`](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/edcee3d2db1974779f33565b476ea6044caf9c02/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs#L97) as `IVirtualTablet` is only declared [here](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/edcee3d2db1974779f33565b476ea6044caf9c02/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs#L264) and only there.
- No Windows Ink support. Pen passthrough binding relies on DI as described above, however doing the same on Windows Ink will yield the "system virtual tablet", which is null on Windows.

Cons of merging:
- Pen tip is automatically binded.

gonX:
It's not that I mind the pen tip automatically being bound, only the side effect of pressure not being adjustable. It's very likely it can be fixed in a plugin if anyone wants to.

Without X9VoiD/OpenTabletDriver#6 its however constantly emitting `BTN_TOUCH` which I interpret as going against [Linux kernel event code spec](https://www.kernel.org/doc/Documentation/input/event-codes.txt):
> `BTN_TOUCH` should be used to report when the tool is in contact with the tablet.

At the very least we should be not be blindly emitting `BTN_TOUCH` when pressure is not 0, it'd very likely break application unknowingly.

gonX:
> Current [master](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/edcee3d2db1974779f33565b476ea6044caf9c02/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs#L106) always sends it.

It does not, [because pressure is rewritten in the binding handler](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/edcee3d2db1974779f33565b476ea6044caf9c02/OpenTabletDriver.Desktop/Binding/ThresholdBindingState.cs#L13).

PR #1748, opened 2021-12-16
Add help command to OTD.Console

### LINKED ISSUE #1674: OTD Console doesn't know the "help" command
## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Arch Linux
| OpenTabletDriver Version |  181b77b
| Tablet                   | XP-Pen Star G640

Doesn't know the help command yet shows help anyway.

image

vedattt:
## Changes

- Add a `help` command to the console, shows the exact same help that `System.CommandLine` normally does.

## Issues

- Closes #1674 

---

Honestly this is more of a mistake on [dotnet/command-line-api](https://github.com/dotnet/command-line-api)'s part and I would guess that they'll do something about it as their API matures, but for the time being this does close an issue here, so.

### DIFF
--- a/OpenTabletDriver.Console/Program.Commands.cs
+++ b/OpenTabletDriver.Console/Program.Commands.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
+using System.CommandLine.Help;
+using System.CommandLine.IO;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


gonX:
I feel like we should probably be closing the original issue with `wontfix` since as you've described this feels more like an upstream issue that is so impressively minor for us, especially because this adds 19 lines of code just to trim 3 lines of output in an edge case situation.

If someone can find an upstream issue tracking this that'd be cool.

vedattt:
Closing as this is actually not specific to the `help` command.

The same issue occurs when any valid option is attempted without the prefix `--`, it would appear that the dotnet CommandLine API is showing the correct help but messing it up by omitting the option prefix of the suggested alternative.

PR #1750, opened 2021-12-16
Switch to manually adding/removing tabs instead of TabPage.Visible

### LINKED ISSUE #1723: macOS shows tabs which are intended to be hidden
## Description

macOS does not hide tabs as intended, a workaround is likely required.

image

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | macOS Big Sur
| OpenTabletDriver Version | 4b9e63f58ba5c1a271daa995acdf51eb2fb623de

vedattt:
## Changes

- Stopped relying on `TabPage.Visible` entirely in `ControlForm.cs`.
- Upon each change, tabs are added onto `tabControl.Pages` from scratch depending on the state.

## Issues

- Closes #1723

### DIFF
--- a/OpenTabletDriver.UX/Controls/ControlPanel.cs
+++ b/OpenTabletDriver.UX/Controls/ControlPanel.cs
@@ -17,60 +17,53 @@ namespace OpenTabletDriver.UX.Controls
     {
         public ControlPanel()
         {
-            tabPages = new(new TabPage[]
+            new TabPage
+            {
+                Text = "Output",
+                Content = outputModeEditor = new()
+            };
+            new TabPage
+            {
+                Text = "Filters",
+                Padding = 5,
+                Content = filterEditor = new()
+            };
+            new TabPage
+            {
+                Text = "Pen Settings",
+                Content = penBindingEditor = new PenBindingEditor()
+            };
+            new TabPage
+            {
+                Text = "Auxiliary Settings",
+                Content = auxBindingEditor = new AuxiliaryBindingEditor()
+            };
+            new TabPage
+            {
+                Text = "Mouse Settings",
+                Content = mouseBindingEditor = new MouseBindingEditor()
+            };
+            new TabPage
+            {
+                Text = "Tools",
+                Padding = 5,
+                Content = toolEditor = new()
+            };
+            new TabPage
+            {
+                Text = "Info",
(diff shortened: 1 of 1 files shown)


X9VoiD (changes requested):
Seems fine.

But it's probably safer and cleaner to introduce a new class that holds all of the tabs so that the "keys" are enforced via type information instead of strings.

```cs
internal class TabPageStore : IEnumerable<IForgotTheInterfaceOfTabPages>
{
    // Put the nullable tab pages here.
    // Then use yield return to enumerate the non-null contents of the store in order.
    // If Clear() is really needed then...
    public void Clear()
    {
        // Set pages to null to disable
    }
}
```

vedattt:
Hmm, I'm having some trouble understanding the change here, let me try to tell you what I think I understood. 

`TabPageStore` would basically be `tabPages` but moved into its own class, and it would be a direct substitute for `TabControl.Pages`. Instead of controlling what gets added into `TabControl.Pages`, we would control tabs by setting them to null in `TabPageStore`, which would effectively hide them away from any iterators. 

So for example, we point `tabControl.Pages` to an instance of `TabPageStore`, and if the profile change event is triggered, we use some method in `TabPageStore` to `null`ify its tab elements that are not needed.

Assuming my understanding is correct, I have a question though:
I thought since `TabControl.Pages` is readonly, there wouldn't be a way to directly set `TabPageStore` as its actual page store.
As a workaround, maybe iterate its contents into a cleared `tabControl.Pages` upon every change, would that be a good idea?

vedattt:
Okay, after trying my hand at a `TabPageStore` implementation, I realized I was just doing the dictionary stuff from within another class and wasn't really getting anywhere. I took a step back and realized this was way overkill from the very beginning, so I rebased the entire PR from scratch.

X9VoiD (approved):

Sublimelime:
I like this, pretty clean. Haven't seen `new()` before like that, guessing it just creates a new instance of whatever's on the left like `var` does for what's on the right?

vedattt:
> Haven't seen `new()` before like that, guessing it just creates a new instance of whatever's on the left like `var` does for what's on the right?

Yep;
image

InfinityGhost (approved):

PR #1751, opened 2021-12-16
Fix `detect tablet` not disposing outdated pipelines

### LINKED ISSUE #1745: `Detect tablet` doesn't destroy previous pipelines
## Description
All the pipelines end up working simultaneously:

https://user-images.githubusercontent.com/38468635/146342182-65c7c94f-a9cc-4722-9da7-adf0bcec293d.mp4

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Bimbows 10
| OpenTabletDriver Version | ecc3218c49b5ca5796c434b994f730cd81d238e7

vedattt:
Seems to work on my end, #1745 is no longer reproducible after following the steps in the video.
Haven't noticed any other side effects.

## Changes

- Add a call to `Dispose` during tablet detection in `Driver`.
- Modify `Dispose` to use `List`s since .NET doesn't allow modifications during enumeration.

## Issues
- Closes #1745

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -201,8 +201,8 @@ namespace OpenTabletDriver
 
         public void Dispose()
         {
-            foreach (InputDeviceTree tree in InputDevices)
-                foreach (InputDevice dev in tree.InputDevices)
+            foreach (InputDeviceTree tree in InputDevices.ToList())
+                foreach (InputDevice dev in tree.InputDevices.ToList())
                     dev.Dispose();
         }
     }
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -42,6 +42,7 @@ namespace OpenTabletDriver
 
             Log.Write("Detect", "Searching for tablets...");
 
+            Dispose();
             InputDevices.Clear();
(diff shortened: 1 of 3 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -42,6 +42,7 @@ public virtual bool Detect()
 
             Log.Write("Detect", "Searching for tablets...");
 
+            Dispose();
```
InfinityGhost, line 45: While this is a fix, I don't think this is the right way to go about things. We should probably create a class for `InputDevices` based on a list that disposes all of them on `Clear()` rather than disposing the Driver class itself.

InfinityGhost (changes requested):

vedattt:
> We should probably create a class for InputDevices based on a list that disposes all of them on Clear() rather than disposing the Driver class itself.

Was this (3dd3f0e51b11b415151de7378267f96a2a1baec9) what you had in mind?

vedattt:
Oops, made a mistake with the list implementation, will adjust according to this: [Collection.Clear - Notes to inheritors](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.clear?view=net-6.0#notes-to-inheritors)

InfinityGhost (approved):

PR #1753, opened 2021-12-16
Refactor HandleTabletsChanged in TabletSwitcherPanel to be less redundant

Sublimelime:
Small PR to refactor the `HandleTabletsChanged` function in TabletSwitcherPanel to be a bit cleaner and less redundant, relying on existing functionality instead of checking for existing profiles - instead, we just call `GetProfile` for every tablet, which already generates profiles for tablets without them.

This should be more efficient, resulting in fewer queries of the profile collection (avoids the linq as well) and fewer foreach loops.

I've also renamed the aforementioned `GetProfile` function to `GetProfileOrGenerate` to more accurately reflect its behavior. This probably could/should be in its own PR but it's just a rename.

Tested on my setup with my one tablet, seems to work fine, would appreciate further testing to ensure I haven't caused any regressions or edge cases here. As far as I can tell, it's 1:1 behavior.

### DIFF
--- a/OpenTabletDriver.Desktop/Profiles/ProfileCollection.cs
+++ b/OpenTabletDriver.Desktop/Profiles/ProfileCollection.cs
@@ -26,7 +26,7 @@ namespace OpenTabletDriver.Desktop.Profiles
         public Profile this[TabletReference tablet]
         {
             set => SetProfile(tablet, value);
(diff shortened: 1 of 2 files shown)


Sublimelime:
Superseded by #2352

PR #1758, opened 2021-12-18
Fix NPE on macos

AkiSakurai:
On macos TrayIndicator can not be shown with null menu


vedattt:
Pretty sure that line is not supposed to be there in the first place... Git blame points to the self-updater merge, weird.

gonX (approved):

InfinityGhost (approved):

PR #1781, opened 2021-12-21
Added Gaomon 830 (RNG Edition) Configurations

Mrcubix:
# Verification:

https://discord.com/channels/615607687467761684/789348845372178482/921856897294942308

# Changes:

- Add Gaomon 830 Configuration

# Note

- Require Luck on the user's part for it to be detected as the strings index are randomized after every requests, although it seem like this index is the most consistent
- RNG Tablet


X9VoiD:
Do these "randomly generated strings" not have any pattern to them? Like "S830". Can't merge a config that works "randomly".

PR #1783, opened 2021-12-22
Fix memory leaks on MacOS

AkiSakurai:
Memory leaks can be avoided by using shared FormattedText to draw text.


OpenTabletDriver.UX/Controls/Output/Area/AreaDisplay.cs
```diff
@@ -12,11 +12,29 @@ namespace OpenTabletDriver.UX.Controls.Output.Area
...
+                graphics.DrawText(sharedFormattedText, location);
+            }
+        }
+
         private AreaSettings area;
         private bool lockToUsableArea;
         private string unit, invalidForegroundError, invalidBackgroundError;
         protected IEnumerable<RectangleF> areaBounds;
         private RectangleF fullAreaBounds;
+        private readonly TextDrawer textDrawer = new ();
```
InfinityGhost, line 37:
```suggestion
        private readonly TextDrawer textDrawer = new();
```

InfinityGhost (changes requested):

InfinityGhost (approved):

PR #1788, opened 2021-12-24
Improve config for UG-LOGIC/1060n

AkiSakurai:
I find from old discord message that there are indeed two variants of xp-pen star 03 v1 with same pair of pid/vid and same set of device strings.

This first variant don't have output report, and can only be used on windows with winusb.
[discord](https://discord.com/channels/615607687467761684/615611007951306863/825352999961362433)
I own this one.

The second variant have extra output report for activating the device. No need to use winusb on widows.
[discord](https://discord.com/channels/615607687467761684/615611007951306863/834760454516506674)
[discord](https://discord.com/channels/615607687467761684/615611007951306863/834758580099874816
)

Besides, I find no confirmation for pressure, so the pressure of original config is probably wrong. 
The pressure should be 2047.

Depends on #1793

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/UC-Logic/1060N.json
+++ b/OpenTabletDriver.Configurations/Configurations/UC-Logic/1060N.json
@@ -8,7 +8,7 @@
       "MaxY": 30480.0
     },
     "Pen": {
-      "MaxPressure": 8191,
+      "MaxPressure": 2047,
       "Buttons": {
         "ButtonCount": 2
       }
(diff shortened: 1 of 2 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -151,6 +151,7 @@ where device.CanOpen
                 where identifier.InputReportLength == null || identifier.InputReportLength == device.InputReportLength
                 where identifier.OutputReportLength == null || identifier.OutputReportLength == device.OutputReportLength
                 where DeviceMatchesStrings(device, identifier.DeviceStrings)
+                where DeviceMatchesAttribute(device, identifier.Attributes)
```
Sublimelime, line 154: This should probably be moved into it's own PR, as this could potentially break detection of other tablets.

If this change is necessary for this tablet, create the secondary PR, and mark this PR as dependent on it.

AkiSakurai:
There are probably more variants for ths model other then listed two.

PR #1789, opened 2021-12-24
Tablet Debugger: Enhancements

gonX:
Pre-PR:
- Tablet debugger lacks row count for some tablets, especially on Windows where they can't resize the window.
- Tablet visualizer has asinine screen size requirements for larger tablets; accounting for DPI makes visualization on a display tablet impossible (also makes it very annoying to use on tiling WM's without allotting a full monitor to it)

Post-PR:
- More rows for raw data (Fixes OpenTabletDriver#1785)
- Visualizer supports resizing and has a minimum height of 200 instead.
- Report rate and related options have been moved next to tablet name to relax the window height requirements a bit. Font size for fields on same row have been increased for cleanliness because the checkbox is increasing the height of the row.
- Moved some magic numbers into consts

### DIFF
--- a/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
+++ b/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
@@ -130,7 +130,8 @@ namespace OpenTabletDriver.UX.Windows.Tablet
             var splitter = new Splitter
             {
                 Orientation = Orientation.Vertical,
-                Width = 450,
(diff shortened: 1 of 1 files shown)


Kuuuube (approved):
I can't say anything on the actual code but this works much better than before. Looks good to me.

InfinityGhost:
This looks really good on GTK
image

InfinityGhost (approved):
lg2m

PR #1798, opened 2021-12-25
Enforce json extension when saving presets

### LINKED ISSUE #1795: Preset files not saved with json extension are not detected during refresh
## Description

As in title, if you save a preset without including the .json extension at the end, the files are not detected by OTD when presets are refreshed.

## Reproduction:

- Save preset, do not provide extension
- Hit refresh presets, none will appear
- Rename file to "whatever.json" and hit refresh again, will detect

## Fix

I can fix this, but one of two ways:

- Detect all files in preset folder regardless of extension (may cause errors though if there's unrelated clutter there)
- Ensure files are saved with .json extension (appended if not provided by user)

Would appreciate advice, **leaning towards the second option** (as windows doesn't like extension-less files anyway, and they *are* json files so should have json extension).

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Linux (but should apply to all platforms, can't remember if windows forces extensions)
| OpenTabletDriver Version | Master (v0.6.0.0+f06a0ed6)
| Tablet                   | Irrelevant

vedattt:
## Changes

Self-explanatory single line PR.

## Issues

- Closes #1795

### DIFF
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -513,7 +513,7 @@ namespace OpenTabletDriver.UX
             {
                 case DialogResult.Ok:
                 case DialogResult.Yes:
-                    var file = new FileInfo(fileDialog.FileName);
+                    var file = new FileInfo(fileDialog.FileName + (fileDialog.FileName.EndsWith(".json") ? "" : ".json"));
                     if (App.Current.Settings is Settings settings)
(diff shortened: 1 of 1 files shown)


X9VoiD (approved):

Sublimelime (approved):
Ah darn ya beat me to it. Pretty much exactly what I was gonna do.

InfinityGhost (approved):

PR #1802, opened 2021-12-26
DesktopInterop: Use fallback timer on Linux

### LINKED ISSUE #1727: Memory leak when using interpolators (Linux only?)
## Description
InfinityGhost's guess is that its happening in P/Invoke to Linux threads (my words).

Reproduction steps:
1) Enable any filter which uses an interpolator
2) Memory usage starts climbing immediately until the interpolator is disabled

For convenience [HERE](https://github.com/OpenTabletDriver/OpenTabletDriver/files/7701778/BezierInterpolator.dll.gz) is a precompiled [BezierInterpolator](https://github.com/AbstractQbit/AbstractOTDPlugins/tree/master/BezierInterpolator) build that should work on git tag versioned builds (if your `master` OTD reports 0.5.3.1, ie. Arch), otherwise a `dotnet build -c Release` in its respective location should be enough to get a working .dll.

Can also be observed with `dotnet-trace collect --profile gc-verbose` - `dotnet-trace` is available from `dotnet tool`

Advice how to debug this would be appreciated

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Arch Linux
| OpenTabletDriver Version | 09c841f
| Tablet                   | GD-0608

gonX:
This closes #1727 and seems to have good enough performance that we could in theory use this in time for the 0.6.x milestone

It does at least run BezierInterp at 1000Hz which I think is good enough.

Reminder that this opens a new issue, one that restores `LinuxTimer`.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
+++ b/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
@@ -72,7 +72,6 @@ namespace OpenTabletDriver.Desktop.Interop
         public static ITimer Timer => CurrentPlatform switch
         {
             PluginPlatform.Windows => new WindowsTimer(),
-            PluginPlatform.Linux   => new LinuxTimer(),
             _                      => new FallbackTimer()
         };
 


OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
```diff
@@ -72,7 +72,6 @@ public static void OpenFolder(string path)
         public static ITimer Timer => CurrentPlatform switch
         {
             PluginPlatform.Windows => new WindowsTimer(),
-            PluginPlatform.Linux   => new LinuxTimer(),
```
AbstractQbit, line 75: This should be instead commented out with a TODO of fixing the underlying leak, else everyone might just forget about it

gonX: I see where you're coming from but I'll make an issue if this gets merged.

InfinityGhost (approved):
For a quick and dirty workaround, this will work. Maybe in the future would could provide some sort of setting for this.

PR #1808, opened 2021-12-26
Add GitHub super-linter to and clean up .NET workflow

vedattt:
## Changes

- Add workflow for GitHub's very own [super-linter](https://github.com/github/super-linter), default configuration.
- Separate all steps with whitespace
- Reorder step notation to have name first, add missing names for checkouts
- Bump `actions/checkout@v1` to `v2`
- Rename workflow to ".NET CI"

See example run here: https://github.com/vedattt/OpenTabletDriver/runs/4638132743
(This was before the editorconfig commit though)

Note: The linting seems to take around 140 seconds, most of which goes to pulling the super-linter image.
If that's not good enough and super-linter isn't to be merged, I'll just have to rebase it out of this PR since the cleanup is still good to go.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -1,4 +1,4 @@
-name: .NET Core
+name: .NET CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


vedattt:
Closing since this PR has become a mess and just using `dotnet format` seems to be a superior alternative to super-linter.
Will make a new PR for this purpose.

PR #1819, opened 2021-12-28
Added 10moon 1060 Configurations

Mrcubix:
# Verification:

anything before https://discord.com/channels/615607687467761684/922220574531534848/922948452139298866

# Changes

- Add TenMoonReportParser
- Add 10moon 1060 Configurations

# Note

- Doesn't meet any of the advertised specs (Only 6.52 lines per mm instead of the advertised 200 for example)
- Require multiple Feature Init
- An offset is required for pressure on aux press
- Cannot press multiple keys at the same time due to the fact bits get overwritten once a new aux key is pressed
(leading me to using bytes instead of checking for bits)

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
+++ b/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
@@ -1,0 +1,54 @@
+{
+  "Name": "10moon 1060N",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 254.0,
+      "Height": 152.4,
+      "MaxX": 4095.0,
+      "MaxY": 4095.0
+    },
+    "Pen": {
+      "MaxPressure": 8191,
+      "Buttons": {
+        "ButtonCount": 2
+      }
+    },
+    "AuxiliaryButtons": null,
+    "MouseButtons": null,
+    "Touch": null
+  },
+  "DigitizerIdentifiers": [
+    {
+      "VendorID": 2290,
+      "ProductID": 26641,
+      "InputReportLength": 64,
+      "OutputReportLength": null,
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.TenMoon.TenMoonReportParser",
+      "FeatureInitReport": null,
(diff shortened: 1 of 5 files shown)


TABLETS.md
```diff
@@ -27,6 +27,7 @@
 | Huion WH1409                  |     Supported     |
 | Huion WH1409 V2               |     Supported     |
 | Parblo A640 V2                |     Supported     |
+| 10moon 1060                   |     Supported     | Doesn't meet any of it's advertised specs
```
gonX, line 30: Grammar:
```suggestion
| 10moon 1060                   |     Supported     | Doesn't meet any of its advertised specs
```

Though, I'd prefer the comment to be reworded instead

gonX (commented):
Is it our init that makes it go 4096x4096? Or does the official driver on Win Ink result in similar accuracy?

OpenTabletDriver.Configurations/Parsers/10moon/10moonAuxParser.cs
```diff
@@ -0,0 +1,35 @@
...
+using OpenTabletDriver.Plugin;
+
+namespace OpenTabletDriver.Configurations.Parsers.TenMoon
+{
+    public struct TenMoonAuxReport : IAuxReport
+    {
+        public TenMoonAuxReport(byte[] report)
+        {
+            Log.Debug("10moon Parser", $"{BitConverter.ToString(report)}");
+
```
X9VoiD:
```suggestion
```

X9VoiD (changes requested):

OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
```diff
@@ -0,0 +1,54 @@
+{
+  "Name": "10moon 1060N",
```
X9VoiD, line 2: 1060N or 1060? Update `TABLETS.md` if this is 1060N.

X9VoiD (changes requested):

PR #1823, opened 2021-12-29
Correct the Deco Mini4 name

jamesbt365:
The Deco Mini4 name was incorrect, having a space where it shouldn't be and i have corrected this.


X9VoiD (changes requested):
The official name is `XP-Pen Deco mini4`, with small `m`.

PR #1826, opened 2021-12-29
Update PTH-851.json

SnowLire:
The ProductID seems to be in the wrong base. The correct ID should be 791 decimal, which is 0317 hex.
https://usb-ids.gowdy.us/read/UD/056a/0317

Discussion: https://discord.com/channels/615607687467761684/723399565780189234/925744155005706320


X9VoiD (changes requested):
Remove `Wacom PTH-851` from `Untested`.

PR #1841, opened 2021-12-31
Invert Y tilt axis when using UCLogicTiltReportParser

jkcdarunday:
## Introduction
While testing UCLogic tilt support using RTP-700 with Krita and MyPaint on Linux, I've observed that the Y tilt axis is inverted. 

## Description
These changes inverts the Y tilt axis when a TiltTabletReport is generated by the UCLogicTiltReportParser.

## Notes
- This assumes that the tilt axes for other devices have no issues when on Linux (making this specific to the UCLogic parser and unlikely somewhere else like libevent)
- Haven't tested on Windows/Mac (If it works properly on other platforms, then it's possible that the problem lies with the libevent integration instead)

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicTiltReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicTiltReportParser.cs
@@ -9,7 +9,7 @@ namespace OpenTabletDriver.Configurations.Parsers.UCLogic
             if (data[1].IsBitSet(6))
                 return new UCLogicAuxReport(data);
             else
-                return new TiltTabletReport(data);
+                return new TiltTabletReport(data, false, true);
         }
     }
 }
--- a/OpenTabletDriver.Plugin/Tablet/TiltTabletReport.cs
+++ b/OpenTabletDriver.Plugin/Tablet/TiltTabletReport.cs
@@ -5,7 +5,7 @@ namespace OpenTabletDriver.Plugin.Tablet
 {
     public struct TiltTabletReport : ITabletReport, ITiltReport
     {
-        public TiltTabletReport(byte[] report)
+        public TiltTabletReport(byte[] report, bool invertTiltX, bool invertTiltY)
         {
             Raw = report;
 
@@ -16,8 +16,8 @@ namespace OpenTabletDriver.Plugin.Tablet
             };
             Tilt = new Vector2
             {
-                X = (sbyte)report[10],
-                Y = (sbyte)report[11]
+                X = (sbyte)(invertTiltX ? -1 : 1) * (sbyte)report[10],
+                Y = (sbyte)(invertTiltY ? -1 : 1) * (sbyte)report[11]
             };
             Pressure = Unsafe.ReadUnaligned<ushort>(ref report[6]);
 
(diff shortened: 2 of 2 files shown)


terencode:
Thanks, that's also what I noticed with my Kamvas pro 20. I will give it a try soon.

terencode:
I modified the parser responsible for my tablet directly because it seemed to redefine itselfs the X, Y tilts :
https://github.com/terencode/OpenTabletDriver/blob/d44bdeb2a74482c970931f167532c73e5973483f/OpenTabletDriver.Configurations/Parsers/Huion/GianoReport.cs#L21
Tilt is correct now.
It would be interesting to test on other devices if this is parser specific or needs to be changed globally.

AbstractQbit:
What are the values shown in the debugger? They should be negative for tilt to the left or up, positive for tilt to the right or down (for X and Y respectively). Two huion tablets that I had my hands on (hs610 and hs611) as well as wacom tablets report tilt correctly, so this issue is either device-specific or something is wrong with how we inject tilt on linux. In any case flipping Y axis shouldn't be the default

terencode:
> What are the values shown in the debugger? They should be negative for tilt to the left or up, positive for tilt to the right or down (for X and Y respectively). Two huion tablets that I had my hands on (hs610 and hs611) as well as wacom tablets report tilt correctly, so this issue is either device-specific or something is wrong with how we inject tilt on linux.

Without the Y flipped, I get positive values when tilting up, so device-specific ?

> In any case flipping Y axis shouldn't be the default

Yeah, my modification was just a dirty fix for testing.

gonX:
This seems sort of sensible but I wonder if we should be fixing this another place in OTD, or with an attribute in the device configuration file.

Also HS610 and HS611 does not use a tilt-compatibler parser (at the moment), though they should very likely be using the `UCLogicTiltReportParser` as used here.

I'd like for this to have more testing, especially on the HS610/HS611 as this'll determine which course of action is best. @AbstractQbit can you confirm HS610/HS611 tilt axises are correct on `configs` (not this PR) if you use `UCLogicTiltReportParser`?

AbstractQbit:
Back at home now, seems like I was wrong. Just checked hs611 and it does have Y axis inverted, so it's possible that a lot of UClogic tablets report tilt that way. ~Can't check hs610 because I can't find it.~

AkiSakurai:
https://github.com/DIGImend/digimend-kernel-drivers/blob/0b445150ef491cb52174aef484f6873083770171/hid-uclogic-params.c#L472
According to DIGImend, it is flipped on device using v2 protocol.

gonX:
Sorry for not getting back to this earlier. This needs to be rebased to the latest version of `configs` before we can proceed with merging.

sezanzeb:
Rebasing worked locally without conflicts.

For the huion kamvas 16 (2021) this needs to be adjusted in the `GianoReport` class.

I think this doesn't take into account that the input can be rotated. When I hardcoded the inversion into the `GianoReport` class, I had to invert the X axis, not the Y axis.

daishishi:
Glad that this issue is known to the OTD team, just a bummer that still isn't merged. Although I believe the Huion V2 models are a rare breed currently. (I have a H610 V2 Pro)

PR #1846, opened 2022-01-01
Fix middle button binding on macos

AkiSakurai:
The enum values are incorrect.

### DIFF
--- a/OpenTabletDriver.Native/OSX/Generic/CGEventType.cs
+++ b/OpenTabletDriver.Native/OSX/Generic/CGEventType.cs
@@ -12,8 +12,8 @@
         kCGEventRightMouseDragged = 7,
         kCGEventKeyDown = 10,
         kCGEventKeyUp = 11,
-        kCGEventOtherMouseDown = 16,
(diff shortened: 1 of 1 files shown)


gonX (approved):
Ctrl+F `OtherMouseUp` [here](http://jnaerator.sourceforge.net/sites/rococoa-auto/apidocs/constant-values.html#org.rococoa.cocoa.coregraphics.CoreGraphicsLibrary.CGEventType.kCGEventTabletPointer) seems to match those values. Looks like a sensible change to me.

Has this been tested?

InfinityGhost (approved):

InfinityGhost:
> Ctrl+F `OtherMouseUp` [here](http://jnaerator.sourceforge.net/sites/rococoa-auto/apidocs/constant-values.html#org.rococoa.cocoa.coregraphics.CoreGraphicsLibrary.CGEventType.kCGEventTabletPointer) seems to match those values. Looks like a sensible change to me.
> 
> Has this been tested?

I've already commented that the values may be incorrect and this makes more sense. Untested, however it wouldn't break it any worse if it isn't the fix.

PR #1847, opened 2022-01-01
Adjust some wording within the UX.

jamesbt365:
Adjusted some wording within the UX in order to make it easier to understand, and explaining some less explained features.

Also, during this i adjusted "OTD" in the guide and manpages to "OpenTabletDriver"

### DIFF
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -178,7 +178,7 @@ namespace OpenTabletDriver.UX
             var pluginManager = new Command { MenuText = "Open Plugin Manager..." };
             pluginManager.Executed += (sender, e) => App.Current.PluginManagerWindow.Show();
 
-            var wikiUrl = new Command { MenuText = "Open wiki..." };
+            var wikiUrl = new Command { MenuText = "Open Wiki..." };
             wikiUrl.Executed += (sender, e) => DesktopInterop.Open(WikiUrl);
 
             var showGuide = new Command { MenuText = "Show guide..." };
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -178,8 +178,8 @@ namespace OpenTabletDriver.UX
             var pluginManager = new Command { MenuText = "Open Plugin Manager..." };
             pluginManager.Executed += (sender, e) => App.Current.PluginManagerWindow.Show();
 
-            var faqUrl = new Command { MenuText = "Open FAQ Page..." };
-            faqUrl.Executed += (sender, e) => DesktopInterop.Open(FaqUrl);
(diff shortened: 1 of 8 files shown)


OpenTabletDriver.UX/MainForm.cs
```diff
@@ -178,8 +178,8 @@ private MenuBar ConstructMenu()
             var pluginManager = new Command { MenuText = "Open Plugin Manager..." };
             pluginManager.Executed += (sender, e) => App.Current.PluginManagerWindow.Show();
 
-            var faqUrl = new Command { MenuText = "Open FAQ Page..." };
-            faqUrl.Executed += (sender, e) => DesktopInterop.Open(FaqUrl);
+            var wikiUrl = new Command { MenuText = "Open wiki..." };
```
gonX, line 181:
```suggestion
            var wikiUrl = new Command { MenuText = "Open Wiki..." };
```

OpenTabletDriver.UX/Windows/Greeter/Pages/PluginPage.cs
```diff
@@ -17,6 +17,7 @@ public PluginPage()
                 "Plugins can be downloaded from the plugin manager at your own risk.",
                 "The plugin manager can be found in the Plugins menu in the main window.",
                 "Filters modify the output of the tablet, an example of this is smoothing.",
+                "Using multiple filters at once can cause unexpected effects.",
```
gonX, line 20: Like what? Just for documentation purposes, I don't necessarily think this needs further clarification, but it would be nice.

gonX (approved):
Looks OK, I ran `grep -Ri otd` and couldn't find anything else that makes sense to change for now.

OpenTabletDriver.UX/Windows/Greeter/Pages/PluginPage.cs
jamesbt365: What do you think should be put instead or should i just cut it out?

gonX: I just want it clarified here in the PR, not in a separate commit necessarily. I just want to know what sort of side effects can be expected, and in which combinations of filters. I don't using multiple filters should be causing unexpected effects unless you mean something like using multiple anti-smoothing filters

jamesbt365: Use of multiple interpolators can cause issues, but with the interpolators and filter tab combined into one i thought just to put multiple filters, should i change this?

InfinityGhost: It seems reasonable to me.

jamesbt365:
Should be good to go if nothing else needs changing.

InfinityGhost (approved):

PR #1857, opened 2022-01-03
DriverDaemon.cs: Check for stdout before blindly assuming it exists

gonX:
On some systems stdout might be missing.

This handles that by checking if the `BaseStream` exists before doing a `WriteLine`.

Fixes OpenTabletDriver#1855

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -37,7 +37,8 @@ namespace OpenTabletDriver.Daemon
             Log.Output += (sender, message) =>
             {
                 LogMessages.Add(message);
-                Console.WriteLine(Log.GetStringFormat(message));
+                if (((StreamWriter)Console.Out).BaseStream != null)
+                    Console.WriteLine(Log.GetStringFormat(message));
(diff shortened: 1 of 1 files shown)


gonX:
Probably worth noting that this doesn't cover if stdout closes after startup. I wonder if replacing `Console.WriteLine` with `Console.Out?.WriteLine` would fix that.

InfinityGhost:
> I wonder if replacing `Console.WriteLine` with `Console.Out?.WriteLine` would fix that.

I would suggest this in place of this setup, since it's able to check at any point if it can write to stdout.

InfinityGhost (changes requested):
Apply discussed changes.

vedattt:
[Relevant Discord convo](https://discord.com/channels/615607687467761684/628651971267788800/928020159271170159)

`Console.Out` is never null due to its dotnet implementation, so direct nullchecks won't achieve anything.
Instead, the discussed alternative is to nullcheck the `BaseStream` of the `StreamWriter` instance, which also seems to be the preferred way from what I've seen looking around online.

gonX:
Updated to use new approach

gonX:
Closing as stale, there are probably better ways to solve the linked issue.

PR #1859, opened 2022-01-03
Fix PTZ-x30/31W hover distance

Kuuuube:
Currently PTZ-x30/31W tablets have a few mm of their hover cut off ~~and around half of the reports sent when using a ZP-600 pen get ignored (for reasons detailed below).~~

This fixes hover distance and is sent when nearly at max hover distance but still contains proper position data.
```
0xA0 => new IntuosV1TabletReport(data),
```

~~These two fix the ZP-600. For some reason it sends a combination of 0xE0 and 0xEA reports when hovering at low distance and 0xA0 and 0xAA at near max distance.~~
```
0xEA => new IntuosV1TabletReport(data),
0xAA => new IntuosV1TabletReport(data), 
```

~~It is also worth noting that the ZP-600 sends rotation reports so that should be supported eventually. But that is outside the scope of this PR.~~


Kuuuube:
On further inspection the ZP-600 changes probably shouldn't be made like this since they send rotation data over the pressure bytes. Handling this properly requires more parser changes so I'll remove the ZP-600 changes from this PR.

PR #1885, opened 2022-01-07
Check if settings are changing before running area hooks

### LINKED ISSUE #1409: Using lock to usable area with multiple tablets can cause issues with tablet area display
## Description

In absolute output mode when switching between settings for tablets of different active area sizes the UX's area display can get changed. This only occurs when `Lock to usable area` is enabled on tablet area. 

For example:
Switching from a tablet with 159.6 x 99.75 mm active area using full area to a tablet with 152 x 95 mm active area then back to the first tablet will result in the first tablet's area being shrunk to 152 x 95 mm instead of staying at 159.6 x 99.75 mm.

This can also happen with area offset. Putting the tablet area somewhere on the larger tablet where the smaller tablet does not have space for then switching between the two tablets will result in the larger tablet's area being shifted over.
## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Windows 10
| OpenTabletDriver Version | 0.6.0 Pre-release

InfinityGhost:
Fixes #1409

### DIFF
--- a/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
+++ b/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
@@ -78,6 +78,7 @@ namespace OpenTabletDriver.UX.Controls.Output
 
         private bool handlingArLock;
         private bool handlingForcedArConstraint;
+        private bool handlingSettingsChanging;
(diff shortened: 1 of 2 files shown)


Kuuuube (approved):
Looks good, it works.

PR #1892, opened 2022-01-08
Add support for XP-Pen Artist 15.6

jamesbt365:
Verification: https://canary.discord.com/channels/615607687467761684/929767124187111444/929808745960177684

thanks cubix

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Artist 15.6.json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Artist 15.6.json
@@ -1,0 +1,55 @@
+{
+  "Name": "XP-Pen Artist 15.6",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 344.19,
+      "Height": 194.61,
+      "MaxX": 34419.0,
+      "MaxY": 19461.0
+    },
+    "Pen": {
+      "MaxPressure": 8191,
+      "Buttons": {
+        "ButtonCount": 2
+      }
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 6
+    },
+    "MouseButtons": null,
+    "Touch": null
+  },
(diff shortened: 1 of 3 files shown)


Mrcubix:
Output init: `ArAc`
Aux Init String: `4`

- This index doesn't look like it would do anything but aux isn't sent unless it is requested
- The tablet is Advertised as 200 lpmm but this is definitely proven wrong.
- The tablet would require it's own aux parser but i think we should add a new argument to `XP_PenAuxReport`'s constructor instead, being an index

```cs
using OpenTabletDriver.Plugin.Tablet;

namespace OpenTabletDriver.Configurations.Parsers.XP_Pen
{
    public struct XP_PenAuxReport : IAuxReport
    {
        public XP_PenAuxReport(byte[] report, int index)
        {
            Raw = report;

            AuxButtons = new bool[]
            {
                report[index].IsBitSet(0),
                report[index].IsBitSet(1),
                report[index].IsBitSet(2),
                report[index].IsBitSet(3),
                report[index].IsBitSet(4),
                report[index].IsBitSet(5),
                report[index].IsBitSet(6),
                report[index].IsBitSet(7),
                report[index + 1].IsBitSet(0),
            };
        }

        public bool[] AuxButtons { set; get; }
        public byte[] Raw { set; get; }
    }
}
```

As such, XP_PenReportParser would become this

```cs
using OpenTabletDriver.Plugin.Tablet;

namespace OpenTabletDriver.Configurations.Parsers.XP_Pen
{
    public class XP_PenReportParser : IReportParser<IDeviceReport>
    {
        public IDeviceReport Parse(byte[] report)
        {
            if (report[1].IsBitSet(4))
                return new XP_PenAuxReport(report, 2);
            else if (report.Length >= 10)
                return new XP_PenTabletReport(report);
            else
                return new TabletReport(report);
        }
    }
}
```

PR #1900, opened 2022-01-11
Backup settings file before migrating

Sublimelime:
Small PR to make the existing settings file be backed up to `settings.json.old` before migration happens.

### DIFF
--- a/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
+++ b/OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
@@ -18,6 +18,11 @@ namespace OpenTabletDriver.Desktop.Migration
         public static void Migrate(AppInfo appInfo)
         {
             var file = new FileInfo(appInfo.SettingsFile);
+
+            // Back up existing settings file for safety, in case migration goes wrong
+            var fileCopy = file.CopyTo(file.Name + ".old");
+            Log.Write("Settings", $"Old settings found, and have been backed up to {fileCopy.FullName}.");
+
             if (Migrate(file) is Settings settings)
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
```diff
@@ -18,6 +18,11 @@ public static class SettingsMigrator
         public static void Migrate(AppInfo appInfo)
         {
             var file = new FileInfo(appInfo.SettingsFile);
+
+            // Back up existing settings file for safety, in case migration goes wrong
+            var fileCopy = file.CopyTo(file.Name + ".old");
+            Log.Write("Settings", $"Old settings found, and have been backed up to {fileCopy.FullName}.");
```
InfinityGhost, line 24: This will run on every attempt to migrate settings, best if this log message is not included.

InfinityGhost (changes requested):

OpenTabletDriver.Desktop/Migration/SettingsMigrator.cs
Sublimelime: Alternatively I could resolve this by making it not copy multiple times if a recent backup already exists (and just print when it makes the first file) - this would also fix an exception that's going to be thrown, as per further reading into CopyTo:

>IOException

>An error occurs, **or the destination file already exists.**

Will iterate on this tomorrow.

InfinityGhost: This should instead be moved inside the `if` block below and remove the logging.

PR #1906, opened 2022-01-12
Add device identifier conflict detection test

X9VoiD:


Sublimelime:
It may be wise to make this **only run on PRs to the configs branch**, otherwise it may be running redundantly (for changes that don't even touch configs or anything related) and slowing down CI.

X9VoiD:
Windows CI runs significantly slower than the unit test, and they all run in parallel. As seen in this PR it ran even faster than `dotnet format`

InfinityGhost (approved):

PR #1909, opened 2022-01-12
Add build script for Windows

X9VoiD:
Mirrors the behavior of `build.sh` except for some changes:

- Bring back current working directory to previous working directory after build.
- Build in debug mode.
- Single file, embedded debugging symbols.


gonX:
I would prefer if the configuration (Release/Debug) were the same for both for consistency's sake.

InfinityGhost (approved):

PR #1936, opened 2022-01-13
Add support for XP-Pen Deco 01 V2 (variant 2)

jamesbt365:
Verification: https://discord.com/channels/615607687467761684/930837536585187360/930961622250434601

This tablet is a V2 V2 due to XP-Pen calling this tablet a V2 and this is a variant of this, so therefore this is a V2 V2, however their may be a better way of naming this, but i'm following previous variants that have been added.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco 01 V2 V2.json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco 01 V2 V2.json
@@ -1,0 +1,43 @@
+{
+  "Name": "XP-Pen Deco 01 V2 V2",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 254.0,
+      "Height": 158.75,
+      "MaxX": 50800.0,
+      "MaxY": 31750.0
+    },
+    "Pen": {
+      "MaxPressure": 8191,
+      "Buttons": {
+        "ButtonCount": 2
+      }
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 8
+    },
+    "MouseButtons": null,
+    "Touch": null
+  },
+  "DigitizerIdentifiers": [
+    {
+      "VendorID": 10429,
+      "ProductID": 2309,
+      "InputReportLength": 12,
+      "OutputReportLength": 10,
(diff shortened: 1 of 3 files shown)


OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco 01 V2 V2.json
```diff
@@ -0,0 +1,43 @@
+{
+  "Name": "XP-Pen Deco 01 V2 V2",
```
InfinityGhost, line 2: I really don't know how to feel about this name, might be better to call it something like `XP-Pen Deco 01 V2 Rev B` or anything that could help differentiate it. Really annoying that manufacturers do this an don't update their model names.

X9VoiD: Should probably change scheme to `Tablet Model (variant 1)`

Since these are not guaranteed a revision, nor a version, unless specified by OEM. This scheme rolls off the tongue much better too for this tablet.

`XP-Pen Deco 01 V2 (variant 1)`
`XP-Pen Deco 01 V2 (variant 2)`

We may hide variant name from UX.

Or just integrate spec variants directly in configs. Depending on traction here we'll raise this to a new feature issue.

InfinityGhost: For now I'm in favor of referring to it as such 
```suggestion
  "Name": "XP-Pen Deco 01 V2 (variant 2)",
```
TODO in the future to add a better system for this variant issue.

TABLETS.md
```diff
@@ -145,6 +145,7 @@
 | XP-Pen Artist Pro 12          |  Missing Features | Tilt and wheel are unsupported.
 | XP-Pen Artist Pro 15.6        |  Missing Features | Tilt and wheel are unsupported.
 | XP-Pen Deco 01 V2             |  Missing Features | Tilt is unsupported.
+| XP-Pen Deco 01 V2 V2          |  Missing Features | Tilt is unsupported
```
InfinityGhost, line 148: Fix name

InfinityGhost (changes requested):

InfinityGhost (approved):

PR #1939, opened 2022-01-14
Update source selector DataStore on profile change

### LINKED ISSUE #1925: Filters tab is not refreshed when switching tablets
## Description

Filters tab still shows the settings for the previously active tablet after selecting a filter when switching tablets instead of resetting or showing that filter's settings for the currently active tablet. The user must select another filter then go back to the previous selected filter to get it to refresh. However, this is not possible when there is only one filter installed and it just gets stuck.

Reproduction steps:
1. Have two or more tablets plugged in
2. Have a plugin with a filter installed
3. Open the filters tab and change some of the settings for one tablet
4. Switch to the other tablet

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Win10
| OpenTabletDriver Version | 0.6.0/0.6.0.1

InfinityGhost:
Fixes #1925

### DIFF
--- a/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
+++ b/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
@@ -5,7 +5,6 @@ using Eto.Drawing;
 using Eto.Forms;
 using OpenTabletDriver.Desktop;
 using OpenTabletDriver.Desktop.Reflection;
-using OpenTabletDriver.UX.Controls.Generic;
 using OpenTabletDriver.UX.Controls.Generic.Reflection;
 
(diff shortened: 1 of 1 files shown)


Kuuuube (changes requested):
This doesn't appear to have done anything for #1925. The behavior appears the be the exact same. 

The filters tab either needs to be refreshed for the filter selected to show the settings for the selected tablet or deselect filters entirely when switching tablets.

Tested on Win10 with Wacom PTH-460 and CTL-471.

InfinityGhost:
All I can assume is that its not updating the selected index when refreshing, however it does this on GTK.

Kuuuube (approved):
Works.

Tested on Win10.

Issue #1717, opened 2021-12-11
Wacom PTZ-x30/31W FeatureInitReport does not work properly


Kuuuube:
## Description

The current Wacom PTZ-x30/31W does not actually init the tablet in any working way. Inits are accepted but they seem to go nowhere. 

`AgI=` init alone does init the tablet properly into 100hz mode. `AgI=` as the last init sent regardless of inits before it does init the tablet properly into 100hz mode. Any combination of inits that does not put `AgI=` as the last init or as the only init, does not init the tablet at all.

Sending `AgI=` alone then `BAA=` alone (detecting tablet between two different configs without replugging) OR plugging in the tablet then sending `AgI=` and `BAA=` together (detecting tablet with a two configs the first with no init the second with both) inits the tablet properly into 200hz mode. 

`BAA=` does nothing on its own and must be paired with `AgI=`.

`CAIDAAA=` appears to do nothing regardless of what it is paired with.

The current config seems to be sending inits too quickly after plugging in the tablet and they get nullified somehow.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| OpenTabletDriver Version | 4b9e63f
| Tablet                   | Wacom PTZ-x30/31W

gonX:
Would a config attribute enabling delayed send of multiple inits solve this issue? And does the tablet respond with anything after the first init that we can wait for?

gonX:
Talked with Kuuube about it, sounds like if we have to meet this for 0.6.x, simply delaying additional reports by a second should be fine - the side effect is really just a slight delay in activating higher poll rates.

Kuuuube:
Not sure if this was conveyed properly in the initial post but if this is not fixed people using Wacom PTZ-x30/31W tablets will not have a working tablet. As a workaround they can detect tablet a second time every time they start OTD but that is very annoying. 

Alternatively, configs could be stripped of `BAA=` and `CAIDAAA=` inits and let the tablet go into 100hz mode instead of trying for 200hz. This will allow for the tablets to work without any extra action from the user but only 100hz... that option isn't great but **something** needs to be done about this before 0.6.0 releases.

Kuuuube:
Putting this back on the 0.6.0 milestone. 

As I mentioned in my previous comment something needs to be done about this before 0.6.0. Even if that means we revert back to only using `AgI=` inits for the Wacom PTZ-x30/31W. If this done then this issue can be put back on the backlog.

[...]

PR #1943, opened 2022-01-14
InputDevice: Add support for feature initialization delay

gonX:
This is one of many commits that's going to help out with #1717 - this enables the usage of the tablet configuration attribute `FeatureInitDelayMs`

### DIFF
--- a/OpenTabletDriver/InputDevice.cs
+++ b/OpenTabletDriver/InputDevice.cs
@@ -31,7 +31,7 @@ namespace OpenTabletDriver
 
             if (Configuration.Attributes.TryGetValue(DelayAttributeKeyName, out var delayStr))
                 if (!UInt32.TryParse(delayStr, out FeatureInitDelayMs))
-                    Log.Debug("Device", $"Could not parse '{delayStr}' from attribute {DelayAttributeKeyName}");
+                    Log.Write("Device", $"Could not parse '{delayStr}' from attribute {DelayAttributeKeyName}", LogLevel.Warning);
 
             Start();
         }
--- a/OpenTabletDriver/InputDevice.cs
+++ b/OpenTabletDriver/InputDevice.cs
@@ -10,8 +10,8 @@ namespace OpenTabletDriver
 {
     public class InputDevice : DeviceReader<IDeviceReport>
     {
-        private uint FeatureInitDelayMs;
-        private readonly string DelayAttributeKeyName = "FeatureInitDelayMs";
+        private readonly uint _featureInitDelayMs;
+        private const string DELAY_ATTRIBUTE_KEY_NAME = "FeatureInitDelayMs";
 
         public InputDevice(IDriver driver, IDeviceEndpoint device, TabletConfiguration configuration, DeviceIdentifier identifier)
             : base(device, driver.GetReportParser(identifier))
@@ -29,9 +29,9 @@ namespace OpenTabletDriver
             Configuration = configuration;
             Identifier = identifier;
 
-            if (Configuration.Attributes.TryGetValue(DelayAttributeKeyName, out var delayStr))
-                if (!UInt32.TryParse(delayStr, out FeatureInitDelayMs))
-                    Log.Write("Device", $"Could not parse '{delayStr}' from attribute {DelayAttributeKeyName}", LogLevel.Warning);
+            if (Configuration.Attributes.TryGetValue(DELAY_ATTRIBUTE_KEY_NAME, out var delayStr))
+                if (!uint.TryParse(delayStr, out _featureInitDelayMs))
+                    Log.Write("Device", $"Could not parse '{delayStr}' from attribute {DELAY_ATTRIBUTE_KEY_NAME}", LogLevel.Warning);
 
             Start();
         }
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/InputDevice.cs
```diff
@@ -25,6 +29,10 @@ public InputDevice(IDriver driver, IDeviceEndpoint device, TabletConfiguration c
             Configuration = configuration;
             Identifier = identifier;
 
+            if (Configuration.Attributes.TryGetValue(DelayAttributeKeyName, out var delayStr))
+                if (!UInt32.TryParse(delayStr, out FeatureInitDelayMs))
+                    Log.Debug("Device", $"Could not parse '{delayStr}' from attribute {DelayAttributeKeyName}");
```
Sublimelime, line 34:
```suggestion
                    Log.Write("Device", $"Could not parse '{delayStr}' from attribute {DelayAttributeKeyName}", LogLevel.Warning);
```

This should be Warning not Debug, as it shouldn't happen if the config is working.

gonX:
I can't find the Discord conversation right now but I seem to recall that the affected user has a functioning fix by simply waiting after initializing the string index.

That means the sleep should be done _before_ running SetFeature. This would also fix the unneeded last sleep the current version of the code does.

OpenTabletDriver/InputDevice.cs
```diff
@@ -67,6 +78,8 @@ protected override bool Initialize()
                 {
                     ReportStream.SetFeature(report);
                     Log.Debug("Device", "Set device feature: " + BitConverter.ToString(report));
+                    if (FeatureInitDelayMs != 0)
+                        Thread.Sleep((int)FeatureInitDelayMs);
```
X9VoiD, line 82:
```suggestion
                    if (FeatureInitDelayMs != 0)
                        Thread.Sleep((int)FeatureInitDelayMs);
                    ReportStream.SetFeature(report);
                    Log.Debug("Device", "Set device feature: " + BitConverter.ToString(report));
```
There should be a delay in between the GetString call and SetFeature.

X9VoiD (changes requested):
Probably also worth it to generalize the attribute instead to `InitializationDelayMs` and add the delay in all calls to GetString, SetFeature and Write within `Initialize()`.

OpenTabletDriver/InputDevice.cs
```diff
@@ -52,6 +60,9 @@ protected override bool Initialize()
             Log.Debug("Device", $"Initializing device '{Endpoint.FriendlyName}' {Endpoint.DevicePath}");
             Log.Debug("Device", $"Using report parser type '{Identifier.ReportParser}'");
 
+            if (FeatureInitDelayMs != 0)
```
Sublimelime, line 63: Is this check even necessary? Does calling `Thread.Sleep(0);` result in undefined behavior? Theoretically if it doesn't cause issues it'd be better to just always call it, and if it's 0, it just sleeps for 0, which _should_ be a no op? Unless that's not allowed, I haven't looked at the Thread.Sleep docs.

vedattt: It looks like you meant to comment on line 81. This check right here is in fact necessary to avoid a redundant log message.

vedattt: Anyway, the check may be significant because `Sleep(0)` is not a no op statement. As per the docs, it yields execution to any thread of equal priority waiting to run.

jamesbt365:
This pr works with the PTZ-630 (thanks @OniCado) with 125ms delay (115 works as well). This was tested with Voids patch on top of this.

gonX:
I prefer testing this without VoiD's patch first. We don't want to add more lines than necessary.

gonX:
Marking as ready btw since I'm pretty sure it should work

X9VoiD (approved):

Kuuuube (changes requested):
Edit: I misunderstood what this PR is doing but if it is just adding a new option to configs it should either make the changes to fix #1717 or have a PR depending on it that fixes #1717.

~~Does not fix the issue. I assume this was not tested with a config that had multiple inits or it was not tested plugging in a tablet while otd is running.~~

~~Those are the two things this pr should fix.~~

~~For further testing:~~
~~1. These are configs this should be tested with:
[configurations.zip](https://github.com/OpenTabletDriver/OpenTabletDriver/files/8423299/configurations.zip)~~
~~2. The tablet should only be plugged in while otd is running. Plugging in the tablet then starting otd will not replicate the issue in #1717.~~

Kuuuube (approved):
Tested this with PTZ-431W and was able to get this working with as low as 70ms. Not sure what the time needed depends on. Would be good to have more testing and this really should have a pr based off it for the needed config changes.

OpenTabletDriver/InputDevice.cs
```diff
@@ -9,6 +10,9 @@ namespace OpenTabletDriver
 {
     public class InputDevice : DeviceReader<IDeviceReport>
     {
+        private uint FeatureInitDelayMs;
+        private readonly string DelayAttributeKeyName = "FeatureInitDelayMs";
```
InfinityGhost, line 14: Don't directly apply this suggestion, refactor from within your IDE.
```suggestion
        private uint _featureInitDelayMs;
        private const string DELAY_ATTRIBUTE_KEY_NAME = "FeatureInitDelayMs";
```

InfinityGhost (changes requested):

X9VoiD:
Resolved by #2183

PR #1950, opened 2022-01-15
EvdevVirtualTablet: Code cleanup

gonX:
This deduplicates some code and makes future expansion of supported event codes a bit more resilient


gonX:
Will be superseded in an upcoming PR

PR #1973, opened 2022-01-17
Change parser for Wacom CTE-450

imkunet:
The CTE-450 configuration uses a parser which does not work for this particular tablet.

The Intuos report parser does not work for the Bamboo tablet and is the change of this pull request.

Side note: This configuration error is also present for the CTE-650 but is indeed out of scope for this pull request *[see here for more details](https://discord.com/channels/615607687467761684/932116307674021919)*.

- [x] I have personally tested this modification to work on my own CTE-450


gonX:
Looks to me from the Discord thread you linked that the CTE-650 should indeed have that change applied to it as well. I would be fine with having the PR have this as well, or making a separate with the discord thread as documentation.

gonX (changes requested):
As this is a configuration change this should target the `configs` branch as per [CONTRIBUTING.md](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/CONTRIBUTING.md)

PR #1975, opened 2022-01-17
Fix string in Huion 420 Config

Kuuuube:
Huion thought it would be an amazing idea to make variations of the 420 with spaces after string 6...

Verification: 
https://discord.com/channels/615607687467761684/615611007951306863/932479275095961630
https://discord.com/channels/615607687467761684/615611007951306863/932474098758602824


InfinityGhost:
This is actually stupid.

InfinityGhost (approved):

PR #1976, opened 2022-01-17
On MacOS, prevent user from launching the app when it is translocated.

AkiSakurai:
The app doesn't work properly with App Translocation [^1].

[^1]: https://lapcatsoftware.com/articles/app-translocation.html


OpenTabletDriver.UX.MacOS/Program.cs
```diff
@@ -7,6 +8,18 @@ class Program
...
         {
+            var baseDirectory = AppDomain.CurrentDomain.BaseDirectory;
+
+            if (baseDirectory.StartsWith("/private/var/folders"))
+            {
+                _ = new Application(Eto.Platforms.Mac64);
+                MessageBox.Show(
+                    "OpenTabletDriver cannot be run from this folder. " +
+                    "Please move it to elsewhere, " +
+                    "such as to the Applications folder.");
```
vedattt, line 19:
```suggestion
                    "OpenTabletDriver cannot be run from this folder (/private/var/folders). " +
                    "Please move it elsewhere, such as to the Applications folder.");
```
Rewording and clarification.

vedattt (changes requested):

OpenTabletDriver.UX.MacOS/Program.cs
AkiSakurai: The path /private/var/folders is invisible to the user.

From the users point of view, the app is still in Downloads folder. 

The system silently copy the app to that folder and launch from there if the app is downloaded from internet.

vedattt: Ah, I see, I'm not really familiar with macOS.
Probably still worth considering clarifying it without the path though:
```suggestion
                    "OpenTabletDriver cannot be run from this folder for security reasons. " +
                    "Please move it elsewhere, such as to the Applications folder.");
```

vedattt (approved):

OpenTabletDriver.UX.MacOS/Program.cs
```diff
@@ -7,6 +8,17 @@ class Program
...
         public static void Main(string[] args)
         {
+            var baseDirectory = AppDomain.CurrentDomain.BaseDirectory;
+
+            if (baseDirectory.StartsWith("/private/var/folders"))
+            {
+                _ = new Application(Eto.Platforms.Mac64);
+                MessageBox.Show(
+                    "OpenTabletDriver cannot be run from this folder for security reasons. " +
+                    "Please move it elsewhere, such as to the Applications folder.");
```
InfinityGhost, line 18:
```suggestion
                MessageBox.Show("OpenTabletDriver must be installed to the Applications folder.");
```

InfinityGhost (changes requested):

gonX:
need rebase

jamesbt365 (approved):

PR #1979, opened 2022-01-17
IntuosV1TabletReport: Match upstream distance

gonX:
Measuring the stock Wacom driver for my GD-0608 reveals that our distance was twice as high as it should've been

This change makes it report the same distance as the upstream Wacom driver


gonX:
Converted to draft as this change is incorrect for some of the affected tablets. It's correct on first gen Intuos, though.

gonX:
Closing as it seems like this behavior has been changed again.

PR #1982, opened 2022-01-18
EvdevVirtualTablet: Change some consts to public

gonX:
We 'need' this for an upcoming libinput quirks generator


gonX:
It has been suggested that this is done as part of the quirks generator instead

PR #1983, opened 2022-01-18
Libinput quirks generator

### LINKED ISSUE #1514: Consider working around libinput's smoothing
## Description
libinput has smoothing disabled for tablets with AES pens, which means it's likely possible for OTD to fake this information in order to bypass libinput's smoothing, eliminating the need for the libinput patch that removes the smoothing (https://gitlab.freedesktop.org/libinput/libinput/-/issues/225)

The changes in libinput to detect AES pens is here:
https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/512/diffs?commit_id=a506d092b8fc5d05c0ef7579b58d36be20b2c2af

Doing this would allow us to more accurately control the input

### LINKED ISSUE #1680: libinput has its own pressure offset that we should work around
## Description

When in Linux Artist Mode, I need to apply more than 5% of my threshold (post-threshold slider) before libinput registers it as a tip down event.

~~Since OTD is powerful enough to have a threshold slider and remap the values (so that 51% pressure on a 50% threshold reads as 2%), I think that's a good enough reason for us to want to circumvent this tip detection.~~ ~~This is not the case on newer versions.~~ Tested working on v0.6.5.1, so this should be reconsidered and _retested_ on `avalonia`/`master`.

The reason why this pressure offset exists in libinput is explained here: https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#pressure-offset-on-worn-out-tools

A workaround is to add the `AttrPressureRange` libinput quirk for the tablet, but that means it can only be fixed in packaging.

[libinput device quirks](https://wayland.freedesktop.org/libinput/doc/latest/device-quirks.html) states:
> AttrPressureRange=N:M
>
>    Specifies the touch pressure required to trigger a press (N) and to trigger a release (M)

This might effectively not be an issue if we're interested in keeping the libinput behavior.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Linux (Arch)
| OpenTabletDriver Version | 53a7dcf
| Tablet                   | Assumed all (but GD-0608)

### LINKED ISSUE #2895: Discussion: Disable Artist Mode smoothing with libinput quirks file
### Description

By default, libinput smooths all tablet input, even if it is smoothed by OpenTabletDriver already.

I propose disabling the smoothing by adding the following quirks to begin with:

```
[OpenTabletDriver Virtual Tablets]
MatchName=OpenTabletDriver*
AttrTabletSmoothing=0
```

The file path should be `/usr/share/libinput/30-vendor-opentabletdriver.quirks`

---

**Arguments for disabling libinput smoothing by default:**
- Raw(er) input
- Artist mode now handles like absolute mode (as absolute mode is detected as an absolute mouse rather as an tablet device)
- More users are likely interested keeping it disabled, rather than having it enabled.

**Valid arguments against disabling the smoothing provided by libinput:**

- Does not match upstream behavior with other tablets (ie. tablet will feel different from stock driver)
- Can make further configuration (e.g. OpenTabletDriver plugins) mandatory for chattery or otherwise inaccurate tablets, adding to an already somewhat complicated install process

---

Sadly exposing this as an easy user-controlled setting is ambitious as it seems we can only do this in packaging, requiring the user to have root permissions to set local quirk overrides, or root permissions to repackage with their preferred settings..

Down the line I would like to see something like https://github.com/OpenTabletDriver/OpenTabletDriver/pull/1983 being implemented as well, so this would just be a stopgap solution until then (though, the user would still need root). With that in place, users would be able to generate their own quirks file that could re-enable the smoothing if they so desire.

(loosely copied from OpenTabletDriver/OpenTabletDriver.Packaging#22)

### Acknowledgements

gonX:
This is going to be useful (and hopefully mandatory) for packagers to ship (and adjust) a libinput quirks file. 

This fixes #1514, fixes #1680 and fixes #2895.

I've left the option in should an expert user decide to keep it

Unrelated to the PR but we should probably check for the presence of this file, or probe quirk behavior for virtual tablet endpoint on daemon start so that we can ensure behavior is as expected.

If anyone is aware of setting these quirks _in code_ instead of in packaging this would be greatly appreciated, as it'd ensure that we can get a consistent experience across all distros without having to worry about packaging.

Packagers/maintainers are advised to put this quirks file as `/usr/share/libinput/30-vendor-opentabletdriver.quirks` to match existing libinput naming convention.

Run example:

```
$ dotnet run -- testfile.quirks
Writing quirks to 'testfile.quirks'
$ cat testfile.quirks 
[OpenTabletDriver]
MatchName=OpenTabletDriver Virtual Artist Tablet
AttrTabletSmoothing=0
```

There is also support for setting pressure ranges (libinput quirk `AttrPressureRange`) but defaulting to any value at the moment seems unreasonable, so we should just let libinput handle button debouncing (the default).

- [x] merge prevention checkbox because there are reviews i need to address
- [ ] check whether it's necessary to update `README.md`, and if it is, do so
- [ ] update packaging script to include this

### DIFF
--- a/OpenTabletDriver.Tools.LibinputQuirks/Program.cs
+++ b/OpenTabletDriver.Tools.LibinputQuirks/Program.cs
@@ -38,6 +38,8 @@ namespace OpenTabletDriver.Tools.LibinputQuirks
 
         static async Task WriteQuirksAsync(TabletAttributes attr, FileInfo output, bool verbose)
         {
+            if (attr.TipUpPressurePermille.HasValue ^ attr.TipDownPressurePermille.HasValue)
+                Console.WriteLine("WARNING: Both Tip-down and Tip-up must be specified! Ignoring pressures.");
 
             if (attr.TipUpPressurePermille.HasValue
                     && attr.TipDownPressurePermille.HasValue
--- a/OpenTabletDriver.Tools.LibinputQuirks/Program.cs
+++ b/OpenTabletDriver.Tools.LibinputQuirks/Program.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.CommandLine;
+using System.CommandLine.NamingConventionBinder;
+using System.CommandLine.Invocation;
+using System.IO;
+using System.Threading.Tasks;
+using OpenTabletDriver.Desktop.Interop.Input.Absolute;
+
+namespace OpenTabletDriver.Tools.LibinputQuirks
+{
+    class Program
+    {
+        static readonly bool DEFAULT_SMOOTHING = false;
+        static readonly uint DEFAULT_TIPDOWN_PRESSURE_PERMILLE = 10;
+        static readonly uint DEFAULT_TIPUP_PRESSURE_PERMILLE = 9;
+        static readonly string LIBINPUT_HEADER = "[OpenTabletDriver]";
+        static readonly string CONFIG_KEY_NAME_STRING = "MatchName";
+        static readonly string CONFIG_KEY_TABLET_SMOOTHING = "AttrTabletSmoothing";
+        static readonly string CONFIG_KEY_PRESSURE_RANGE = "AttrPressureRange";
+
+        static void Main(string[] args)
+        {
+            var root = new RootCommand("OpenTabletDriver libinput quirks generator")
+            {
+                new Option<uint>(new[] { "--tip-down-pressure-permille", "-pd" }, () => DEFAULT_TIPDOWN_PRESSURE_PERMILLE, "Pressure threshold for tip down in permille (1/1000)"),
+                new Option<uint>(new[] { "--tip-up-pressure-permille", "-pu" }, () => DEFAULT_TIPUP_PRESSURE_PERMILLE, "Pressure threshold for tip up in permille (1/1000)"),
+                new Option<bool>(new[] { "--smoothing", "-s" }, () => DEFAULT_SMOOTHING, "Allow libinput tablet smoothing"),
+                new Option(new[] { "--no-pressure", "-np" }, $"Do not include '{CONFIG_KEY_PRESSURE_RANGE}'"),
+                new Argument<FileInfo>("output", "Resulting .quirks file"),
+                new Option(new[] { "--verbose", "-v" }, "Verbose output")
+            };
+
+            root.Handler = CommandHandler.Create<uint, uint, bool, bool, FileInfo, bool>(WriteQuirksAsync);
+            root.Invoke(args);
+        }
+
+        static async Task WriteQuirksAsync(uint tipDownPressurePermille, uint tipUpPressurePermille, bool smoothing, bool skipPressure, FileInfo output, bool verbose)
+        {
+            if (tipUpPressurePermille >= tipDownPressurePermille) throw new ArgumentOutOfRangeException("Tip-up pressure must be less than tip-down pressure");
+
+            bool unclean = false;
+
+            var path = output.FullName.Contains(Directory.GetCurrentDirectory()) ? output.Name : output.FullName;
+            Console.WriteLine($"Writing quirks to '{path}'");
+
+            var filenameContainsQuirks = !output.Name.EndsWith(".quirks");
+            if (filenameContainsQuirks)
+                Console.WriteLine("WARNING: File name does not end with '.quirks'. Your file will not be parsed by libinput!");
+            unclean |= filenameContainsQuirks;
+
+            if (output.Exists)
+                output.Delete();
+            if (!output.Directory.Exists)
+                output.Directory.Create();
+
+            using (var sw = output.AppendText())
+            {
+                if (verbose) Console.WriteLine("Adding header");
+                await sw.WriteLineAsync(LIBINPUT_HEADER);
+                foreach (var line in CreateQuirks(smoothing, skipPressure, tipDownPressurePermille, tipUpPressurePermille))
+                {
+                    if (verbose) Console.WriteLine($"Adding {line.Substring(0,line.IndexOf("="))}");
+                    await sw.WriteLineAsync(line);
+                }
+            }
+            if (unclean) Console.WriteLine("WARNING: The generator encountered warnings and your file may not work correctly");
+        }
+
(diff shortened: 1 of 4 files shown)


gonX:
This needs to also update README.md similar to how ./generate-rules.sh did

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            bool unclean = false;
+
+            var path = output.FullName.Contains(Directory.GetCurrentDirectory()) ? output.Name : output.FullName;
+            Console.WriteLine($"Writing quirks to '{path}'");
+
+            var filenameContainsQuirks = !output.Name.EndsWith(".quirks");
+            if (filenameContainsQuirks)
+                Console.WriteLine("WARNING: File name does not end with '.quirks'. Your file will not be parsed by libinput!");
+            unclean |= filenameContainsQuirks;
```
Sublimelime, line 49: I dislike this - instead of having that conditional in a boolean, just put that directly into the if conditional. The "unclean" thing is also not good - not only will this redundantly log (if filenameContainsQuirks is false, you mention it won't be parsed by libinput.... so fix that for them, or fail out lmao - plus you then use "unclean" to log again that there might be issues - they already know.)

I would simply check if the filename ends in .quirks, and if it doesn't, append it for them. Simple. Nix the unclean thing too, we should either autofix errors like the above, or just fail out. Generating a likely useless file is bad.

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+                new Option(new[] { "--no-pressure", "-np" }, $"Do not include '{CONFIG_KEY_PRESSURE_RANGE}'"),
+                new Argument<FileInfo>("output", "Resulting .quirks file"),
+                new Option(new[] { "--verbose", "-v" }, "Verbose output")
+            };
+
+            root.Handler = CommandHandler.Create<uint, uint, bool, bool, FileInfo, bool>(WriteQuirksAsync);
+            root.Invoke(args);
+        }
+
+        static async Task WriteQuirksAsync(uint tipDownPressurePermille, uint tipUpPressurePermille, bool smoothing, bool skipPressure, FileInfo output, bool verbose)
```
Sublimelime, line 37: Large amount of parameters here, and makes line 33 above hard to read too with it's long "<uint, uint, bool, bool....".

I would collapse the first 4 into a struct that you could pass to it instead. Something like QuirkDetails or something.

Sublimelime (changes requested):

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
gonX: Ok. I'll do a class though. I don't think a struct makes sense, we're not really modifying any values. So a class should be more performant - not that it matter though.

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
X9VoiD: `unclean` variable here is used just in case there will be future other "warnings".

The best compromise here imo is to notify user of the rename but not as a warning, and therefore it won't be "unclean".

I see an exception a few lines before this, and I'd argue using the unclean thing for that instead of throwing an exception there to keep the `unclean`. Humans parse sentences better than a sentence in a stack trace.

gonX: I don't like assuming things about a users environment. It's mostly meant to be used by packagers but with the flexibility of being usable on random distros having niche requirements or doing things differently. The udev tool has no similar check and will fail in a similar way if not named correctly, so I think it's better to err on the side of caution here.



As for the `unclean` variable that mostly exists if further error checking is needed down the line, then the final error check line does not look messy.

If there is more opposition to `unclean` I'll remove it.

Btw, since this is 2 issues in 1 conversation it's hard for me to really resolve this.

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
Sublimelime: Sure, whatever works, In c++ structs and classes are basically the same thing with the same performance but structs are used for simple groups of variables - I'm used to c++ so I guess use whatever the c# equivalent is.

doubledge-spec:
update?

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+                        GeneratePressureRange(
+                            PressureConvertPermilleToUnits(tipDownPressurePermille),
+                            PressureConvertPermilleToUnits(tipUpPressurePermille)
+                        ));
+            }
+        }
+
+        static string GenerateKeyValue(string left, string right) => left + "=" + right;
+
+        static string GeneratePressureRange(uint tipDownPressure, uint tipUpPressure) => tipDownPressure + ":" + tipUpPressure;
```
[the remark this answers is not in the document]
gonX: Resolved in e3e5cf8ae

OpenTabletDriver.Tools.LibinputQuirks/Program.cs
gonX: Both resolved in 7e24c4c652 - I personally don't like the current implementation since the output name is "magically" changed. I'm not sure the current warning message is clear enough, so please take a look.

Ironically, the `unclean` variable may have a comeback to indicate that changes have occurred, at the end of the program.

gonX:
Rebased and resolved convos.

Remaining steps will be added to the OP.

I've opted to remove the defaulting of the pressure range - this seems unfeasible to set with the current implementation of OpenTabletDriver and that particular feature has seen 0 interest from anyone as far as I know, but if nothing else it is reasonable to leave it in should a user want to change this with some convenience (as it's set referenced to the pressure range OTD will report)

Sublimelime (approved):
Looks solid and readable. Little bit out of the loop, but nothing jumps out to me as wrong, for what it's worth.

gonX:
Thinking about it, this is probably better as a standalone application. Only real use is that we can grab the virtual tablet name and its max pressure from code, but we can honestly do both of those too by just having it loosely copy+pasted and then using some sort of test script to ensure it matches whenever a PR changes this, without the maintenance burden that is a full CLI app.

PR #1984, opened 2022-01-18
Artistmode revamp

gonX:
This cleans up and enhances some code related to Linux Artist Mode. I'm not doing a big writeup like last time but will try to argue for some of the changes.

## Add and implement `IToolReport`
This allows tool serial and ID to be reported on supported tablets (currently only IntuosV1 parser)

There's some quirkiness with tool ID sometimes not submitting correctly. Debugging the code makes me think the `ABS_MISC` report is getting dropped at some point in uinput because even `evtest` isn't always seeing it. We could also be emitting it incorrectly, someone please help find the relevant documentation for this.

Stock Wacom driver IIRC has worse functionality than this so it is at least an upgrade over that, but we may consider not emitting `ABS_MISC` for the time being.

## Don't send `BTN_TOUCH` on nonzero pressure
It's more intuitive for the user to have to bind a button, so we're moving the responsibility of emitting `BTN_TOUCH` to a binding handler instead.

In practice, libinput was handling this fine (despite my initial pessimism)

## Reintroduce pen tip support in `LinuxArtistModeButtonBinding`
This mostly exists because some users have gotten confused over no bindings being the correct solution

While we still can't rewrite pressure, this at least allows the user to control when `BTN_TOUCH` is emitted which should hopefully cover any remaining edge cases with incorrect tip down detection

## Allow "usage" of the 3 standard mouse buttons in artist mode
This allows users to switch from absolute mode to artist mode without being required to change their bindings

However, we are sending the tablet-equivalent events to mouse buttons, and not actually sending LMB/RMB/etc events.

This is because libinput silently drops mouse events from endpoints that passes the libinput tablet sanity check.

The user is incorrectly still able to bind Forward/Backwards mouse buttons which are unimplemented. I have not tested passing those buttons through, however.

**This part of the commit needs extra testing under X11 and other Wayland compositors that isn't Sway.**

## Expose `ABS_DISTANCE` to uinput
For now, I've set a maximum height of 100 as I doubt any tablet is going to go above this. `libwacom` seems to assign the max on a per-tablet basis.

The only way to enhance this would be if we knew the max distance of the tablet ahead of time, similar to how we have MaxPressure at the moment.

libinput seems to drop this, but the same seems to be the case with the upstream driver, so I don't think we're doing anything wrong in that regard. It's still visible in evdev.

## Drop the lift-off height restriction
This previously existed because we needed some way to report negative tool proximity. After we've had `Reset()` and `OutOfRangeReport`s introduced this is no longer necessary.

We have no good events to report for `IProximityReport`s so it's just returning nothing for now and upstream driver seems to have similar behavior (that, or I missed the event it's emitting).

- [ ] this will be extended to also remove `IProximityReport` eventually

# Footnotes

Test test test even if you think your setup is generic

# Known issues that needs to be fixed before merge:

- [ ] Pen tip gets stuck sometimes, seems like a libinput issue. Hard to pinpoint what exactly is causing it, since it seems to be a amalgamation of the commits that provoke the behavior.

### DIFF
--- a/OpenTabletDriver.Desktop/Binding/LinuxArtistMode/LinuxArtistModeButtonBinding.cs
+++ b/OpenTabletDriver.Desktop/Binding/LinuxArtistMode/LinuxArtistModeButtonBinding.cs
@@ -16,7 +16,8 @@ namespace OpenTabletDriver.Desktop.Binding.LinuxArtistMode
         public static string[] ValidButtons { get; } = {
             "Pen Button 1",
             "Pen Button 2",
-            "Pen Button 3"
+            "Pen Button 3",
+            "Pen Tip"
         };
 
         [Property("Button"), PropertyValidated(nameof(ValidButtons))]
@@ -39,6 +40,7 @@ namespace OpenTabletDriver.Desktop.Binding.LinuxArtistMode
                 "Pen Button 1" => EventCode.BTN_STYLUS,
                 "Pen Button 2" => EventCode.BTN_STYLUS2,
                 "Pen Button 3" => EventCode.BTN_STYLUS3,
+                "Pen Tip" => EventCode.BTN_TOUCH,
                 _ => throw new InvalidOperationException($"Invalid Button '{Button}'")
             };
 
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -50,6 +50,7 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
             var pressure = new input_absinfo
             {
+                flat = (int)(MaxPressure * 0.005d),
                 maximum = MaxPressure
             };
             input_absinfo* pressurePtr = &pressure;
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -102,7 +102,6 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
         public void SetPressure(float percentage)
         {
-            Device.Write(EventType.EV_KEY, EventCode.BTN_TOUCH, percentage > 0 ? 1 : 0);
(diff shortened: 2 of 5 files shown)


OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
```diff
@@ -113,7 +129,8 @@ public void SetEraser(bool isEraser)
 
         public void SetProximity(bool proximity)
         {
-            this.proximity = proximity;
+            // matches upstream behavior; proximity isn't exposed to userspace
+            return;
```
Sublimelime, line 133: This shouldn't fail silently - if this is supposed to be impossible, we should either straight up remove the function or `throw new NotSupportedException`, unless I'm misunderstanding.

Just silently returning in a void function without doing anything spells hard to find breaks later.

Sublimelime (changes requested):

OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
gonX: Ah, I forgot after the rework X9VoiD did that we now have `IProximityHandler`, so I'll just not inherit from that interface

gonX:
I am tracking down some issue with stuck pen tip so this should not be merged for now

OpenTabletDriver.Desktop/Binding/LinuxArtistMode/LinuxArtistModeButtonBinding.cs
```diff
@@ -16,7 +16,8 @@ public class LinuxArtistModeButtonBinding : IStateBinding
         public static string[] ValidButtons { get; } = {
             "Pen Button 1",
             "Pen Button 2",
-            "Pen Button 3"
+            "Pen Button 3",
+            "Pen Tip"
```
X9VoiD, line 20: Put "Pen Tip" as the first valid button.

gonX:
X9VoiD suggested that we just get rid of the `IProximityHandler` interface since the PR removes the only bit of code that was actually using that interface, and `NearProximity` is still usable with plugins if we remove that.

gonX:
Superseded by #2503 #2504 #2505 #2506 #2507 #2508

Some later commits in this PR were not implemented as they were (unsuccessful) attempts at fixing the weird behavior noted in the bottom of the OP, so they've been split so they can be tested individually.

PR #1994, opened 2022-01-22
Move packaging files to the main repository and add a basic Makefile

hwsmm:
This PR contains a basic Makefile that can be used for packaging and files that originally belong to OpenTabletDriver.Packaging project.

... I'm not sure if my Makefile is okay lol

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ BenchmarkDotNet.Artifacts/
 # Nix
 .envrc
 shell.nix
+
+# Linux packagaing
+!Common/Linux/bin
+
--- a/Common/Linux/bin/opentabletdriver
+++ b/Common/Linux/bin/opentabletdriver
@@ -1,0 +1,10 @@
+#!/bin/bash
+systemctl --user --quiet is-active opentabletdriver
+daemonactive=$?
+
+if [ $daemonactive != 0 ]
+then
+  systemctl --user start opentabletdriver
+fi
+
+/usr/share/OpenTabletDriver/OpenTabletDriver.UX.Gtk "$@"
--- a/Common/Linux/bin/otd
+++ b/Common/Linux/bin/otd
@@ -1,0 +1,10 @@
+#!/bin/bash
+systemctl --user --quiet is-active opentabletdriver
+daemonactive=$?
+
+if [ $daemonactive != 0 ]
+then
+  systemctl --user start opentabletdriver
+fi
+
+/usr/share/OpenTabletDriver/OpenTabletDriver.Console "$@"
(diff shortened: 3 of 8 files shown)


Sublimelime:
Thumbs down on this. Almost completely dislike it. Several points:

- Moving packaging into the main repo clutters things more than they are already, and is unnecessary.
- Having a makefile in a c# project is not standard. The major advantage of a makefile, and using make, is partial compilation file-by-file (where only changed files are recompiled) while dotnet projects are compiled per project (we can't compile parts of the daemon for example, the whole thing gets `dotnet publish`-ed). As far as I know that's just how the language is, so make only really works with langs like C++ that can be compiled file-by-file.
- Doubly stupid is having a makefile that just calls build.sh. 
- (This one is more my opinion) `make install` is bad form for distributing software in general - it puts things in various places and you're reliant on `make uninstall` working to clean it from your system, lest you have to track down a bunch of stuff. It also makes things hard to update, which if you're compiling from `master` you are likely to do and often. Only packages should ever be installed, and the package manager should know about it. Everything else should just run portable once compiled, which OTD is capable of, after the udev rules and modprobe stuff is done (which is explained in the readme, "if you're building from source you have to install the udev rules....").

hwsmm:
The reason I made a Makefile for OTD is not mainly for letting users install OTD with Makefile, but your points makes more sense than I am trying to do with this PR, I think.

Main purpose of this PR is to make packaging easier and fit to the standard of packaging in OpenTabletDriver.Packaging.
As @korewaChino said in discord chat, current way of packaging OTD is really fragile and unexpectable as there is no source package generated. Current OTD packaging just builds projects from source and puts them directly to the package folder. This is also never a good practice.

I just tried to fix this situation with this PR and https://github.com/OpenTabletDriver/OpenTabletDriver.Packaging/pull/17 .
A debian source package is consisted of an original source tarball, a debian tarball, and a control file, so the packaging files need to be in the original source tarball to make debian packaging possible. Also, dpkg detects Makefile automatically, so it makes packaging much easier with it (as you see in debian/rules packaging#17, there is nothing particularly written. dpkg just handles everything by itself), and also in RPM, `%make_build` and `%make_install` are enough to build if there is a Makefile.

We also can have a normal source package from now on with it, too, but it's also true that it's a dotnet program that relys on NuGet, which is a service on the web, so the reproducibility of the package is still low.. It's just something we can't solve for now.
I didn't mean to make users use Makefile directly at all. At least, my sole purpose of this is for packaging. and linux users should know using `make` directly can cause problems at any time.

Uh... and for `./build.sh`, I don't know much about dotnet build system so I wasn't sure I could run it in parallel.

Anyway... It seems many people are against of this, so I am closing this PR.

PR #2030, opened 2022-01-30
build: add R2R

imba-tjd:
This option can decrease startup time but increase build time.

Note I didn't test it.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -11,7 +11,8 @@ shift
 config=(--configuration='Release')
 
 options=(${config} --framework='net6.0' --self-contained='false' --output='./bin' \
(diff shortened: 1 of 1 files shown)


vedattt:
As you say, this option makes builds take significantly longer (from 20 to 70 seconds) and probably increases binary sizes as well. 

The following may also apply:
> it also reduces the chances that JIT may do its job of producing optimized instructions, since R2R uses AOT to generate generic compiled code but "deoptimization" is still a field where .NET is lacking so it can't go back to attempt to reoptimize to non-generic architecture

With these considered, this option may not be a good idea for OTD. 
I've not actually noticed any need to optimize startup times either, frankly. It's always been pretty quick on my end, but perhaps it might make sense to look further into it if it's been slow for you.

imba-tjd:
🆗

PR #2048, opened 2022-02-06
Prevent infinite exceptions from UX when daemon disconnected

### LINKED ISSUE #2028: Infinite Message boxes are shown when daemon is crashed.
## Description

Log.Exception is called when unhandled exception happens.
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/502589660232add52a1ae7bd7de584e7192f2ec2/OpenTabletDriver.UX/App.cs#L128

The output handler of Log is async function without try-catch.
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/34f27af7e1b14591fc7832f1527699e93d83454d/OpenTabletDriver.UX/MainForm.cs#L329

Output handler is invoked without await, so exception happens inside WriteMessage cant be catched.
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/caf21a15b3130fe85e02255099e126b36e154406/OpenTabletDriver.Plugin/Log.cs#L19

ShowUnhandledException is invoked again to handle unhandled exception from WriteMessage.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | MacOS
| OpenTabletDriver Version | 0.6.0.2

vedattt:
## Changes

- Check if RPC client is connected to the daemon before trying to send it a log message to print.
It makes sense to check only the daemon output handler rather than the `ShowUnhandledException` method as `Log.Output` may have sensible handlers that should run even if daemon is dead.

## Issues

- Closes #2028

### DIFF
--- a/OpenTabletDriver.Desktop/RPC/RpcClient.cs
+++ b/OpenTabletDriver.Desktop/RPC/RpcClient.cs
@@ -14,6 +14,7 @@ namespace OpenTabletDriver.Desktop.RPC
         private IList<Action<T>> reconnectHooks = new List<Action<T>>();
 
         public T Instance { private set; get; }
+        public bool IsConnected { get => rpc != null && !rpc.IsDisposed; }
 
         public event EventHandler Connected;
         public event EventHandler Disconnected;
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -326,7 +326,7 @@ namespace OpenTabletDriver.UX
         private void HandleDaemonConnected(object sender, EventArgs e) => Application.Instance.AsyncInvoke(async () =>
         {
             // Hook events after the instance is (re)instantiated
-            Log.Output += async (sender, message) => await Driver.Instance.WriteMessage(message);
+            Log.Output += async (sender, message) => { if (Driver.IsConnected) await Driver.Instance?.WriteMessage(message); };
             Driver.TabletsChanged += (sender, tablet) => SetTitle(tablet);
 
             // Load the application information from the daemon


AkiSakurai:
Null checking is not useful.
This is the exception it throw.
```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'StreamJsonRpc.JsonRpc'.
   at Microsoft.Verify.NotDisposed(IDisposableObservable disposedValue, String message)
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult](RequestId id, String targetName, IReadOnlyList`1 arguments, IReadOnlyList`1 positionalArgumentDeclaredTypes, IReadOnlyDictionary`2 namedArgumentDeclaredTypes, CancellationToken cancellationToken, Boolean isParameterObject)
   at OpenTabletDriver.UX.MainForm.<>c.<<HandleDaemonConnected>b__11_1>d.MoveNext() in /Users/shawn/Desktop/Study/otd/OpenTabletDriver/OpenTabletDriver.UX/MainForm.cs:line 333
  ```

vedattt:
@AkiSakurai Trying an exception-based approach. Could you try reproducing with this build?

AkiSakurai (commented):
The exception is not catchable here.

https://stackoverflow.com/questions/53215577/why-does-exception-from-async-void-crash-the-app-but-from-async-task-is-swallowe

vedattt:
> The exception is not catchable here.

@AkiSakurai Catching the exception left a bit of a sour taste anyway, I ended up exposing the internal RPC connection status. Have a look.

AkiSakurai:
It works.

vedattt:
Nice! Thanks for confirming.

InfinityGhost (approved):
lg2m

PR #2049, opened 2022-02-06
CI: Move format project argument before file list

vedattt:
I messed up in #2017. I hadn't tested it locally with the `include` option, which actually eats up the project argument right after. Moving it earlier fixes it locally, this should fix it for good.


InfinityGhost (approved):
Bruh

PR #2086, opened 2022-02-20
Log inner exceptions

### LINKED ISSUE #2080: Log InnerException(s) too
## Description

An user of my plugin had an issue (it was me being an idiot and not packaging a required library, but the logs (see image) didn't store the [InnerException](https://docs.microsoft.com/en-us/dotnet/api/system.exception.innerexception?view=net-6.0), which made the debugging of it seem much more complicated than it actually was.
Logs:
image

When loading up the project and debugging the daemon:
image

So, yeah, it would be very nice to log [InnerException](https://docs.microsoft.com/en-us/dotnet/api/system.exception.innerexception?view=net-6.0)s too when present.

To the person who had the issue: i made a [fix](https://github.com/OpenTabletDriver/Plugin-Repository/pull/39).

adryzz:
*Fine,  i'll do it myself.*

Fixes #2080 

- [x] Log inner exceptions
- [ ] Don't crash when there is one
- [ ] Roll back settings when the current ones are invalid

### DIFF
--- a/OpenTabletDriver.Plugin/Log.cs
+++ b/OpenTabletDriver.Plugin/Log.cs
@@ -85,9 +85,11 @@ namespace OpenTabletDriver.Plugin
         {
             if (ex == null)
                 return;
-
             var message = new LogMessage(ex);
             Write(message);
+
+            if (ex.InnerException != null)
+                Exception(ex.InnerException);
         }
     }
 }
--- a/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
+++ b/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
@@ -3,6 +3,7 @@ using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
 using Newtonsoft.Json;
+using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Attributes;
 using OpenTabletDriver.Plugin.DependencyInjection;
 using OpenTabletDriver.Plugin.Tablet;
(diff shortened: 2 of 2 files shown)


OpenTabletDriver.Plugin/Log.cs
```diff
@@ -85,9 +85,11 @@ public static void Exception(Exception ex)
         {
             if (ex == null)
                 return;
-
             var message = new LogMessage(ex);
             Write(message);
+
+            if (ex.InnerException != null)
+                Exception(ex.InnerException);
```
InfinityGhost, line 92: Is there a potential for infinite recursion here?

vedattt: The dotnet library [`Exception.GetBaseException`](https://github.com/dotnet/runtime/blob/68fb7fc68cc1af800bee1d38af22b5027bf4ab4e/src/libraries/System.Private.CoreLib/src/System/Exception.cs#L62) method is doing the same iteration over `InnerException`s too, so this should be fine.
```cs
while (inner != null)
    inner = inner.InnerException;
```

adryzz: no, there isn't because you create an exception with inner exception via constructor. so you'd have to call the constructor recursively, which doesn't work

adryzz:
well i kinda forgot about this... i wanted to do more stuff but this works just as well

gonX:
If this change is still relevant, it needs to be retargeted for the `0.6.x` or the `avalonia` branch.

gonX:
Closing as #4408 supersedes this for `0.6.x`, and `master` is no longer taking code contributions.

PR #2093, opened 2022-02-23
Add Intuos4 middle button to AuxButtons

Hypfer:
During testing of a newly acquired PTK-1240, I've noticed that the button in the middle of the touch wheel wasn't handled.

Because the button was added in the middle of the array, this PR **will break** existing bindings for the four buttons at the bottom.


jamesbt365:
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/CONTRIBUTING.md

Hypfer:
What did I miss?

jamesbt365:
rebase to configs branch.

Hypfer:
But it's not a config change 🤔

Oh well. Will do

X9VoiD (changes requested):
Use `IsBitSet()`, and no need to document things already parsed here, the code itself *is* the documentation in really trivial cases like this.

PR #2117, opened 2022-03-12
Add identifier to XP-PEN Deco mini7

jamesbt365:
Verification: https://canary.discord.com/channels/615607687467761684/615611007951306863/952017593772961972

This pull request adds another identifier, this one adds support for wireless.

Could all the identifiers be merged into 12, null instead of listing each one like 12 12, 12 10, 12 21 and 12 32?

Also corrected incorrect TABLETS.md entry listing missing features without saying what, however the tablet is fully supported as shown above in the verification.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco mini7.json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco mini7.json
@@ -58,6 +58,19 @@
       ],
       "DeviceStrings": {},
       "InitializationStrings": []
+    },
+    {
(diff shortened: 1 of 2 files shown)


X9VoiD:
I'll accept the PR with `null` output report length, if there will be a PR that'll make sure that configs with non-zero length output report will only match identifiers that satisfies the required length (to avoid dumb situations where OTD mistakenly uses a 12:0 or similar endpoints)

X9VoiD:
Merging, but will open another issue to track non-zero output report match on `null` length config.

PR #2143, opened 2022-03-31
Add support for XENX tablets

X9VoiD:
- [x] XENX X1-640
- [x] XENX P1-640
- [x] XENX P3-1060

## Verification

https://discord.com/channels/615607687467761684/958666363667365928/961183975068549170


InfinityGhost (approved):
lg2m

Kuuuube (changes requested):
Tested and working on XENX P1-640 and XENX X1-640. The button count for the P1-640 needs to be changed to 6 though.

Also for future reference if there is a way to add this properly later: The feature init to activate "osu mode" is `IAM=`. It can't be used in the default config though since it removes all pressure except 0 and 1.

Issue #2146, opened 2022-03-31
Diagnostics export should not be possible without a file extension


gonX:
## Description
Not having a file extension (preferably `json` or `txt`) messes with Discord embeds, making it harder for Discord helpers.

We should probably check for a file extension, and if there isn't any, we append `.json` to it

[OpenTabletDriver.UX/MainForm.cs:587](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver.UX/MainForm.cs#L587)

PR #2150, opened 2022-04-01
Restrict file extension

satodayo:
It is my first contribution.
Please let me know if this PR is not good.
Thank you.

### DIFF
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -603,7 +603,8 @@ namespace OpenTabletDriver.UX
                 {
                     case DialogResult.Ok:
                     case DialogResult.Yes:
-                        var file = new FileInfo(fileDialog.FileName);
+                        string[] options = { ".json", ".txt", ".log" };
+                        var file = new FileInfo(fileDialog.FileName + (options.Any(fileDialog.FileName.EndsWith) ? "" : ".json"));
                         if (file.Exists)
                             file.Delete();
                         using (var fs = file.OpenWrite())


satodayo:
for #2146

jamesbt365:
If a file extension that isn't provided nothing saves, however .txt and .json do save as intended. Tested on linux.

satodayo:
Thank you for feedback.
I misunderstand, I will fix it.

satodayo:
Resolve #2146

vedattt:
Regular expressions are way way overkill for the task here. Since file extensions, if they exist, are always at the end of `fileDialog.FileName`, opt for a simpler algorithm using [`String.EndsWidth`](https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-6.0). See [here](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/1798/files) for a similar PR from earlier.

Also, if Discord is able to display JSON files as well as TXT files, I'd say forget about checking against `.txt` altogether since the diagnostics being exported are in fact JSON, for the time being.

vedattt:
Apparently Discord plays well, but GitHub issues don't let you upload JSONs. In that case, I think we should enforce `.json` by default, but also allow saving as either `.log` or `.txt`, both of which work fine with Discord and GitHub.

gonX:
> Regular expressions are way way overkill for the task here. Since file extensions, if they exist, are always at the end of fileDialog.FileName, opt for a simpler algorithm using [String.EndsWidth](https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-6.0). See [here](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/1798/files) for a similar PR from earlier.

I agree with this. For readability this could even be done with Linq:
```cs
String fileName = "blabla.txt";
String[] options = { ".json", ".txt", ".log" };
bool result = options.Any(fileName.EndsWith);
```

gonX (changes requested):
Change from Regex to [String.EndsWith](https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-6.0)

satodayo:
@vedattt @gonX 
Thank you for your feedback!
I guess regex is overkill as well.

Please let me check a little.
If file extension is not any, I guess I can't detect with only string.EndsWitdh.
So I guess I need to handle that filename case.
If then, should I add ".json" for filename?  or save as it is ? or unsave?

OpenTabletDriver.UX/MainForm.cs
```diff
@@ -603,7 +604,19 @@ private async Task ExportDiagnostics()
...
+                        switch (match.Groups["extension"].Value) {
+                            case ".json":
+                            case ".txt":
+                            case "":
+                                filename = fileDialog.FileName;
+                                break;
+                            default:
+                                return;
+                        };
+                        var file = new FileInfo(filename);
```
satodayo, line 619: Currently
```suggestion
                        String fileName = fileDialog.FileName;
                        String[] options = { ".json", ".txt", ".log" };
                        if (!options.Any(fileName.EndsWith))
                            return;

                        var file = new FileInfo(fileName);
```

vedattt:
If the given filename ends with json/log/txt, it is saved as it is. In any other case, the file is saved as `filename + ".json"`.

So you could just grab gonX's example using `Any(endsWith)` to test whether you save it is as is, or save it with `.json` appended. That should cover all cases.

You could probably reduce the code down to a single line or two as well, going this way.

satodayo:
That make sense.
I was thinking too compricated, thank you.
I will fix it, please give me some time...

satodayo:
Because I pushed a lot of unnecessary commit, I will re-contribute with new branch.
I am sorry for my many mistake.

jamesbt365:
you don't have to do that, just squash commits.

satodayo:
@jamesbt365 
Thank you so much.
I didn't know anything..., I squashed commits.
I am sorry for the trouble.

jamesbt365:
no worries, will test and review in a bit

jamesbt365 (approved):

jamesbt365:
Works correctly.

OpenTabletDriver.UX/MainForm.cs
```diff
@@ -603,7 +603,8 @@ private async Task ExportDiagnostics()
                 {
                     case DialogResult.Ok:
                     case DialogResult.Yes:
-                        var file = new FileInfo(fileDialog.FileName);
+                        String[] options = { ".json", ".txt", ".log" };
```
vedattt, line 606:
```suggestion
                        string[] options = { ".json", ".txt", ".log" };
```
Sorry, our example snippet made this "mistake" as well, but C# code style tends to prefer using the alias `string` over the class `String`. They both compile down to the same thing, so it's just cosmetic consistency. You can fix this and squash it once again. Looks good to merge otherwise.

vedattt (changes requested):
Super minor thing left

OpenTabletDriver.UX/MainForm.cs
vedattt: P.S., you could go a step further by making it a `const` variable at the top of the method.

satodayo: @vedattt 
Thank you for review.
I fixed it.

gonX (approved):

jamesbt365 (approved):

InfinityGhost (approved):

PR #2175, opened 2022-04-10
Correctly parse hover distance for intuos4

Kert:
Correctly parse hover distance for intuos4
Right now it caps at "62"


Kuuuube:
If we're going to do this might as well remove the ` >> 2` for hover everywhere. So here as well: 

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver.Configurations/Parsers/Wacom/Intuos3/Intuos3MouseReport.cs

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver.Configurations/Parsers/Wacom/Intuos/IntuosTabletReport.cs

Also, tablet parser changes should go to the `configs` branch.

gonX:
#1979 might be relevant to note here

PR #2176, opened 2022-04-10
Added full support for using conf.toml (logLevel)

Vanadiom:
I added full .toml support, could be improved at places, but works rather well for now.
Changes to the .toml (i think) only come into effect once the Daemon restarts.


Vanadiom:
The UX spits errors, I'll put it on hold

PR #2177, opened 2022-04-10
Add LogLevel configuration

X9VoiD:
Default LogLevel is set to `Info`.

The configuration file is loaded only during boot sequence of daemon, and reconnection of UX (why do we still have two separate instances of everything in `OpenTabletDriver.Desktop/Plugin`).

Generated file looks like this:
```toml
# The minimum level of log messages to output. Valid values are Debug, Info, Warning, Error, Fatal
loglevel = "Info"
```

This supersedes #2176.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -34,6 +34,7 @@ namespace OpenTabletDriver.Daemon
         {
             Driver = driver;
 
+            Configuration.Load(AppInfo.Current.ConfigurationFile);
             Log.Output += (sender, message) =>
             {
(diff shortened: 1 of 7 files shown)


InfinityGhost:
Why is this using TOML? This could easily be done in the Settings class.

X9VoiD:
Things are already logged before settings is loaded.

gonX:
I'm not sure if I like the TOML idea either. How about using env vars?

gonX:
@X9VoiD respond pls

X9VoiD:
Env vars is fine, will update at some point.

gonX:
Converting your PR to draft given the above message

PR #2190, opened 2022-04-15
Restore linux native timers

jamesbt365:
With the merging of https://github.com/dotnet/runtime/pull/65406 and dotnet 6.0.4 releasing, the memory leak with native timers has been fixed, and now we can use them again.

Is this the line that was removed, and would this work just throwing it back in now lol?

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
+++ b/OpenTabletDriver.Desktop/Interop/DesktopInterop.cs
@@ -72,6 +72,7 @@ namespace OpenTabletDriver.Desktop.Interop
         public static ITimer Timer => CurrentPlatform switch
         {
             PluginPlatform.Windows => new WindowsTimer(),
+            PluginPlatform.Linux   => new LinuxTimer(),
             _ => new FallbackTimer()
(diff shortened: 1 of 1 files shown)


gonX:
Would it make sense to check for .NET version, or do we just assume it works for everyone now?

doubledge-spec:
I think it's safe to assume it works for everyone now. Since other major Linux distros just follow microsoft's release cycle. The only problem is, is that Arch repos are still on 6.0.2. Is it possible for the time being to make the aur -bin packages for dotnet 6 a dependency?

gonX:
I'll just not update the main AUR package until repos have 6.0.4

I'm inclined to say git users are on their own, but requiring `dotnet-runtime>=6.0.4` is an option too. I'll think about it.

InfinityGhost:
> I'm inclined to say git users are on their own, but requiring `dotnet-runtime>=6.0.4` is an option too. I'll think about it.

That's what I'd suggest to do when you decide to update the AUR repo.

InfinityGhost (approved):

PR #2195, opened 2022-04-17
Fix supported status of Huion New 1060 Plus

jamesbt365:
supported status says its missing tilt, tablet doesn't support tilt.


gonX (approved):

gonX:
For anyone else wondering: https://www.huion.com/pen_tablet/Huion/NEW_1060PLUS.html

PR #2205, opened 2022-04-19
Fix wheel support for Gaomon M6

X9VoiD:


InfinityGhost (approved):
lg2m

nukawa:
Sorry to bother you, I tried to adopt OpenTableDriver for my Gaomon M6 first time.
That is a good experience for me so far. However, these are too difficult to me due to a brand new Linux user, could you help to teach me or tell me the related information to adopt these materials as above? Thanks a lot.

X9VoiD:
OTD doesn't support configuring and binding to wheels yet, only parsing.

nukawa:
Thanks for your reply, I will keep waiting the following update, thanks a lot.

PR #2213, opened 2022-04-24
Add RPC connection timeout

### LINKED ISSUE #2208: A timeout message should be displayed when the daemon fails to connect.
## Description

As of right now if the UX fails to connect to the daemon it keeps trying with no real indicator of if its working or not. There should be a timeout message to indicate to the user that it has failed.

InfinityGhost:
Fixes #2208


jamesbt365:
works as intended.
image

jamesbt365 (approved):

PR #2222, opened 2022-05-03
AreaDisplay: Display lines on intersecting areas

gonX:
This is useful for multi-monitor configurations as it allows users to more distinctly see whether they're on the middle of 2 areas or not.

It is potentially very slow code (would be appropriate to say this is going from O(n) to O(n^2)?), so as an alternative we can just change the existing `AreaBoundsBorderColor` to something more eye catching.

Alternatively I have ideas to optimize the code if we're interested in this behavior, but I just wanted to showcase the idea before sinking too much time into optimization.

image


InfinityGhost:
Requires a rebase to work with the DI rewrite. This may not be necessary as this portion of the UI was rewritten with different styling.

gonX:
Rebased and then I realized what you meant by different styling - it does indeed look acceptable on master without this PR. If anything the border color can be changed but that's a different discussion.

PR #2236, opened 2022-05-10
Improve sleep detection routine

X9VoiD:
## Changes

- Raise from 2 secs to 8 secs interval before triggering wake up events.
- Refactor for easier injection of other implementations (sleep detection is now testable).
- Compare against system clock.

Some platforms (Linux) slide `Stopwatch` to compensate suspend time and make it look like it never happened.


X9VoiD:
Rebased.

OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -106,15 +107,15 @@ public async Task Initialize()
...
-            SleepDetection = new SleepDetectionThread(() =>
+            _sleepDetector.Slept += async () =>
             {
-                Log.Write(nameof(SleepDetectionThread), "Sleep detected...", LogLevel.Debug);
-                _ = DetectTablets();
-            });
-
-            SleepDetection.Start();
+#if DEBUG
+                if (System.Diagnostics.Debugger.IsAttached)
```
gonX, line 113: I would prefer this has a comment why. It's probably intuitive to us why the sleep routine will fail when debugging, but it also makes debugging of the sleep routine harder. I suggest adding said comment to `SleepDetector.cs`, however.

OpenTabletDriver.Desktop/Components/SleepDetector.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        private void Stop()
+        {
+            lock (sync)
+            {
+                if (task == null)
+                    return;
+
+                cts!.Cancel();
+                task.Wait();
```
gonX, line 65: check warnings

OpenTabletDriver.Daemon/DriverDaemon.cs
X9VoiD: When you hit a breakpoint during debugging it'll trigger the sleep detection routine if not taken care of. It's a lot harder to debug other things with this component active during debugging. I added the if so a person can just remove the if he so wants to debug sleep stuff.

gonX: Works for me

PR #2237, opened 2022-05-10
Add F13-F24

jamesbt365:
le troll macos


InfinityGhost (approved):

sealvessel:
Actually a reasonable request. Got f13-f15 keys here that could use some binding for some system functions like volume, etc.

X9VoiD:
The changes here are already merged. It's not a feature request.

PR #2243, opened 2022-05-11
Remove Octokit dependency on updater interfaces

X9VoiD:
And refactor out Github networking code to separate class.

This allows for mocking the version returned by update checks, and avoid issues when other branches became out-of-sync, causing false-positive test failures.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -5,7 +5,6 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Diagnostics;
-using Octokit;
 using OpenTabletDriver.Desktop;
 using OpenTabletDriver.Desktop.Binding;
 using OpenTabletDriver.Desktop.Contracts;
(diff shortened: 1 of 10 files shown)


OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -463,9 +462,9 @@ public Task<bool> HasUpdate()
             return Updater?.CheckForUpdates() ?? Task.FromResult(false);
         }
 
-        public async Task<Release> GetUpdateInfo()
+        public async Task<UpdateInfo?> GetUpdateInfo()
         {
-            return await Updater.GetRelease()!;
+            return await Updater.GetInfo();
```
vedattt, line 467: There doesn't seem to be a reason for `return await` here?

OpenTabletDriver.Desktop/Updater/Updater.cs
```diff
@@ -67,26 +66,24 @@ public async Task<bool> CheckForUpdates(bool forced)
...
         {
-            if (latestRelease == null)
-            {
-                await CheckForUpdates();
-            }
+            if (updateSentinel == 2)
+                return null;
 
-            return latestRelease;
+            return await GetUpdate();
```
vedattt, line 74: There doesn't seem to be a reason for `return await` here?

vedattt (commented):
Just minor things. IMO `DriverDaemon` could benefit from refactoring some of these simple methods to be expression-bodied, but maybe that's another PR.

InfinityGhost (approved):
lg2m

Issue #1367, opened 2021-08-22
Scrollwheel support 


C0D3-M4513R:
## Description
The MTE-450 has a scroll wheel. I would like, for some kind of implementation, that lets the scroll wheel be bound, to e.g. Mouse wheel? or windows wheel.

caseyjp11:
Adding in my request for the HS610's wheel.

gelatinbomb:
I'm curious about something. Can't we have the wheel function as two auxiliary buttons so you can choose what key is gonna be pressed when the wheel turns up or down?

I did a test using evtest and the wheel is detected as buttons so If I rotate the wheel up or down evtest detects it as a quick push and release of the chosen key. Not only that, I tested all the XP Pen devices on my system and the one that holds the wheel is in fact the XP Pen Virtual Keyboard.

I don't know how the other tablets would handle the behavior of the wheel but maybe as a temporal or easy solution the wheel coould be understanded on the Opentabletdriver as a set of auxiliary buttons you can bind.

jamesbt365:
Having the wheel function as 2 auxiliary buttons is not proper support but could be done. Proper support should be within the next version for some tablets.

andreluizrcn:
Hello! hope this message reaches you well, I would like to know if the Veikk A15 pro have some kind of a future wheel support !!! thanks :)

iamalsonothere:
Adding in my request for the XP-PEN Deco 03's wheel.

jamesbt365:
FYI asking for specific tablets wheels to be supported when you haven't provided data for how that tablets wheel functions is futile because there isn't actually anything to go off for supporting said wheel.

In the future unless you are going to provide a data recordings for all the wheels functions that clearly describe what is going on, not just randomly dumping a data recording then ghosting us, please don't ask for specific tablets wheels, they will all eventually be supported when OTD supports wheels.

iamalsonothere:
Sorry, I was under the impression that it was somewhat standardized

Heres a recording of my tablets wheel being turned one way for half the file and the opposite way in the other half

[tablet-data.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/15139177/tablet-data.txt)

Nukley:
Attached is a file containing a relevant snip of data, along with a breakdown on what it means. 

Attached below that is a gif of me showing a really bad demo program on how these values work.

[...]

PR #2256, opened 2022-05-19
Add absolute wheel (ring) support

gonX:
Partial fix for #1367 - code needs a bit of cleanup (magic numbers into vars) and we likely need some UX configurables.

The idea is absolute wheels should just work:tm: without having to set anything, but UX additions are welcome of course. You can PR to my branch of course.

As the code is right now, it's not doing what I want.
I _think_ libinput is handling the events differently because the OTD device is considered a pen and not a pad. We might have to split events into 2 devices here, one for the pen itself and one for the pad if fixing libinput tags (via quirks?) doesn't fix it

Leaving as draft as I'm not entirely satisfied with it yet, but people can take a look and tell me if there's anything glaringly wrong. I have no plans of personally including relative wheel support, though once again you can PR to my branch.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
@@ -17,7 +17,8 @@
       "ButtonCount": 8
     },
     "MouseButtons": null,
-    "Touch": null
+    "Touch": null,
+    "MaxWheelPosition": 71
   },
   "DigitizerIdentifiers": [
     {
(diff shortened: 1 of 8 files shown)


OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
```diff
@@ -17,7 +17,8 @@
       "ButtonCount": 8
     },
     "MouseButtons": null,
-    "Touch": null
+    "Touch": null,
+    "MaxWheelPosition": 71
```
InfinityGhost, line 21: Would prefer another specification type for this instead, so that this could be expanded on.

Issue #1665, opened 2021-11-26
Touch report of other tablets interfere with each other


X9VoiD:
## Description

Some touch parsers contain state, and these states are static. This causes issues when more than one tablet uses the same type of report parser.

### Affected Types
- [WacomTouchReport](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver/Vendors/Wacom/WacomTouchReport.cs) https://github.com/OpenTabletDriver/OpenTabletDriver/blob/5ff9f4dd5640f1c4d16190d5100d1f290abee002/OpenTabletDriver/Vendors/Wacom/WacomTouchReport.cs#L73

- [IntuosV2TouchReport](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver/Vendors/Wacom/IntuosV2/IntuosV2TouchReport.cs) https://github.com/OpenTabletDriver/OpenTabletDriver/blob/5ff9f4dd5640f1c4d16190d5100d1f290abee002/OpenTabletDriver/Vendors/Wacom/IntuosV2/IntuosV2TouchReport.cs#L42

PR #2260, opened 2022-05-22
Add eraser support for intuos1

Kuuuube:


OpenTabletDriver.Configurations/Parsers/Wacom/Intuos1/Intuos1ReportParser.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                ToolStore = report[3];
+                return new Intuos1ToolReport(report);
+            }
+                
+
+            return new DeviceReport(report);
+        }
+
+        public static byte ToolStore;
```
AbstractQbit, line 36: Using statics in report parsers can potentially mess with multi tablet support, see #1665. Just make it a regular member and pass it to report constructor instead

AbstractQbit (changes requested):

jamesbt365 (approved):

gonX:
What happens to the V1 parser from here?

OpenTabletDriver.Configurations/Parsers/Wacom/Intuos1/Intuos1ToolReport.cs
```diff
@@ -0,0 +1,35 @@
+using OpenTabletDriver.Plugin.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Wacom.Intuos1
+{
+    public struct Intuos1ToolReport : IToolReport, IEraserReport, IProximityReport
```
X9VoiD, line 5: Consider removing `IEraserReport` here and `Eraser` so eraser event and position is parsed at the same time frame? This doesn't really matter much so feel free to ignore.

X9VoiD (changes requested):
Is there no more users of IntuosV1? Should delete it if there's none since Intuos1 is directly compatible with V1 (i hate the parser name scheme we have if there even is one lol).

OpenTabletDriver.Configurations/Parsers/Wacom/Intuos1/Intuos1ToolReport.cs
gonX: It's probably a good idea to address this, as `IToolReport`s are not yet implemented and could potentially end up being reported differently from `IEraserReport`s (as `IEraserReport`s are usually a part of a position report, not a tool report).

Should that not be the case then the change will do nothing on that front, but still implies intent better than what the PR proposes.

Kuuuube: It is not possible to parse eraser and position at the same time. They are never sent together.

The tablet sends one report containing the tool data only when the tool changes or goes from out of range to in range. The tool data contained in the tool report is valid for all reports after the tool report, until another tool report is received.

It may seem worth actually using `ToolType.Eraser` or `ToolType.Pen` for this but I don't want to do that since this should be arbitrarily expandable and allow checking for more than eraser or pen tip. Though not supported in OTD yet, this tablet also supports mice and some other pens which may need special handling. The tool reports for those are reported over the same byte.

Kuuuube:
> Is there no more users of IntuosV1? Should delete it if there's none since Intuos1 is directly compatible with V1 (i hate the parser name scheme we have if there even is one lol).

No there are still many tablets that use it. Some Wacom tablets should be split off similar to this PR but even then there's some other tablets that use it.

X9VoiD:
Superseded by #2260

PR #2270, opened 2022-05-26
Add arm64 and x86 building on github actions

jamesbt365:
Adds arm64 builds to github actions.

add for releases would probably break the updater, and the updater doesn't know the difference between arm, x86_64 and x86 anyway so that needs changing before releases can be done

Packaging repo also needs doing :P

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -20,7 +20,7 @@ jobs:
           dotnet-version: '6.0'
           include-prerelease: True
 
-      - name: Linux Publish
+      - name: Linux x86_64 Publish
         env:
           runtime: linux-x64
         run: |
@@ -72,7 +72,7 @@ jobs:
           dotnet-version: '6.0'
           include-prerelease: True
 
-      - name: MacOS Publish
+      - name: MacOS x86_64 Publish
         env:
           runtime: osx-x64
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnet.yml
```diff
@@ -100,6 +130,21 @@ jobs:
           name: OpenTabletDriver win-x64
           path: build/win-x64
 
+      - name: Windows arm Publish
```
InfinityGhost, line 133:
```suggestion
      - name: Windows ARM64 Publish
```

InfinityGhost (changes requested):
Since we're targeting multiple platforms now, we should include the architectures in the workflow names as well, so update the old ones to include the architectures.

jamesbt365:
> Since we're targeting multiple platforms now, we should include the architectures in the workflow names as well, so update the old ones to include the architectures.

Should the path it builds to also be changed to x86_64 from x64 if we are going this route?

jamesbt365:
Eto likely needs to be bumped for native arm builds to work on macos. #2336

jamesbt365:
Probably best to rip out macos arm for now because it requires something we cannot supply rn and has to be built and signed on macos

jamesbt365:
Lots has changed since this pull request was opened. Not sure if we really "need" x86 builds anymore due to dotnet 7 phasing out Windows 7 and 8 support. Handling arm MacOS builds requires signing on MacOS for it to work so in reality it seems only really arm windows (which hardly anybody uses) and arm linux exist to implement if we are still interested in doing something like that.

PR #2247, opened 2022-05-13
Add CTH-661-SE config

TheOnlyPhoenix:

PR #2271, opened 2022-05-27
Add identifier for CTH-661

jamesbt365:
Verification: https://discord.com/channels/615607687467761684/615611007951306863/979856404196241412

Weird problem with the debugger causing it to crash when opened so the position value couldn't be verified, but due to the fact that corners can be reached on the screen and the user has said that changing the area and trying it still keeps accuracy that they are the same.

<sup><sup>also merge #2247 first i don't think phoenix has the brain to figure out merge conflicts in the slightest</sup></sup>


X9VoiD:
Resolve conflict

jamesbt365:
Resolved.

PR #2272, opened 2022-05-28
Add native notifications when the fallback timer is used on Windows.

jamesbt365:
Is this how you do it :voidvoid:

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Timer/WindowsTimer.cs
+++ b/OpenTabletDriver.Desktop/Interop/Timer/WindowsTimer.cs
@@ -43,7 +43,7 @@ namespace OpenTabletDriver.Desktop.Interop.Timer
                     }
                     else
                     {
-                        Log.WriteNotify("Timer", "Unsupported interval detected, will use fallback timer. Expect high CPU usage", LogLevel.Warning);
+                        Log.WriteNotify("Timer", "Unsupported interval detected, will use fallback timer. Expect high CPU usage. Please use 1000hz, 500hz, 250hz or 125hz instead.", LogLevel.Warning);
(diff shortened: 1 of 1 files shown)


jamesbt365:
image

vedattt:
> won't go through if the user has notifications from otd off

Does the notification get logged to at least the OTD console in this case?
So long as there is *some* indicator out there, I think that's enough. User is responsible beyond that point if they've disabled OTD notifications.

jamesbt365:
Its still logged, just the notification won't appear.

jamesbt365:
Rebased. No clue if it still works but i'm gonna assume it does because why wouldn't it.

OpenTabletDriver.Desktop/Interop/Timer/WindowsTimer.cs
```diff
@@ -43,7 +43,7 @@ public unsafe void Start()
                     }
                     else
                     {
-                        Log.Write("Timer", "Unsupported interval detected, will use fallback timer. Expect high CPU usage", LogLevel.Warning);
+                        Log.WriteNotify("Timer", "Unsupported interval detected, will use fallback timer. Expect high CPU usage", LogLevel.Warning);
```
InfinityGhost, line 46: What interval? This should be a bit more specific for the user to know what to change.

InfinityGhost (changes requested):

OpenTabletDriver.Desktop/Interop/Timer/WindowsTimer.cs
jamesbt365: Fixed.

PR #2286, opened 2022-06-02
Add Lifetec LT9570 support

X9VoiD:


TABLETS.md
```diff
@@ -138,6 +138,7 @@
 | Huion Kamvas Pro 20           |  Missing Features | Touch bar is not yet supported.
 | Huion Kamvas Pro 24           |  Missing Features | Touch bar is not yet supported.
 | Huion Q620M                   |  Missing Features | Wheel is not yet supported.
+| Lifetec LT9570                |  Missing Features | Aux buttons are not yet supported.
```
gonX, line 141: Assuming the parsers tilt portion is right (`// Tilt is given as azimuth and altitude`), I'm assuming this needs to report that tilt is not yet supported:
```suggestion
| Lifetec LT9570                |  Missing Features | Aux buttons and tilt is not yet supported.
```

PR #2216, opened 2022-04-25
Add Veikk A30 V2

SnowLire:
Still waiting on the guy who asked for support to come back and test it, although I suspect it's the same as the Veikk S640 V2 and only needed a parser change.

Conversation: https://discord.com/channels/615607687467761684/615611007951306863/965638577394425926

PR #2293, opened 2022-06-05
Fixed aux buttons for Veikk A30 V2

### LINKED ISSUE #1854: VEIKK A50 does not have working aux buttons
## Description
VEIKK A50 currently uses `SkipByteTabletReportParser` which does not implement aux buttons. The ReportParser used should be changed to something that does.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Any
| OpenTabletDriver Version | d758023183600
| Tablet                   | VEIKK A50

Julian-Sz:
@jamesbt365 said in this PR https://github.com/OpenTabletDriver/OpenTabletDriver/pull/2216 that the aux buttons might not be working due to the OEM drivers still being installed, but that's not the case (they are also broken on a fresh windows 10 install). 

I reverse-engineered the OEM driver and it turns out the tablet needs 3 initial reports (one for the digitizer, one for the aux buttons, one for the touchpad). I've added a parser that can parse the aux button reports. I don't know if that solution works for other Veikk tablets as well, it might fix #1854.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxReport.cs
@@ -1,0 +1,22 @@
+using OpenTabletDriver.Plugin.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Veikk
+{
+    public struct VeikkAuxReport : IAuxReport
+    {
+        public VeikkAuxReport(byte[] report)
+        {
+            Raw = report;
+            AuxButtons = new bool[]
+            {
+                report[4].IsBitSet(0) && report[3].IsBitSet(0), // First
+                report[4].IsBitSet(1) && report[3].IsBitSet(0), // Second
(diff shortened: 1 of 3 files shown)


OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxReport.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace OpenTabletDriver.Configurations.Parsers.Veikk
+{
+    public struct VeikkAuxReport : IAuxReport
+    {
+        public VeikkAuxReport(byte[] report)
+        {
+            Raw = report;
+            AuxButtons = new bool[]
+            {
+                report[4] == 1 && report[3].IsBitSet(0), // First
```
X9VoiD, line 12: `report[4].IsBitSet(0)` etc etc.

X9VoiD (changes requested):

X9VoiD (approved):
If a VEIKK tablet has more than 4 buttons then we can just extend this code.

PR #2327, opened 2022-06-25
EvdevVirtualTablet: Support forward/backwards mouse buttons

gonX:
Pre-PR:
Pen bindings using mouse button Forwards/Backwards did nothing

Post-PR:
Mouse button bindings with Forward/Backwards now does something

Notes:
I've removed the `GetCode(MouseButton button) => null` override entirely as the binding of the other 3 mouse buttons should be harmless, as they're not defined in evdev as being events that can happen from that endpoint, and in my testing they are filtered out as expected. This means that mouse{1..3} behavior is effectively not different from pre-PR.


InfinityGhost (approved):
lg2m

PR #2335, opened 2022-07-01
Adds "needs-triage" label automatically to issues and pull requests

jamesbt365:


.github/ISSUE_TEMPLATE/tablet_configuration.yml
```diff
@@ -1,7 +1,7 @@
 name: Configuration Request
 title: "Add support for "
 description: Request for a new tablet device to be supported
-labels: ["configuration"]
+labels: ["configuration", "need-triage"]
```
vedattt, line 4:
```suggestion
labels: ["configuration", "needs-triage"]
```

.github/ISSUE_TEMPLATE/support_request.yml
```diff
@@ -1,7 +1,7 @@
 name: Support Request
 title: "Support Request: "
 description: Request help for something that isn't an obvious bug
-labels: ["support request"]
+labels: ["support request", "need-triage"]
```
vedattt, line 4:
```suggestion
labels: ["support request", "needs-triage"]
```

vedattt (changes requested):

vedattt:
The only reason we've found that justifies having a `needs-triage` label over filtering by unlabeled issues is that our issue forms already auto label upon opening. In other words, there are almost* no issues that are ever completely unlabeled, hence the need for a way to distinguish.

*: Blank issues aren't labeled at all, but they're on the rare side being mostly used for non-user-facing meta topics.

jamesbt365:
All issues and pull requests opened will have the needs-triage label so that somebody can label them correctly so things are easier to sort though later

.github/labeler.yml
```diff
@@ -0,0 +1,2 @@
+needs-triage:
+  - '*'
```
vedattt, line 2:
```suggestion
needs-triage:
  - '.*'
```
`*` is not valid regex, as it attempts to quantify a token that doesn't exist. If issue-labeler somehow interprets '*' as a glob pattern then perhaps it's still functional and this could be ignored.

Otherwise, to match everything we should use either `.*` or an empty regex. Both seem to result in a match for any string in JS, which I believe is the flavor of regex that applies to GitHub Actions.

vedattt (changes requested):

.github/labeler.yml
jamesbt365: oops

InfinityGhost (approved):

PR #2359, opened 2022-07-23
Handle asleep display on macos

AkiSakurai:
CGGetActiveDisplayList does not return asleep displays.
Use CGGetOnlineDisplayList instead.


gonX:
Changing this behavior seems problematic. What if the user wakes the display after the driver has started?

X9VoiD:
I think MacOS still maps input to those sleeping monitors so OTD would have sent incorrect data if there's a sleeping monitor.

@AkiSakurai can you confirm?

AkiSakurai:
When device wakes up, otd is start before the displays wake up. Since otd can't find any active display when it is starting, it will crashes with NPE.

It happens only on 0.7 branch, as otd is now re-detecting the displays configuration when device wake up.

gonX:
Sorry, I think I read the the opposite way around. Setting milestone to next major release.

gonX (approved):
Can't test, but code makes sense to me.

InfinityGhost:
This would need updated to fix merge conflicts but lg2m

PR #2363, opened 2022-07-27
Add identifier for Deco 01 V2

nukanoto:
My Deco 01 V2 can only use the upper left 1/4. So it should use variant 2 configuration.
The difference between them seems to be device string at the index 5.

device strings:
```
1=UGTABLET
2=10 inch PenTablet
3=000000
4=UG901_BPU1002
5=2020-10-23_Release3
```

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco 01 V2 (variant 2).json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco 01 V2 (variant 2).json
@@ -34,6 +34,22 @@
         "4": "UG902_BPG1002"
       },
       "InitializationStrings": []
+    },
+    {
+      "VendorID": 10429,
+      "ProductID": 2309,
+      "InputReportLength": 12,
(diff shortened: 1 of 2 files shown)


jamesbt365:
I'm not happy with the way that you have done that. Export a diagnostic and send it here as well as a string dump.

nukanoto:
[deco01v2.zip](https://github.com/OpenTabletDriver/OpenTabletDriver/files/9197392/deco01v2.zip)

jamesbt365:
This just seems like a 200lpi having no real difference to the 100lpi variant and i don't know if this would break other variants but i'll leave that to void.

nukanoto:
Sorry, the diagnostics.json I just sent is the wrong one. Here is the correct one.
[diagnostics.json.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/9197470/diagnostics.json.txt)

jamesbt365:
You rebased to the wrong branch.

PR #2366, opened 2022-07-28
Add IProximityReport to BambooTabletReport

Kuuuube:
The tablets using BambooReportParser contain data for NearProximity so that should be parsed.

They do not contain normal precise hover distance data but it is required for IProximityReport to use HoverDistance. Using byte 1 for this isn't perfect but the data from it is reasonable to use in place of more precise hover distance data.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Wacom/Bamboo/BambooTabletReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Wacom/Bamboo/BambooTabletReport.cs
@@ -34,7 +34,7 @@ namespace OpenTabletDriver.Configurations.Parsers.Wacom.Bamboo
             };
 
             NearProximity = report[1].IsBitSet(7);
-            HoverDistance = (uint)report[1];
+            HoverDistance = (uint)report[1] >> 4;
         }
 
         public byte[] Raw { set; get; }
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Configurations/Parsers/Wacom/Bamboo/BambooTabletReport.cs
```diff
@@ -32,6 +32,10 @@ public BambooTabletReport(byte[] report)
                 report[7].IsBitSet(5),
                 report[7].IsBitSet(6),
             };
+
+            NearProximity = report[1].IsBitSet(7);
+            HoverDistance = (uint)report[1];
+
```
gonX, line 38:
```suggestion
```
This will put styling in line with other parsers

gonX (approved):
Looks OK, absolutely miniature suggestion included

Kuuuube:
Interesting... My MTE-450 does not send D0 or 50 in any case.

Kuuuube:
> In my MTE-450 (which should fall under this parser), the second byte has the following meanings (from experimental testing #656). When coming closer with a pen the second byte goes from 50 -> D0 -> 90. in L. 37 this would suggest, that D0 is closer than 90. That is not the case though.

With this change, using hover distance limiter plugin, it is possible to make any series of inclusions and exclusions of 50, D0, and 90. While I understand that D0 being shown as closer than 90 is dumb that's what wacom decided to do so we will have to deal with it.

> Also L. 36 is not really accurate for distinguishing between D0 and 90. 50 is really close to "disconnecting" the pen from tracking, but D0 and 90 is almost all the tracking height of the MTE-450 in experimental tracking.

This is how NearProximity is intended to work. D0 and 90 should both be included in NearProximity (as they are) based off your description of them.

PR #2397, opened 2022-08-25
Add support for CTE-660

jamesbt365:
Verification: https://discord.com/channels/615607687467761684/615611007951306863/1012369379830288444

Also changes parser for the CTE-460 as it probably follows the same case as the CTE-660 which doesn't work with intuosreportparser

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-660.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-660.json
@@ -1,0 +1,35 @@
+{
+  "Name": "Wacom CTE-660",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 216.48,
+      "Height": 135.0,
+      "MaxX": 21648.0,
+      "MaxY": 13500.0
+    },
+    "Pen": {
+      "MaxPressure": 511,
+      "ButtonCount": 2
+    },
+    "AuxiliaryButtons": null,
+    "MouseButtons": null,
+    "Touch": null
+  },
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Configurations/Configurations/Wacom/CTE-460.json
```diff
@@ -23,7 +23,7 @@
       "ProductID": 106,
       "InputReportLength": 8,
       "OutputReportLength": null,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Intuos.IntuosReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Bamboo.BambooReportParser",
```
X9VoiD, line 26: Is CTE-460 not working before?

X9VoiD (changes requested):

OpenTabletDriver.Configurations/Configurations/Wacom/CTE-460.json
jamesbt365: No proof on it doesn't but its most likely the same case as the other ones after reportid was removed. I can revert this change if you don't think its the right call here

X9VoiD: Remove it.

jamesbt365: Reverted.

X9VoiD (approved):

PR #2403, opened 2022-08-26
Add more user guidance to build.sh and generate-rules.sh

Sublimelime:
Small PR that adds some more guidance to people using build.sh and generate-rules.sh to build and run OTD from source. Notably,

- Build.sh now checks for udev rules in the typical `/etc/udev/rules.d/` directory, and if not found, warns the user that they need to also run generate-rules.sh.
- Generate-rules.sh now explains what to do with the generated rules file.
- (added this in as a bonus) build.sh now spams less text while building, via providing the --verbosity=quiet tag. `dotnet publish` will still output errors (if any) just fine, but it won't spam individual files being built. 

The only thing unaddressed is running the binaries themselves, but I don't think that this is necessary - the script explains where the binaries are created by build.sh, and they're named obviously. You run `OpenTabletDriver.Daemon` for the daemon, etc. 

If the user wants automated running of OTD (and isn't planning on installing a package to have the service), they should set that up via their DE's startup system, potentially creating .desktop files pointing at the created binaries. Alternatively we could provide some pre-made .desktop files for the user to edit and move where they're supposed to go for their DE.

If they want to start OTD for dev work, a simple `./bin/OpenTabletDriver.Daemon` from the root folder works. No need for a dedicated run script.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -42,8 +42,8 @@ dotnet publish OpenTabletDriver.UX.Gtk ${options[@]} $@ || exit 3
 
 echo "Build finished successfully. Binaries created in ./bin"
 
-if [ ! -f /etc/udev/rules.d/99-opentabletdriver.rules ]; then
-    echo "\nUdev rules don't seem to be installed in /etc/udev/rules.d."
+if [ ! -f /etc/udev/rules.d/99-opentabletdriver.rules ] && [ ! -f /usr/lib/udev/rules.d/99-opentabletdriver.rules ]; then
+    echo "\nUdev rules don't seem to be installed in /etc/udev/rules.d or /usr/lib/udev/rules.d."
     echo "If your distribution installs them elsewhere, ignore this message."
     echo "If not, generate them by running generate-rules.sh, then follow the prompts."
 fi
--- a/build.sh
+++ b/build.sh
@@ -40,4 +40,10 @@ dotnet publish OpenTabletDriver.Console ${options[@]} $@ || exit 2
 echo -e "\nBuilding GTK UX...\n"
 dotnet publish OpenTabletDriver.UX.Gtk ${options[@]} $@ || exit 3
 
-echo "Build finished. Binaries created in ./bin"
+echo "Build finished successfully. Binaries created in ./bin"
+
+if [ ! -f /etc/udev/rules.d/99-opentabletdriver.rules ]; then
+    echo "\nUdev rules don't seem to be installed in /etc/udev/rules.d."
+    echo "If your distribution installs them elsewhere, ignore this message."
+    echo "If not, generate them by running generate-rules.sh, then follow the prompts."
+fi
--- a/build.sh
+++ b/build.sh
@@ -10,7 +10,7 @@ shift
 
 config=(--configuration='Release')
 
-options=(${config} --framework='net6.0' --self-contained='false' -v=q --output='./bin' \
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -40,4 +40,10 @@ dotnet publish OpenTabletDriver.Console ${options[@]} $@ || exit 2
 echo -e "\nBuilding GTK UX...\n"
 dotnet publish OpenTabletDriver.UX.Gtk ${options[@]} $@ || exit 3
 
-echo "Build finished. Binaries created in ./bin"
+echo "Build finished successfully. Binaries created in ./bin"
+
+if [ ! -f /etc/udev/rules.d/99-opentabletdriver.rules ]; then
```
jamesbt365, line 45: Check /usr/lib/udev/rules.d as well as this is also a place the rules could be.

X9VoiD (changes requested):
Do the same verbosity change on build.ps1 for consistency.

build.sh
```diff
@@ -40,4 +40,10 @@ dotnet publish OpenTabletDriver.Console ${options[@]} $@ || exit 2
 echo -e "\nBuilding GTK UX...\n"
 dotnet publish OpenTabletDriver.UX.Gtk ${options[@]} $@ || exit 3
 
-echo "Build finished. Binaries created in ./bin"
+echo "Build finished successfully. Binaries created in ./bin"
+
+if [ ! -f /etc/udev/rules.d/99-opentabletdriver.rules ] && [ ! -f /usr/lib/udev/rules.d/99-opentabletdriver.rules ]; then
```
gonX, line 45: I don't like this check, we should check if `/dev/uinput` can be opened, or a similar reliable check. Also check if we're being run as root, probably.
If you really wanna check for files I suggest also looking in `/usr/share/udev/` and checking for `*-opentabletdriver.rules` instead

gonX (changes requested):

build.sh
Sublimelime: >I don't like this check, we should check if /dev/uinput can be opened, or a similar reliable check. 

Doesn't this have to be done by the driver not by the build script? If I understand correctly...

>Also check if we're being run as root, probably.

We have no reason to be run as root, nor do anything with root. This is just a compilation script, and this PR is just to add info and guidance not functionality.

>If you really wanna check for files I suggest also looking in /usr/share/udev/ and checking for *-opentabletdriver.rules instead

I can add that directory too, but checking for * like that isn't necessary - we have it hardcoded to make 99-opentabletdriver.rules.

gonX: I've seen `90-opentabletdriver.rules` before. It's not necessary to assume that much about the users environment, maybe use `??-opentabletdriver.rules`? This should match anything that we'd be interested in - I don't think numbers higher than 99 are supported in this regard (since 100 would sort before 11).

> We have no reason to be run as root, nor do anything with root. This is just a compilation script, and this PR is just to add info and guidance not functionality.

This was mostly in relation if you wanted to go through with the `/dev/uinput` check. The check is pointless if its run as root, but @X9VoiD brought up a good point [on Discord](https://discord.com/channels/615607687467761684/628651971267788800/1012841999763767407) in regards to that we should also really be checking if tablet endpoints are accessible, but IMO at that point the complexity is probably too much.

So leaving it with a file check is fine.

gonX (approved):

X9VoiD (approved):

InfinityGhost (approved):

PR #2444, opened 2022-10-04
Initial port to .net 7

adryzz:
- [x] Ported all regex to the new source generator
- [ ] More stuff (?)
- [ ] I acknowledge that this checkbox serves no purpose whatsoever.

Yes, i know that this will take lots of time before getting merged, also because .net 7 isn't even out yet, but wanted to prepare early.

EDIT: also CI will likely fail


X9VoiD:
Split the regex changes to another PR.

Issue #1158, opened 2021-06-09
Mouse wheel as one of the buttons on pen


WoodyTheMighty:
Would be nice if you could bind the mousewheel to one  of the buttons like on hawku drivers so you can change audio and stuff with pen

PR #2453, opened 2022-10-13
WIP: Linux scroll bindings

InfinityGhost:


tylort69:
if this is ready to be added that would be super dope so I don't need to move my hand to my mouse every time I want to change the volume in osu and it also adds support for easily changing brush width in many applications

gonX:
Also fixes #1158

Would like to see this cleaned up with Windows and MacOS support though.

PR #2454, opened 2022-10-14
Fix crash when mutiple tablets are disconnected simultaneously.

AkiSakurai:
When tablets are disconnected using USB switch, InputDevices.Remove is invoked from DeviceReader thread, which causes race condition.


InfinityGhost (approved):
Looks good to me

InfinityGhost (approved):

PR #2463, opened 2022-10-20
Add checkbox to make sure the user reads

jamesbt365:
Relevant conversation: https://discord.com/channels/615607687467761684/628651971267788800/1032707136154763375


InfinityGhost:
I get that people aren't reading what they're checking but this is excessive imo. Any issues that fail to meet those requirements should instead be closed unless there's a legitimate reason for it to be ignored.

PR #2464, opened 2022-10-22
Deco 02 Area Size Correction

Satanepigone:
Based on both [vendor info](https://www.ugee.com.cn/goods/84.html) and my own measurement, the area is 254mm x 142.9mm. (or 142.875mm)

(Deco 02 seems have different LPI/LPmm on x-axis and y-axis)

p.s.: XP-Pen Deco 02 is the same hardware as UGEE RB170


jamesbt365:
The size should be based off the maximum x and y positioning / lpi

Also you based off the wrong branch. Should be on configs

Satanepigone:
It's a hardware design flaw. This model does have different LPI on x and y axes; both near 100px/mm (or 2540 LPI) but not same.

PR #2471, opened 2022-10-26
add profile and tool settings in diagnostics

### LINKED ISSUE #2201: Log plugin settings
## Description
Logging the configuration of individual plugins, similar to how we log which plugins are activated, would be handy for detecting misconfigured plugins in diagnostics.

adryzz:
Fixes #2201


X9VoiD (changes requested):
Fix nullability warnings.

gonX:
Please `git pull --tags` and `git push <your remote> --tags` so my system doesn't think I'm installing 0.5.2.3 :)

I was going to test but then I realized we don't have plugins for 0.7 yet, but I like the idea.

X9VoiD:
It also shows settings for builtin plugins, so those are testable.

gonX:
`master` is EOL. Please target this for either `0.6.x` or `avalonia`.

PR #2502, opened 2022-11-19
Fix pen mapping to mouse button does not work on Linux

a1exwang:
Previously if I set the pen binding 1 or 2 to a mouse button (e.g mouse right button or middle button), clicking the pen button does not trigger a ButtonPress event in artist mode.

It seems just because those events are not enabled

image

Tested on Ubuntu 22.04 with Gnome on X11.

Unknown issues: wayland session still does not work and I don't know why but at least we are getting somewhere.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -72,7 +72,12 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
                 EventCode.BTN_STYLUS2,
                 EventCode.BTN_STYLUS3,
                 EventCode.BTN_SIDE,
-                EventCode.BTN_EXTRA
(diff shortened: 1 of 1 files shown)


X9VoiD:
Those event codes are *never* meant to be enabled for a virtual "tablet". It's why it doesn't work on wayland. There are no physical tablets that sends out a mouse "click" while sending out pressure. It is nice to have this but it would be impossible to get this working consistently for all platforms, and most likely only works on X11.

X9VoiD:
Will close for now. fwiw, if you want mouse bindings then just use the mouse output modes (abs/rel mode)

gonX:
See d131beb84b93e0bf97744cf262b69bea45887f56 (from #1984) for how to better implement it

a1exwang:
> See [d131beb](https://github.com/OpenTabletDriver/OpenTabletDriver/commit/d131beb84b93e0bf97744cf262b69bea45887f56) (from #1984) for how to better implement it

Thanks! I just build from your branch for now and wait for your PR to be merged.

PR #2504, opened 2022-11-19
Linux Artist Mode: Expose ABS_DISTANCE to userspace

gonX:
This was unintentionally not enabled despite us emitting `ABS_DISTANCE` events

We're using a max value of 100 for now as that is unlikely to be out of bounds for any currently supported tablet

This supersedes dd358b477a95fc13feb2945f2027afa0286d5411 from #1984


X9VoiD (approved):
Lgtm, tho I am not sure how userspace handles this info anyway, and we don't have max hover as well so we can't normalize for now to this range.

gonX:
Gonna put this on the backlog since as you mentioned the userspace handling is questionable. I suggest we don't merge it before we also have max hover info, as I assume libinput will get confused if we don't do it correctly.

PR #2505, opened 2022-11-19
Linux Artist Mode: Allow usage of mouse buttons with Artist Mode

gonX:
This should probably be defined better for the user, ie:
LMB is actually pen down
RMB is actually barrel button 1
MMB is actually barrel button 2

- [ ] Checkbox to indicate discussion before merging

This supersedes d131beb84b93e0bf97744cf262b69bea45887f56 from #1984

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -147,5 +147,13 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
         {
             Device.Sync();
         }
+
+        protected override EventCode? GetCode(MouseButton button) => button switch
+        {
+            MouseButton.Left => EventCode.BTN_TOUCH,
+            MouseButton.Middle => EventCode.BTN_STYLUS2,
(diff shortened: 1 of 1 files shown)


X9VoiD:
This feature by itself is not something that we would want. Mouse "clicks" and pen "tips and buttons" are different things and different apps handle them differently. Labeling both as a "mouse binding" is not right as it would cause confusion later on when a pen tip or button event is handled differently from an actual mouse click/button event.

What should happen instead is we should allow a singular set of bindings, that is named generically, to "switch" function. So instead of "Mouse left click", it would be "Adaptive Pen Tip" where its meaning depends on the output mode.

So I agree with this change, but it should be followed by more changes. Will be opening up an issue and put further details there on what the implementation should be.

gonX:
Down the line the plan is to get rid of absolute mode entirely, so always having the necessary endpoint to enable usage of certain features, e.g. mouse buttons would require emulating a mouse clicking the same point as we're currently pointing with the tablet seems like the best plan.

I've converted it to a draft for now in anticipation of it being superseded by someone else's PR or me implementing that feature.

gonX:
Closing this PR in favor of a better implementation. VoiD brings up some good points on why this isn't ideal.

gonX:
This is now better implemented via #4010

PR #2506, opened 2022-11-19
Linux Artist Mode: Don't submit BTN_TOUCH based off pressure 

gonX:
The original intention of the code was under the assumption that the threshold slider rewrites pressure.

Since we do not rewrite pressure at the moment, it seems to behave libinput spec if we instead only send `BTN_TOUCH` when the tip is actually touching the tablet.

This supersedes 15b8341a2bd19f9193288d60ef986d70824a1148 and daf0fdce831e755beeb17aafb851fa84fc0769ed from #1984


X9VoiD (approved):

InfinityGhost:
Has this been tested yet?

Issue #2519, opened 2022-12-04
No tablets were detected with warning "Another tablet driver found: Gaomon"


nanakura:
### Acknowledgements

- [X] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [X] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

### Steps to Reproduce

1. Uninstall original Gaomon tablet driver
2. Restart windows
3. Open OpenTabletDriver.UX.Wpf.exe

### Expected Behavior

detect Gaomon tablet 

### Observed Behavior

no tablets detected and display some warning
![image](https://user-images.githubusercontent.com/69560752/205477477-29710011-91c0-4586-bccc-cfdfcb20230f.png)
But I have uninstalled the official Gaomon tablet driver

### Tablet Device

Gaomon M6

### Diagnostic Information

[diagnostics.json.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/10147979/diagnostics.json.txt)

jamesbt365:
There is no configuration for this tablet.

tablet > device string reader
9580
109
dump all, save that and send here.

nanakura:
> There is no configuration for this tablet.
> 
> tablet > device string reader 9580 109 dump all, save that and send here.

[dumps.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/10148282/dumps.txt)
It looks like it has an encoding problem...

jamesbt365:
thats just how it looks.

jamesbt365:
- Make a `Configurations` folder next to the executables of otd
- Download [this](https://cdn.discordapp.com/attachments/628651971267788800/1048913443614380042/M6.json) file
- Move it into that folder.

jamesbt365:
Once you have done that, it should detect but we aren't done yet.

nanakura:
It worked

jamesbt365:
we aren't done.

jamesbt365:
- Go into otd
- Go to `tablet > tablet debugger`
- Move your pen to the bottom right of your tablet (ignore the on screen visualiser)
- Get the maximum x and y coordinates that you can and the max pressure.
- Check if the pen buttons go to true when pressing them
[...]

PR #2520, opened 2022-12-04
Add new identifier to Gaomon M6

jamesbt365:
awaiting answer on #2519.


X9VoiD:
Is answered I guess. If not then we can just fix up later.

PR #2525, opened 2022-12-08
Migrate to .NET 7.0

X9VoiD:
- Centralized framework version to `Directory.Build.props` (we now explicitly support only two flavors of .NET, "Latest" and "LTS").
- Also added some default `.vscode/[tasks/launch].json`.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '6.0'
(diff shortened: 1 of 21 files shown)


jamesbt365:
supersedes #2444

InfinityGhost (changes requested):
Some improvements for the vscode launch files
Unfortunately GitHub won't allow .diff/.patch or any reasonable file extension on uploads.

[vscode.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/10181565/vscode.txt)

InfinityGhost (approved):

PR #2533, opened 2022-12-13
Disallow power throttling

### LINKED ISSUE #1999: Hawku and Devocub filters feel weird after minimizing OTD into the taskbar
## Description
Whenever i minimize OTD into the taskbar the filters start feeling 60hz on polling rate instead of the value i set in the plugin settings, same happens when running the daemon instead of the UX and sadly i cannot try this on windows 10 neither linux since i own a 12th gen intel cpu. Nothing shows up in the console and when testing on osu it shows the right polling rate but it still feels like it is set to 60hz or lower.

The only temporary fix i've found is to re-enable the filters and save&apply while not minimizing the OTD and letting the UX run in the background.

Do you guys know what might be causing this? Is it maybe a windows update causing this? It started happening a few weeks ago for me and until then it was fine.

## System Information:
| Name             | Value |
| ---------------- | ----- |
| Operating System         | Windows 11 |
| OpenTabletDriver Version | v0.6.0.0 |
| Tablet                   | CTL-472 |

X9VoiD:
- Fixes #1999

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -2,8 +2,11 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Collections.ObjectModel;
+using System.CommandLine.Invocation;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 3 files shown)


X9VoiD:
Gonna track instead entirely on #2534

PR #2534, opened 2022-12-13
Set process priority to high by default

X9VoiD:
- Depends on #2533


OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>$(FrameworkLatest)</TargetFramework>
     <NoWarn>VSTHRD100; VSTHRD101; VSTHRD110; VSTHRD200</NoWarn>
     <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
InfinityGhost, line 8: Maybe we should put this in `Directory.Build.props`?

InfinityGhost (commented):
Looks good

OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
X9VoiD: Probably on another PR, doesn't have to be now.

Issue #2312, opened 2022-06-16
Add support for Bamboo Pad CTH-301


fhilipecrash:
### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

### Manufacturer

Wacom

### Product Name

CTH-301

### Device Width

141.4mm

### Device Height

166.5mm

### Diagnostic Information

[log.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/8934621/log.txt)

jamesbt365:
For ease of reading upload the diagnostic as a txt instead of just pasting it into box, makes reading 10x  easier.

jamesbt365:
download this file, remove the .txt at the end of it then move it to `/etc/udev/rules.d`
[99-opentabletdriver.rules.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/8986540/99-opentabletdriver.rules.txt)
after you move it run the following commands

sudo udevadm control --reload-rules
sudo udevadm trigger
systemctl --user restart opentabletdriver

After you do this export a diagnostic again.

fhilipecrash:
Here is the new log
[log.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/9034718/log.txt)

jamesbt365:
fixed by #2545

PR #2545, opened 2022-12-19
Add support for Wacom Bamboo Pad USB CTH-301 tablets

aomsin2526:
This thing is different from any other wacom tablets, so new parser is needed. Pen, eraser, pen and aux button works. No touch support yet.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Wacom/BambooPad/BambooPadAuxReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Wacom/BambooPad/BambooPadAuxReport.cs
@@ -1,0 +1,23 @@
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using OpenTabletDriver.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Wacom.BambooPad
+{
+    public struct BambooPadAuxReport : IAuxReport
+    {
+        public BambooPadAuxReport(byte[] report)
+        {
+            Raw = report;
+
+            AuxButtons = new bool[]
+            {
+                report[23] == 1,
+                report[23] == 2
+            };
+        }
+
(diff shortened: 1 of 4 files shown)


jamesbt365:
closes #2312

32-bit-link:
Thank you aom sin for this magnificent pull request 🙏

OpenTabletDriver.Configurations/Parsers/Wacom/BambooPad/BambooPadAuxReport.cs
```diff
@@ -0,0 +1,23 @@
...
+    public struct BambooPadAuxReport : IAuxReport
+    {
+        public BambooPadAuxReport(byte[] report)
+        {
+            Raw = report;
+
+            AuxButtons = new bool[]
+            {
+                report[23] == 1,
+                report[23] == 2
```
X9VoiD, line 16: This prevents it from sending both buttons as active. Is it not possible for the hardware to send both?

```suggestion
                report[23].IsBitSet(0),
                report[23].IsBitSet(1)
```

X9VoiD (changes requested):

OpenTabletDriver.Configurations/Parsers/Wacom/BambooPad/BambooPadAuxReport.cs
aomsin2526: As far as i tested, This value is only returns 1 or 2, It is not possible to click both buttons as same time even on official drivers. It is also prioritize right button.

aomsin2526: After another test i found problem related to long press, maybe i will try improve it later.

X9VoiD (approved):

PR #2559, opened 2022-12-28
Add touch property to CTH-480's config

MafinBoy:
This is useless unless #2558 gets merged.


gonX (approved):
Doesn't necessarily depend on that PR IMO

Also I don't believe touch specifications have been completely ironed out yet, but this is a good start.

PR #2563, opened 2022-12-29
Rework updater code

X9VoiD:
## Design

Make the interface a bit more explicit of what it does by exposing two new classes:
- `UpdateInfo`
- `Update`

`IUpdater` provides `UpdateInfo` when an update is available. For now it only contains `Version`, but can be easily extended to contain more, for example, changelogs. `UpdateInfo` provides a way to get the actual `Update`, which entails downloading the actual binaries for the update.

The installation proceeds by passing the `Update` object to `IUpdater` via `IUpdater.Install`, where the binaries contained in `Update` is finally "installed".

`IUpdater` also has new events for future extensions (mainly targets UX):
- `UpdateInstalling`
- `UpdateInstalled`
- `RollbackCreated`

### Previous Backup Behavior

`Updater` had an explicit list of files/folders to include and exclude to its file operations. This makes it hard to maintain and easy to break.

Backup of binary files relies on knowing what the files are in advance, while backup of application data like `settings.json` is an *all-inclusive* operation (with a small exclusion to avoid infinite recursion).

That makes it possible to end up in situations where the backup folder contains data that we never wanted to be there.

### New Backup Behavior

*`Updater` now knows what files are going to be updated,* and appropriately responds by backing up only those files. For most cases, this is going to be fine as a backup. Plus, it doesn't need any need for explicitly listing the binaries that should be backup'ed.

Backup of application data is now controlled by a whitelist rather than a blacklist. For now, we are copying the following files/folders:

- `settings.json`
- `Presets`

## Testing

The original intention of this effort was to improve the reliability of updater tests. Due to design improvements, the results of the tests are now 100% reproducible, 100% of the time (no network access needed anymore either).

We are also now able to test the actual install and backup of stuff.


OpenTabletDriver.Desktop/Updater/IUpdater.cs
```diff
@@ -1,11 +1,38 @@
...
+        /// Installs an update.
+        /// </summary>
+        /// <param name="update">The update to install.</param>
+        /// <returns>The result of installing the update.</returns>
+        Task Install(Update update);
+
+        /// <summary>
+        /// Occurs when an update is being installed.
+        /// </summary>
+        event Action<Update> UpdateInstalling;
```
vedattt, line 26: No progress information during update? I suppose OTD updates may not be lengthy enough to justify the added complexity though.

X9VoiD: Yeah, no progress information for now.

PR #2632, opened 2023-02-05
Potential tilt support for Veikk tablets.

brkeejp:
Tested on Veikk vk640 graphic tablet.
previously I stated it but seems like no one implemented yet.

for my tablet:
default tilt is `00` which is 0 degree
minimum value is `b7` which is -73 degree
maximum value is `49` which is 73 degree

field `11` is X and field `12` is Y value

Don't know default output range but exported untouched state which is from -73 to 73. 1 value per degree.

image


X9VoiD (changes requested):
Keep this report parser unchanged and then do this on another parser. I suggest `VeikkTiltReportParser`. Tablets that rely on `Veikk(V2)ReportParser` but reports less than 13 bytes would have thrown `ArgumentOutOfRangeException`.

brkeejp:
Like you mentioned I created new parser, changed vk640's config file to use new parser for it.

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltedTabletReport.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            Position = new Vector2
+            {
+                X = Unsafe.ReadUnaligned<ushort>(ref report[3]) | (report[5] << 16),
+                Y = Unsafe.ReadUnaligned<ushort>(ref report[6]) | (report[8] << 16)
+            };
+            Tilt = new Vector2
+            {
+                X = unchecked((sbyte)report[11]),
+                Y = unchecked((sbyte)report[12])
```
X9VoiD, line 21: I don't think unchecked is needed here as this is a same-length cast.
```suggestion
                X = (sbyte)report[11],
                Y = (sbyte)report[12]

```

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltedReportParser.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            
+            // Aux Report
+            if (report[1] == 0x42) return new VeikkAuxReport(report);
+
+            // Touchpad Report
+            if (report[1] == 0x43) return new DeviceReport(report); // returns DeviceReport because of not supported yet
+
+            // Unknown Report
+            return new DeviceReport(report);
```
X9VoiD, line 30: No need to comment them like this, it adds no value to the code as it already reads the same as the comment.

Also no need to manually throw an argument out of range exception as well.

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltedReportParser.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using OpenTabletDriver.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Veikk
+{
+    public class VeikkTiltedReportParser : IReportParser<IDeviceReport>
```
X9VoiD, line 6: Change class name to `VeikkTiltReportParser` to match current convention.

X9VoiD (changes requested):

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltedTabletReport.cs
brkeejp: well actually it's effects how it's works. if i remove unchecked it will throw error.
`Constant value '128' cannot be converted to a 'sbyte' (use 'unchecked' syntax to override)`

otherwise i need to use formula but I'm not math guy so if I can achieve result simpler way I do.
because of that only part i can't touch.

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltedReportParser.cs
brkeejp: I can remove all those comments

X9VoiD (approved):

PR #2580, opened 2023-01-07
Configuration File for XP-Pen Artist 16 (2nd Gen)

SerialBoxes:
config file for the xp-pen artist 16 2nd gen based off of the config files for the other tablets in this same product family and tested on my device.

PR #2635, opened 2023-02-10
Added support for the XP-Pen Artist 16 (2nd Gen)

williambulin:
The specs were pulled from XP-Pen's official website for the dimensions: https://www.xp-pen.com/product/1203.html
Configuration files checked against the other models of the same generation by XP-Pen (such as the Artist 13 2nd Gen).
The MaxX and MaxY properties were acquired via the Tablet Debugger inside of OpenTabletDriver.
The Width/MaxX ratio matches the one of the other already submitted models (Artist 13 2nd Gen as a reference).

There is an issue where the last button on the tablet isn't being detected by the tablet debugger and therefore isn't usable at all.
However the raw tablet data does change with the 4th byte having it's 2nd bit flipped, it just isn't enumerated in the AuxButtons collection. (might have to modify the ReportParser?)

Asides from that button issue, everything else works as intended.

Self-tested on:
- Windows 11 22H2 (via the proprietary XP-Pen cable)
- macOS Ventura 13.1 (via a USB-C 3.1 Gen1 cable)

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Artist 16 (2nd Gen).json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Artist 16 (2nd Gen).json
@@ -9,9 +9,7 @@
     },
     "Pen": {
       "MaxPressure": 8191,
-      "Buttons": {
-        "ButtonCount": 2
-      }
+      "ButtonCount": 2
     },
     "AuxiliaryButtons": {
       "ButtonCount": 9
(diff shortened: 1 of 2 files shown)


williambulin:
Fixed the width and height of the digitizer settings and added support for the 10th button since it works on master. Everything works now and this PR is finished.

jamesbt365 (commented):
Duplicate of #2580 which has already been merged.

However there is a slight variation in the configurations.

You managed to get 1 line further on the X axis and 3 short on the Y axis. Which means yours is slightly more accurate on the X axis.

But, you don't seem to filter the report length at all? This may cause issues on some tablets and should be filtered to the exact lengths were possible, you haven't posted a diagnostic for us to go off so I don't have information to prove that it is a required change or your tablet differs from the one in #2580.

InfinityGhost:
Closing as it has been superseded by #2580, we can make any changes later if needed.

PR #2650, opened 2023-03-02
Add Wacom DTH-271 support

Kuuuube:
Aux buttons do not work.


gonX (approved):
LGTM, same as my comments in #3535 (theoretically possible but is unlikely to break parsing on IntuosV2 tablets)

PR #2658, opened 2023-03-14
Backport exception dialog

InfinityGhost:
Backported the new exception dialog from master to improve user experience.


X9VoiD:
This only makes the UX instantly crash without warning.

PR #2669, opened 2023-03-19
partial-absolute-wheel-support

joha4270:
This commit adds some support for tablets with rings (#1367), by treating the ring as a bunch of buttons.

It most certainly needs a second pass, but its far enough along that I would like some thoughts on the general approach/structure.

In short, this pull adds support for treating the ring as positions + 2 individual buttons, the first two buttons being clockwise and anticlockwise buttons. This can be configured in a new "Wheel" UI tab. 

This is a different approach than what #2256 does, which I was not aware of when I started on this. A final solution probably includes content from both.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/HS610.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/HS610.json
@@ -12,7 +12,10 @@
       "ButtonCount": 2
     },
     "AuxiliaryButtons": {
-      "ButtonCount": 12
+      "ButtonCount": 13
+    },
+    "Wheel": {
+      "StepCount": 12
     },
     "MouseButtons": null,
     "Touch": null
@@ -23,7 +26,7 @@
       "ProductID": 109,
       "InputReportLength": 12,
       "OutputReportLength": null,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicTiltReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Huion.HuionWheelReportParser",
       "FeatureInitReport": null,
       "OutputInitReport": null,
       "DeviceStrings": {
--- a/OpenTabletDriver.Desktop/Binding/BindingHandler.cs
+++ b/OpenTabletDriver.Desktop/Binding/BindingHandler.cs
@@ -14,6 +14,7 @@ namespace OpenTabletDriver.Desktop.Binding
         private readonly InputDevice _device;
         private readonly IServiceProvider _serviceProvider;
         private bool _isEraser;
+        private int? _lastWheelPosition = null;
 
         public BindingHandler(IServiceProvider serviceProvider, InputDevice device, BindingSettings settings, IMouseButtonHandler? mouseButtonHandler = null)
         {
@@ -37,6 +38,7 @@ namespace OpenTabletDriver.Desktop.Binding
 
             PenButtons = CreateBindingStates(settings.PenButtons, device, mouseButtonHandler);
             AuxButtons = CreateBindingStates(settings.AuxButtons, device, mouseButtonHandler);
+            WheelAsButtons = CreateBindingStates(settings.Wheel, device, mouseButtonHandler);
             MouseButtons = CreateBindingStates(settings.MouseButtons, device, mouseButtonHandler);
 
             MouseScrollDown = CreateBindingState<BindingState>(settings.MouseScrollDown, device, mouseButtonHandler);
@@ -48,6 +50,8 @@ namespace OpenTabletDriver.Desktop.Binding
 
         private Dictionary<int, BindingState?> PenButtons { get; }
         private Dictionary<int, BindingState?> AuxButtons { get; }
+
+        private Dictionary<int, BindingState?> WheelAsButtons { get; }
         private Dictionary<int, BindingState?> MouseButtons { get; }
 
         private BindingState? MouseScrollDown { get; }
@@ -74,6 +78,8 @@ namespace OpenTabletDriver.Desktop.Binding
                 HandleTabletReport(device.Configuration.Specifications.Pen!, tabletReport);
             if (report is IAuxReport auxReport)
                 HandleAuxiliaryReport(auxReport);
+            if(report is IAbsoluteWheelReport absoluteWheelReport)
+                HandleAbsoluteWheelReport(absoluteWheelReport);
             if (report is IMouseReport mouseReport)
                 HandleMouseReport(mouseReport);
         }
@@ -107,6 +113,32 @@ namespace OpenTabletDriver.Desktop.Binding
         private void HandleAuxiliaryReport(IAuxReport report)
         {
             HandleBindingCollection(report, AuxButtons, report.AuxButtons);
+        }
+
+        private void HandleAbsoluteWheelReport(IAbsoluteWheelReport report)
+        {
+            var wheelMovement = report.WheelPosition != null && _lastWheelPosition != null;
+
+            bool clockWise = false;
+            bool antiClockwise = false;
+            if (wheelMovement && _device.Configuration.Specifications.Wheel != null)
+            {
+                var steps = _device.Configuration.Specifications.Wheel.StepCount;
+                var halfSteps = steps / 2;
+                var treeHalf = halfSteps * 3;
+
+                var movement = (report.WheelPosition - _lastWheelPosition + treeHalf) % steps - halfSteps;
+
+                clockWise = movement < 0;
+                antiClockwise = movement > 0;
+            }
+
+            var movementButtons = new[] { clockWise, antiClockwise };
+            var fixedButtons = WheelAsButtons.Keys.Select(key =>  report.WheelPosition != null && key == report.WheelPosition);
+            var allButtons = movementButtons.Concat(fixedButtons).ToList();
+
+            HandleBindingCollection(report, WheelAsButtons, allButtons);
+            _lastWheelPosition = report.WheelPosition;
         }
 
         private void HandleMouseReport(IMouseReport report)
--- a/OpenTabletDriver/Tablet/IAbsoluteWheelReport.cs
+++ b/OpenTabletDriver/Tablet/IAbsoluteWheelReport.cs
@@ -1,0 +1,16 @@
+using JetBrains.Annotations;
+
+namespace OpenTabletDriver.Tablet
+{
+    /// <summary>
+    /// An auxiliary report containing states of express keys.
+    /// </summary>
+    [PublicAPI]
+    public interface IAbsoluteWheelReport : IDeviceReport
+    {
+        /// <summary>
+        /// The position of the wheel
+        /// </summary>
+        int? WheelPosition { get; set; }
+    }
+}
--- a/OpenTabletDriver/Tablet/WheelSpecifications.cs
+++ b/OpenTabletDriver/Tablet/WheelSpecifications.cs
@@ -1,0 +1,18 @@
+using System.ComponentModel;
+using JetBrains.Annotations;
+
+namespace OpenTabletDriver.Tablet
+{
+    /// <summary>
+    /// Device specifications for a wheel.
+    /// </summary>
+    [PublicAPI]
+    public class WheelSpecifications
+    {
+        /// <summary>
+        /// The amount of steps making up a whole rotation.
+        /// </summary>
+        [DisplayName("Steps")]
+        public uint StepCount { set; get; }
+    }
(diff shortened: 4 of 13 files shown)


OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -109,6 +115,32 @@ private void HandleAuxiliaryReport(IAuxReport report)
             HandleBindingCollection(report, AuxButtons, report.AuxButtons);
         }
 
+        private void HandleAbsoluteWheelReport(IAbsoluteWheelReport report)
```
X9VoiD, line 118: The way it is handled here right now can become problematic really quickly for wheels with relatively large amount of ticks. Imagine a wheel with more than 20 ticks. The use of LINQ + allocation here also makes this a bit unacceptable (tho i don't really like this dict-based binding system either in the very first place). Maybe we should keep it to just clockwise/counter-clockwise bindings for now until a better solution is found.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -109,6 +115,32 @@ private void HandleAuxiliaryReport(IAuxReport report)
...
+
+            bool clockWise = false;
+            bool antiClockwise = false;
+            if (wheelMovement && _device.Configuration.Specifications.Wheel != null)
+            {
+                var steps = _device.Configuration.Specifications.Wheel.StepCount;
+                var halfSteps = steps / 2;
+                var treeHalf = halfSteps * 3;
+
+                var movement = (report.WheelPosition - _lastWheelPosition + treeHalf) % steps - halfSteps;
```
Sublimelime, line 130:
```suggestion
                var tripleHalfSteps = (steps / 2) * 3;

                var movement = (report.WheelPosition - _lastWheelPosition + tripleHalfSteps) % (steps / 2);
```

Bit clearer imo (how I'd do it, feel free to disregard), also saves making 2 variables. Also, `steps - halfSteps` is just halfSteps because "x - 1/2 x" is just... 1/2 x.

OpenTabletDriver/Tablet/IAbsoluteWheelReport.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace OpenTabletDriver.Tablet
+{
+    /// <summary>
+    /// An auxiliary report containing states of express keys.
+    /// </summary>
+    [PublicAPI]
+    public interface IAbsoluteWheelReport : IDeviceReport
+    {
+        /// <summary>
+        /// The position of the wheel
```
Sublimelime, line 12: I would elaborate more here on what the numbers within the `int?` mean. What are valid values? Is it from -x to x? etc

Sublimelime (commented):
Looks sane so far to me, pending someone who knows a bit more doing their review.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
joha4270: Its `(math % steps) - halfSteps`, not `math % (steps - halfSteps)`

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -109,6 +115,32 @@ private void HandleAuxiliaryReport(IAuxReport report)
...
+
+                clockWise = movement < 0;
+                antiClockwise = movement > 0;
```
X9VoiD, line 133: That looks a bit too complicated, try adopting this instead. Kinda a simple geometric interpretation of what you're trying to do.
```cs
bool IsClockwise(uint prev, uint curr, uint max)
{
    uint half = max / 2;
    uint adjCurr = curr + max;
    uint adjPrev = prev + max;
    adjCurr = adjCurr - adjPrev;

    return adjCurr <= half;
}
```
Do tell if it fails with a certain case.

[Test cases](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRjzwgEspgACAWwEMAPSgXkoIIG4Tyq76AJWgBsAZs1YFKAekrpO+AgE4AFABIARAJEhKAb16bhAXwA6UNQEo5AFQCmAZ2BLMMSs5oNLeVIqWfcthycXAhdePwDHN3RQj2t7SJcAZhj6cPigygAWFLTAtwBWHLlvZVzHENcivAilaMr3VLjA5Pqwpsds1tjq9MKuxq8fMqU+lrbiXAAjAHtpwUoASTsAYUFpgGMAawB3UjsbJTIKSgAHMBsANxcjqnWIMDBr7gbzPB08Sk/KG8oACyFRCxeFIZHIvt9nrQ4AArZb3MBiO4PSiIBpgr4/KHQgAK5wuYjOlxRaI+XyxcORLHJ8MoAFpKFjcZc5KTPqgAOwMmEUhEAHhY/xEckME1QmUoNR+hKuEOOSMesp4HjerMoAGVgGByABzSjnOwQQRUFhLVYbHZ7A7SlzynKUAD8lDUZq2u32ako2md02gwBsYFpLot7vRbJ86h00sMdIAfLp5dGWDp9YbgIYLMKgA=)

X9VoiD (changes requested):

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
joha4270: I know both the new index and I'm caching the old position for the direction buttons. I _could_ invoke those two bindings specifically. 
That said, I'm unsure how useful all those buttons actually are, and I wouldn't be sad removing it.

That said, if I do remove the direction buttons, should I also try and route the clock-buttons toward the Aux bindings/UI? The Wheel UI page might be a little overkill for just 2 buttons.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
joha4270:
```
Test(0, 11, max);  //Correctly counter-clockwise
Test(11, 0, max);  //incorrectly counter-clockwise
```
(Movement passing zero)

I'm going to suggest the actual change needed here is a comment/better variable names explaining what's going on.

OpenTabletDriver/Tablet/WheelSpecifications.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// Device specifications for a wheel.
+    /// </summary>
+    [PublicAPI]
+    public class WheelSpecifications
+    {
+        /// <summary>
+        /// The amount of steps making up a whole rotation.
+        /// </summary>
+        [DisplayName("Steps")]
+        public uint StepCount { set; get; }
```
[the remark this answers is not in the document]
joha4270: Okay, you did just say absolute or relative, but I went ahead and added a few other members that will probably too show up in the future.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
joha4270: New, hopefully more understandable version updated.

OpenTabletDriver/Tablet/WheelSpecifications.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Does a increment/positive movement indicate clockwise movement?
+        /// </summary>
+        [DisplayName("Clockwise")]
+        public bool IsClockwise { get; set; }
+
+        /// <summary>
+        /// For relative wheels, the angle on the unit circle, corresponding to a reading of zero from the sensor
+        /// </summary>
+        public float AngleOfZeroReading { get; set; }
```
X9VoiD, line 33: I think you meant "For absolute wheels" here. We can't ever tell where a relative wheel is currently pointing at so this would be of no use for those.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -109,6 +112,44 @@ private void HandleAuxiliaryReport(IAuxReport report)
...
         }
 
+        private void HandleAbsoluteWheelReport(IAbsoluteWheelReport report)
+        {
+            var wheelSpec = _device.Configuration.Specifications.Wheel;
+            if (wheelSpec == null)
+                return;
+
+            //The wheel direction buttons are just handled as extra Auxiliary buttons. So we need to know how many buttons we skip
+            var auxiliaryButtons = (int)(_device.Configuration.Specifications.AuxiliaryButtons?.ButtonCount ?? 0u);
```
X9VoiD, line 122: Still not sold on this one since this makes it indistinguishable from the normal tablet buttons. Should change `AuxPanel.Content` such that it holds two groups, one for aux bindings, another for wheel bindings. Similar to how it was done for the pen-related bindings.

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -109,6 +112,44 @@ private void HandleAuxiliaryReport(IAuxReport report)
...
+            {
+                var movement =
+                    ComputeShortCircleDistance(_lastWheelPosition.Value, report.WheelPosition.Value, wheelSpec.StepCount);
+
+                bool clockWise = movement < 0;
+                bool antiClockwise = movement > 0;
+
+                HandleBindingCollection(report, AuxButtons, new[] { clockWise, antiClockwise }, auxiliaryButtons);
+                //It should only be a momentary press, otherwise movement and stop with resting finger will be read as a long press
+                HandleBindingCollection(report, AuxButtons, new[] { false, false }, auxiliaryButtons);
```
X9VoiD, line 138: As I hinted on other review, wheel bindings should be separate from `AuxButtons`, maybe `BindingState WheelClockwise` and `BindingState WheelCounterClockwise`?

> see `MouseScrollUp` and `MouseScrollDown`

X9VoiD (changes requested):

X9VoiD (dismissed):
Lookin real nice now! :bread: Still some really smol stuff left but nothing that I can't fix myself before merge.

joha4270:
Slight update, now also supports just binding the Wheel as if it was a button.

X9VoiD:
Revert the last commit. That'll only be confusing especially on tablets that do not support "sensing" when the wheel is touched but not rotated yet.

joha4270:
And Done


Den søn. 2. apr. 2023 kl. 14.52 skrev X9VoiD ***@***.***>:

OpenTabletDriver.Desktop/Binding/BindingHandler.cs
```diff
@@ -117,6 +160,13 @@ private void HandleMouseReport(IMouseReport report)
             MouseScrollUp?.Invoke(report, report.Scroll.Y > 0);
         }
 
+        /// <summary>
+        /// Updates the button(binding) states for a collection of bindings based upon an array of buttons reported by a device
+        /// </summary>
+        /// <param name="report">The report from the device</param>
+        /// <param name="bindings">The collection of bindings we are updating</param>
+        /// <param name="newStates">New states of the updated bindings</param>
+        /// <param name="offset">An offset into <paramref name="bindings"/> before <paramref name="newStates"/> applies</param>
```
gonX, line 169: This parameter is missing from the function, so this parameter annotation should be removed (or the functionality should be added to the function if still needed)
```suggestion
```

OpenTabletDriver.Configurations/Configurations/Huion/HS610.json
```diff
@@ -12,7 +12,13 @@
       "ButtonCount": 2
     },
     "AuxiliaryButtons": {
-      "ButtonCount": 12
+      "ButtonCount": 13
```
gonX, line 15: I don't think this should be changed. If it's incremented because it has a wheel that can be clicked this should be added as an attribute to the Wheel properties instead, e.g. `IsClickable` or similar naming.

gonX:
Looks generally pretty good to me, though I would have to play around with the PR myself and see how it behaves.

OpenTabletDriver.Configurations/Configurations/Huion/HS610.json
gonX: I received a reply on Discord:
> Yeah, probably not actually related to the core changes, but the tablet does actually have a 13th button in the center of the wheel

I think I would still prefer wheel to have a separate declaration. Thoughts @X9VoiD ?

iErik:
So.... is this still going to happen?

gonX:
This will be superseded by a forward-port of #3497 to ensure the codebase doesn't deviate too much. Thanks for the well-attempted contribution, and sorry for letting this hang for so long.

PR #2699, opened 2023-04-10
Fix: unable to find auxiliary devices

pirminkager:
I noticed that no touch events where registered with my tablet.
Looks like a variable name was not changed and causes that no auxiliary devices are found.


gonX:
I haven't tested it but that looks very correct to me. Nice catch. Looks like this was introduced in #2583

PR #2750, opened 2023-05-22
Fix PenIsInRange not respecting OOR report

X9VoiD:


X9VoiD:
All usages of PenIsInRange have an earlier check that makes sure that the report is not an instance of `OutOfRangeReport`, which makes this redundant.

PR #2787, opened 2023-06-11
Add packaging scripts

X9VoiD:
Supersedes [OpenTabletDriver.Packaging](https://github.com/OpenTabletDriver/OpenTabletDriver.Packaging).

## Tested Platforms

- [x] Windows
- [x] Debian-based
- [x] RPM-based
- [x] MacOS

### General

- Simplified packaging to improve maintainability and readability.

### Linux

- Packages now use `/usr/lib/opentabletdriver` to stash the real binaries instead of `/usr/share/OpenTabletDriver`. This is to comply better to FHS (`/usr/share` can *only* contain arch-independent files).
- Binary tarball is now better structured for easier system administration/automation to those that uses it.
- Changed systemd user service to start explicitly after `default.target`.
- Improved the kernel module blacklist to apply during boot and runtime (previously it only blocks module loading during boot).


eng/linux/Debian/package.sh
```diff
@@ -0,0 +1,75 @@
...
+Architecture: amd64
+Installed-Size: $(du -s "${output}" | cut -f1)
+Pre-Depends: udev, bash
+Depends: libevdev2, libgtk-3-0, dotnet-runtime-6.0, ${shlibdeps}
+Recommends: libx11-6, libxrandr2
+Conflicts: ${OTD_NAME}
+Replaces: ${OTD_NAME}
+Maintainer: ${OTD_MAINTAINERS[1]}
+Description: A ${OTD_DESC}
+ ${OTD_NAME} has the highest number of supported tablets with great
```
gonX, line 56: Can these sentences be put into a config file that other packaging scripts can source from as well? seems useful to have centralized

build.sh
gonX, line 1: can get rid of this file if you like, just coordinate with packagers first

eng/linux/Generic/usr/lib/modprobe.d/99-opentabletdriver.conf
```diff
@@ -0,0 +1,2 @@
+install wacom /usr/bin/true
+install hid_uclogic /usr/bin/true
```
gonX, line 2:
```suggestion
blacklist wacom
blacklist hid_uclogic
```

Otherwise users would need to uninstall OpenTabletDriver entirely to be able to load the vendor drivers, which may not be desirable for some setups.

eng/linux/Generic/usr/lib/systemd/user/opentabletdriver.service
```diff
@@ -0,0 +1,11 @@
...
+Description=OpenTabletDriver Daemon
+After=default.target
+
+[Service]
+ExecStart=otd-daemon
+Restart=always
+RestartSec=3
+
+[Install]
+WantedBy=default.target
```
gonX, line 11: This won't work because we can't ensure that environment variables are ready yet. I suggest the following instead:

```suggestion
[Unit]
Description=OpenTabletDriver Daemon
PartOf=graphical-session.target

[Service]
ExecStart=otd-daemon
Restart=always
RestartSec=3

[Install]
WantedBy=graphical-session.target
```

See "_2) Set up a dependency chain so that your DE starts a systemd target (advanced)_" in https://github.com/OpenTabletDriver/OpenTabletDriver.Web/issues/43#issue-1105884672 for more details.

The alternative for starting should be starting it on DE start, ie. literally `otd-daemon` in the DE/WM's autostart.

generate-rules.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
+set -eu
```
gonX, line 3: I would prefer just using `set -u` if anything. `set -e` breaks some expected behaviors (like some uses of `&&` and `||` in if's) - better to just affix commands with ` || exit $?`

I suggest reading the `-e` section in `man 1 bash`

(I might have missed this for other BASH scripts but consider this review as for all of them)

gonX (changes requested):
overall nice work

eng/linux/Debian/package.sh
X9VoiD: Would moving it to `lib.sh` suffice? Will remove the extra whitespaces as well.

build.sh
X9VoiD: Won't remove. It still has its use. (provides sane defaults for quickly doing a build and testing things out)

eng/linux/Debian/package.sh
gonX: yes definitely

eng/linux/Generic/usr/lib/modprobe.d/99-opentabletdriver.conf
X9VoiD: The goal is to make OpenTabletDriver not undesirable, so I think keeping those modules unloadable is fine.

gonX: After further research, `modprobe -i` can circumvent `install`s - so leaving this in as-is is fine.

eng/linux/Debian/package.sh
X9VoiD: Done

eng/linux/lib.sh
```diff
@@ -0,0 +1,168 @@
...
+	"${REPO_ROOT}/generate-rules.sh" -v "${REPO_ROOT}/OpenTabletDriver.Configurations/Configurations" "${output_file}" > /dev/null
+}
+
+generate_desktop_file() {
+  local output="${1}"
+
+  mkdir -p "$(dirname "${output}")"
+  cat << EOF > "${output}"
+[Desktop Entry]
+Version=${OTD_VERSION}
```
hwsmm, line 85: This entry is not for the software version, but for desktop entry spec version

X9VoiD: Fixed, should be 1.5 i think?

PR #2839, opened 2023-07-22
Add support for UGEE U1200 tablet

ndkean:
I own the tablet, and thus tested it. I also measured the dimensions of the digitizer myself, as the ones on [UGEE's website](https://shop.ugee.com/products/ugee-drawing-tablet-1200) were off by about a millimeter.

Diagnostics: [u1200 diagnostics.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/12135804/u1200.diagnostics.txt)

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/UGEE/U1200.json
+++ b/OpenTabletDriver.Configurations/Configurations/UGEE/U1200.json
@@ -2,8 +2,8 @@
   "Name": "UGEE U1200",
   "Specifications": {
     "Digitizer": {
-      "Width": 263.5,
-      "Height": 147.5,
+      "Width": 263.22,
+      "Height": 148.055,
       "MaxX": 52644,
       "MaxY": 29611
     },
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Configurations/Configurations/UGEE/U1200.json
```diff
@@ -0,0 +1,37 @@
+{
+  "Name": "UGEE U1200",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 263.5,
+      "Height": 147.5,
```
X9VoiD, line 6: If using the digitizer's sensor resolution, the width and height should be

```suggestion
      "Width": 263.22,
      "Height": 148.055,
```
> `expected size = sensor res / lpi` where `lpi` is 200

The width is most likely more accurate and should change to it, but can you confirm if this suggested height is closer to reality? Instead of measuring the marked active area on the surface of the tablet, try measuring the bounds where the digitizer says it's 0 and whatever is the maximum for that axis.

When in doubt or can't measure with that much accuracy, just use this suggested width and height.

X9VoiD (changes requested):

OpenTabletDriver.Configurations/Configurations/UGEE/U1200.json
ndkean: Yes, those values do appear to be more accurate when measuring that way. Thanks for the suggestion.

PR #2853, opened 2023-07-27
Fix 10moon 1060 parser

user202729:
Refer: #1819 .

I explain the rationale for various modifications below.


OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
```diff
@@ -8,7 +8,7 @@
       "MaxY": 4095
     },
     "Pen": {
-      "MaxPressure": 8191,
```
user202729, line 11: Note that the calculation of pressure takes 1800 subtract some value, so the maximum pressure is clearly at most 1800.

OpenTabletDriver.Configurations/Parsers/10moon/10moonAuxParser.cs
```diff
@@ -1,31 +0,0 @@
...
-                report[12] == 0x33 && !report[11].IsBitSet(0), // Tenth
-                report[12] == 0x33 && !report[11].IsBitSet(1), // Eleventh
-                report[12] == 0x33 && !report[11].IsBitSet(2) // Twelveth
-            };
-        }
-
-        public bool[] AuxButtons { set; get; }
-        public byte[] Raw { set; get; }
-    }
-}
```
user202729, line 31: This part is deleted and merged into the other part, the tablet is capable of reporting both coordinate and auxiliary button at the same time.

OpenTabletDriver.Configurations/Parsers/10moon/10moonTabletReport.cs
```diff
@@ -18,16 +18,33 @@ public TenMoonTabletReport(byte[] report)
 
             var buttonPressed = (report[9] & 6) != 0;
             var prePressure = report[5] << 8 | report[6];
-            Pressure = (uint)(0x0672 - (prePressure - (buttonPressed ? 50 : 0)));
```
user202729, line 21: I increase the value slightly here, the offset is small (only 150). Experimentally, sometimes the calculation before could overflow.

OpenTabletDriver.Configurations/Parsers/10moon/10moonTabletReport.cs
```diff
@@ -18,16 +18,33 @@ public TenMoonTabletReport(byte[] report)
...
+            Pressure = (uint)(1800 - (prePressure - (buttonPressed ? 50 : 0)));
 
             PenButtons = new bool[]
             {
-                report[9].IsBitSet(2),
+                (report[9] & 6) == 4,
                 (report[9] & 6) == 6
```
user202729, line 26: Note that if `(report[9] & 6) == 6` is true then `report[9].IsBitSet(2)` is true as well, so previously the code would erroneously report that both buttons are pressed. Now the bug no longer happen.

OpenTabletDriver.Configurations/Parsers/10moon/10moonTabletReport.cs
```diff
@@ -18,16 +18,33 @@ public TenMoonTabletReport(byte[] report)
...
+                !report[11].IsBitSet(6), // Third
+                !report[11].IsBitSet(5), // Fourth
+                !report[11].IsBitSet(4), // Fifth
+                !report[11].IsBitSet(3), // Sixth
+                !report[12].IsBitSet(4), // Seventh
+                !report[12].IsBitSet(0), // Eightth
+                !report[12].IsBitSet(5), // Nineth
+                !report[11].IsBitSet(0), // Tenth
+                !report[11].IsBitSet(1), // Eleventh
+                !report[11].IsBitSet(2) // Twelveth
```
user202729, line 42: The board is actually capable of reporting multiple auxiliary keys being pressed at the same time.

X9VoiD:
Linking #2615 here for review purposes. They're both trying to fix the same class.

tiagoporsch:
Tested and works fine. What else needs to be done to merge this?

user202729:
Has been a long time but 

* looks like there's some guideline saying `master` shouldn't be targeted https://github.com/OpenTabletDriver/OpenTabletDriver/commit/271285aee056f81819d18e2a3f30e92182aad6d2
* there's some undesirable clash with other tablets for some reason https://github.com/OpenTabletDriver/OpenTabletDriver/commit/fd7f7d227c8e52d289a1b217a4fc0cd6e673912d (because I don't have other tablets it is hard for me to test)

I'll see what options there are.

gonX (changes requested):

Issue #2377, opened 2022-08-04
Fails to load with WACOM Graphire 2 Graphics Art TABLET ET-0405A-U


snovotill:
### Acknowledgements

- [X] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [X] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

### Steps to Reproduce

Connect WACOM Graphire 2 Graphics Art TABLET ET-0405A-U and run OpenTabletDriver.
The driver detects the tablet and then barfs IndexOutOfRangeException:Error as shown below:

[Wacom ET-0405A-U:Info] Output mode: Absolute Mode
[Wacom ET-0405A-U:Info] Display area: [3328x1440@<1664, 720>:0°],
[Wacom ET-0405A-U:Info] Tablet area: [127.6x92.8@<63.8, 46.4>:0°],
[Wacom ET-0405A-U:Info] Clipping: Enabled
[Wacom ET-0405A-U:Info] Ignoring reports outside area: Disabled
[Wacom ET-0405A-U:Info] Tip Binding: [Mouse Button Binding: Left]@0%
[Settings:Info] Driver is enabled.
[IndexOutOfRangeException:Error]        Index was outside the bounds of the array.
   at OpenTabletDriver.Configurations.Parsers.Wacom.Intuos.IntuosTabletReport..ctor(Byte[] report)
   at OpenTabletDriver.Configurations.Parsers.Wacom.Intuos.IntuosReportParser.GetToolReport(Byte[] report)
   at OpenTabletDriver.Configurations.Parsers.Wacom.Intuos.IntuosReportParser.Parse(Byte[] report)
   at OpenTabletDriver.Devices.DeviceReader`1.Main()

### Expected Behavior

Should just load and then the tablet should show in the GUI.

### Observed Behavior

Attempts to load but fails. See steps to reproduce.

### Tablet Device

WACOM Graphire 2 Graphics Art TABLET ET-0405A-U

### Diagnostic Information

{
  "App Version": "OpenTabletDriver v0.6.0.4",
  "Build Date": "5/4/2022",
  "Operating System": {
    "Platform": 2,
    "ServicePack": "",
    "Version": "10.0.22000.0",
[...]

PR #2857, opened 2023-07-28
Fix ET-0405A-U config

Kuuuube:


jamesbt365:
Does this fix #2377? Should do?

Issue #2873, opened 2023-08-16
Add support for musl libc to OpenTabletDriver


Thymelizabeth:
### Description

I am using Gentoo Linux using musl libc, and I would like to be able to use OpenTabletDriver. 

The only thing that I needed to do to get the program to work (compiling from the latest release) was change the `NET_RUNTIME` variable to `linux-musl-x64` in eng/linux/package.sh.

The [dotnet install script](https://github.com/dotnet/install-scripts/blob/main/src/dotnet-install.sh) uses a function `is_musl_based_distro` to check the libc.

### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

PR #2874, opened 2023-08-16
Add build support for musl

Thymelizabeth:
Adds a check in the build system for [musl libc](https://musl.libc.org/).


Thymelizabeth:
Resolves #2873

PR #2887, opened 2023-08-20
Add new string for Huion Kamvas 20

jamesbt365:
Verification: https://discord.com/channels/615607687467761684/615611007951306863/1142601112738877632
Pre support diagnostic: https://cdn.discordapp.com/attachments/615611007951306863/1142580164472090714/diag.json
Strings: https://cdn.discordapp.com/attachments/615611007951306863/1142599438863450173/huiondumpy.txt

No post-support diagnostic at this time.


jamesbt365:
Note: this was applied to the wrong tablet due to a fuckup, but the max x and y seem like they should be on this tablet, not sure how to handle this as it may cause a slight knew in the area without proper from a user with this string

PR #2919, opened 2023-09-06
fix: switch to tilt report parser on HS611

jamesbt365:
This is on hotfix, but not master. I missed it when I was adding the identifier in #2908


jamesbt365:
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/0.6.x/OpenTabletDriver.Configurations/Configurations/Huion/HS611.json

And I just realized the other variant of it too isn't using the tilt report parser on 0.6.x ugh.

PR #2819, opened 2023-06-30
Add Linux distro logging to platform initialization

DanisDGK:
Closes #2744

PR #2944, opened 2023-09-30
Parse /etc/os-release in diagnostics

### LINKED ISSUE #2744: Add Linux distribution info to exported diagnostics
<details><summary>Acknowledgements</summary>
<p>

</p>
</details> 

[Discord convo beginning](https://discord.com/channels/615607687467761684/628651971267788800/1109967778959601766)

## Description

Currently, exported diagnostics on Linux do not contain distro-specific info, e.g.:

```json
"Operating System": {
  "Platform": 4,
  "ServicePack": "",
  "Version": "5.10.0.23",
  "VersionString": "Unix 5.10.0.23"
}
```

According to [a 2011 unix SE answer](https://unix.stackexchange.com/a/6348/510663):

> To get `OS` and `VER`, the latest standard seems to be `/etc/os-release`. Before that, there was `lsb_release` and `/etc/lsb-release`. Before that, you had to look for different files for each distribution.

Assuming this statement still holds up, it might be helpful to include [`/etc/os-release`](https://man.archlinux.org/man/os-release.5) info such as `NAME`, `ID`, `PRETTY_NAME`, `VERSION`, `BUILD_ID` within diagnostics output.

gonX:
Closes #2744 
Supersedes #2819 

Pre-PR diagnostics on Linux:
```json
  "OperatingSystem": {
    "Platform": 4,
    "ServicePack": "",
    "Version": "6.5.5.273",
    "VersionString": "Unix 6.5.5.273"
  }
```

Post-PR diagnostics on Arch Linux:
```json
  "OperatingSystem": {
    "NAME": "Arch Linux",
    "PRETTY_NAME": "Arch Linux",
    "ID": "arch",
    "BUILD_ID": "rolling"
  }
```

All `os-release` pairs are added, except for a (small) key blacklist (currently `LOGO` and `ANSI_COLOR`) as well as any key containing `URL`.

On non-Linux there is a very small (likely wanted) side-effect of having the Platform field be correctly stringified - previously it would post e.g. `"Platform": 4` but now it posts `"Platform": "Unix"`

Code or naming is probably messy, reviews are very welcome in those areas.


OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        public const string OS_RELEASE_PATH = "/etc/os-release";
+
+        // keys that will never help with diagnostics
+        static readonly string[] IGNORED_KEYS = {
+            "ANSI_COLOR",
+            "LOGO"
+        };
+
+        public IDictionary<string,string> Raw;
```
vedxyz, line 17:
```suggestion
        public IDictionary<string, string> Raw;
```

Been a while since I touched C#, but I think this is the correct style?

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public LinuxEtcOSRelease()
+        {
+            if (!File.Exists(OS_RELEASE_PATH))
+                throw new Exception("os-release not found");
+
+            Raw = parse();
+        }
```
vedxyz, line 25: This is really bit of a nit, but I'm not a fan of IO operations in the constructor. I personally would've kept this class instantiable independent of IO, and move the construction with IO to a static method (`FromFile()`? or `FromOsReleaseFile()`?). Might even make the method have a nullable return type instead of throwing an exception.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        static IDictionary<string,string> parse()
```
vedxyz, line 27: Should be pascal case and private?

vedxyz, line 27: On another note, I think the manpage and Python implementations can be consulted for a more robust parser. Although I'm unsure how much of a risk the current implementation poses, if any.

- See example 5 in https://www.freedesktop.org/software/systemd/man/os-release.html
- https://github.com/python/cpython/blob/679af65ca3798dc36dedabfdc66303b4cee49373/Lib/platform.py#L1233-L1287
  - Can be tested too, just saying https://github.com/python/cpython/blob/adf0f15a06c6e8ddd1a6d59b28efcbb26289f080/Lib/test/test_platform.py#L511-L539

vedxyz (changes requested):
I've intended to make a PR for the parent issue several times but still haven't gotten around to it. I would've liked to say I can help with the stuff I've left reviews for here, but not trusting myself on that currently.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+
+namespace OpenTabletDriver.Desktop.Diagnostics
+{
+    public class LinuxEtcOSRelease
```
vedxyz, line 7:
```suggestion
    public class LinuxOSRelease
```

`Etc` irrelevant IMO.

OpenTabletDriver.Desktop/Diagnostics/IDiagnosticInfo.cs
```diff
@@ -11,7 +11,7 @@ public interface IDiagnosticInfo
     {
         string AppVersion { get; }
         string BuildDate { get; }
-        OperatingSystem OperatingSystem { get; }
+        IDictionary<string, string> OperatingSystem { get; }
```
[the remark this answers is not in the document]
gonX: `OperatingSystem` isn't universally useful for us - for the most part we just need the `VersionString`. I re-implemented the rest of the fields in case dotnet decides to expand its functionality on non-Windows systems.

I wish there was a cleaner way to integrate `OperatingSystem` as a Dictionary without the roundabout way of setting the fields individually done currently in the PR.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        public const string OS_RELEASE_PATH = "/etc/os-release";
+
+        // keys that will never help with diagnostics
+        static readonly string[] IGNORED_KEYS = {
+            "ANSI_COLOR",
+            "LOGO"
+        };
```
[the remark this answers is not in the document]
gonX: Could also just stick with just `ID`, extrapolating the other information should be simple.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
gonX: Implicitly private and private methods should be camelCase afaik. If there is precedence for explicit usage of `private` I have missed that. Will check up on this.

OpenTabletDriver.Desktop/Diagnostics/DiagnosticInfo.cs
```diff
@@ -23,11 +24,34 @@ EnvironmentDictionary environmentDictionary
...
         public IEnumerable<LogMessage> ConsoleLog { get; }
 
+        // behaves similarly to if Environment.OSVersion is passed directly to our serializer
+        private static IDictionary<string,string> OSVersionToDict()
+        {
+            IDictionary<string, string> rv = new Dictionary<string, string>();
+            rv.Add("Platform", Environment.OSVersion.Platform.ToString());
+            rv.Add("ServicePack", Environment.OSVersion.ServicePack);
+            rv.Add("Version", Environment.OSVersion.Version.ToString());
+            rv.Add("VersionString", Environment.OSVersion.VersionString);
```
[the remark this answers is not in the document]
gonX: Agree with this, add a suggestion if you want the commit credit

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
gonX: Can just move it into a static function altogether. I mostly wanted it to be an instantiated (and singleton) class so that there'd be less overhead on multiple runs. But with the way things are implemented now it just makes the code harder to read, so I do plan to rework this.

OpenTabletDriver.Desktop/Diagnostics/IDiagnosticInfo.cs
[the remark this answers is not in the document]
gonX: I re-read your comment and I did initially go for a class that implemented those fields, but it would require the class to also be serializable. I ran into some roadblocks in regards to that, hence the current solution.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
gonX: I have not seen those examples but it doesn't look too different from what it's doing now - only immediate difference I'm seeing is they're parsing with a regex, whereas currently the proposed code checks for presence of `=` before parsing the line, which is effectively has the same end result.

A test would be cool but I am not well-versed in C# testing infra. Leaving convo open for last for input on that.

gonX: (actually the code currently doesn't trim the value which we might want to do)

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
vedxyz: Yeah good point, this class could have a nullable, static, cached property `Instance` or something instead of a public constructor or static method.

OpenTabletDriver.Desktop/Diagnostics/IDiagnosticInfo.cs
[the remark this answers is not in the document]
vedxyz: Hmm, the main thing I'm not a fan of is that the contents of `OperatingSystem` (the property) are fluid and inconsistent across OSes as a dictionary. On Linux it's a dump of `os-release` keys, but on non-Linux it's actually structured but with totally different keys.

Setting the fields individually should also not be needed for a new class that can capture relevant OS info (for our purposes) for any OS. It'll just get serialized to JSON directly, might need a couple annotations though.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
vedxyz: Should've added more context. I say irrelevant because `etc/` is not the only place `os-release` can exist, per the manpage. (This is also something that might be good to consider for the parser logic, though IDK how many distros require handling this edge case)

OpenTabletDriver.Desktop/Diagnostics/IDiagnosticInfo.cs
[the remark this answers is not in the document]
vedxyz: Oh, I posted before refreshing the page so I missed your later comments.

OpenTabletDriver.Desktop/Diagnostics/LinuxEtcOSRelease.cs
gonX: Resolved next push

gonX: I did a pass over the code, most notably it was missing handling of comments, which will be added in an upcoming push.

gonX:
I sadly had to butcher the history (a lot of overlapping changes), but thanks for the review @vedxyz

OpenTabletDriver.Desktop/Diagnostics/DiagnosticInfo.cs
```diff
@@ -23,11 +24,34 @@ EnvironmentDictionary environmentDictionary
...
+            {
+                switch (SystemInterop.CurrentPlatform)
+                {
+                    case SystemPlatform.Linux:
+                        return LinuxOSRelease.Raw;
+                    default:
+                        return OSVersionToDict();
+                }
+            }
+        }
```
vedxyz, line 39:
```suggestion
        public IDictionary<string, string> OperatingSystem => SystemInterop.CurrentPlatform switch
        {
            SystemPlatform.Linux => LinuxOSRelease.Raw,
            _ => OSVersionToDict(),
        };
```

@X9VoiD reminded me that this exists. Didn't test but should compile/work?

OpenTabletDriver.Desktop/Diagnostics/LinuxOSRelease.cs
```diff
@@ -0,0 +1,89 @@
...
+                        continue;
+
+                    // ignore comments
+                    if (line.StartsWith('#'))
+                        continue;
+
+                    var parsed = ParseLine(line);
+
+                    // check whitelist
+                    if (Array.IndexOf(ACCEPTED_KEYS, parsed.Key) == -1)
```
vedxyz, line 55: Could have used LINQ's `Contains` here, but no big deal.

vedxyz (approved):
LGTM. Just two small style suggestions.

---

I'm still of the same opinion as my [earlier comment](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/2944/files#r1342183464) regarding dictionary vs class, but no biggie. @X9VoiD said he might have another look anyway so he might go for it if he sees fit.

OpenTabletDriver.Desktop/Diagnostics/DiagnosticInfo.cs
gonX: Will fixup if it doesnt work, looks familiar to me

OpenTabletDriver.Desktop/Diagnostics/LinuxOSRelease.cs
gonX: Makes sense, that's more clear

gonX:
> I'm still of the same opinion as my [earlier comment](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/2944/files#r1342183464) regarding dictionary vs class, but no biggie. @X9VoiD said he might have another look anyway so he might go for it if he sees fit.

There is nothing particularly interesting in the OperatingSystem struct at the moment, so I see no need to complicate the class with a serialization function.

X9VoiD:
Pushed commits to this PR. It looks like this now on Linux:

```json
  "OperatingSystem": {
    "Name": "Fedora Linux",
    "Version": "38 (Workstation Edition)",
    "Attributes": {
      "ID": "fedora",
      "VERSION_ID": "38",
      "VERSION_CODENAME": "",
      "PRETTY_NAME": "Fedora Linux 38 (Workstation Edition)",
      "SUPPORT_END": "2024-05-14",
      "VARIANT": "Workstation Edition",
      "VARIANT_ID": "workstation",
      "KERNEL_VERSION": "6.4.12-200.fc38.x86_64"
    }
  },
```

and this on Windows:

```json
  "OperatingSystem": {
    "Name": "Win32NT",
    "Version": "10.0.22621.0",
    "Attributes": {
      "ServicePack": ""
    }
  },
```

PR #2950, opened 2023-10-02
Console: Check for presence of Daemon before attempting to connect

### LINKED ISSUE #2898: Console application (otd) should notify of daemon connection early on launch
### Description

There is confusion among users in the move from `otd` to `otd-daemon` - we should probably have some sort of message notifying the user that `otd-daemon` is the way to start the daemon.

---

Current behavior is to do nothing but time out after 5 seconds, in which case an exception stack trace is thrown:
```
$ otd
Unhandled exception. System.TimeoutException: Connecting to daemon failed after 5 seconds.
   at OpenTabletDriver.Desktop.RPC.RpcClient`1.Connect() in /home/gonx/.cache/paru/clone/opentabletdriver-git/src/OpenTabletDriver/OpenTabletDriver.Desktop/RPC/RpcClient.cs:line 35
   at OpenTabletDriver.Console.ApplicationLifetime.Run(String[] args) in /home/gonx/.cache/paru/clone/opentabletdriver-git/src/OpenTabletDriver/OpenTabletDriver.Console/ApplicationLifetime.cs:line 24
   at OpenTabletDriver.Console.Program.Main(String[] args) in /home/gonx/.cache/paru/clone/opentabletdriver-git/src/OpenTabletDriver/OpenTabletDriver.Console/Program.cs:line 11
   at OpenTabletDriver.Console.Program.<Main>(String[] args)
/usr/bin/otd: line 3: 17316 Aborted                 (core dumped) /usr/lib/opentabletdriver/OpenTabletDriver.Console "$@"
```

---

Proposed behavior would be:
Initial message:
```
$ otd
Connecting to the OpenTabletDriver daemon... (start 'otd-daemon' if you haven't)
```
Once connection completes:
```
$ otd
Connecting to the OpenTabletDriver daemon... (start 'otd-daemon' if you haven't)
Connected!
<default output text of `otd`>
```

### Acknowledgements

gonX:
Fixes #2898

### DIFF
--- a/OpenTabletDriver.Console/ApplicationLifetime.cs
+++ b/OpenTabletDriver.Console/ApplicationLifetime.cs
@@ -22,7 +22,10 @@ namespace OpenTabletDriver.Console
         public async Task Run(string[] args)
         {
             if (!Instance.Exists("OpenTabletDriver.Daemon"))
-                throw new ("Daemon not running");
+            {
+                System.Console.WriteLine("OpenTabletDriver Daemon not running");
+                Environment.Exit(1);
+            }
 
             await _rpcClient.Connect();
 
--- a/OpenTabletDriver.Console/ApplicationLifetime.cs
+++ b/OpenTabletDriver.Console/ApplicationLifetime.cs
@@ -21,6 +21,8 @@ namespace OpenTabletDriver.Console
 
         public async Task Run(string[] args)
         {
+            System.Console.WriteLine("Connecting to the OpenTabletDriver daemon...");
+
             await _rpcClient.Connect();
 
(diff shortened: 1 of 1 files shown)


X9VoiD (changes requested):
It's probably better to instead have a timeout or check for existence of the global mutex to not break use of OTD.CLI in a shell.

gonX:
Makes sense, maybe outputting to `STDERR` is better? Is that a thing on Windows?

X9VoiD:
It'll still be too noisy, every call to otd cli will cause that to be printed out to stderr.

PR #2970, opened 2023-10-20
Add acknowledgement for using the latest version

jamesbt365:


jamesbt365:
Maybe it will be better to not force requirement on support requests because maybe genuine reasons exist to use older versions?

gonX:
The checkboxes aren't _required_ per se, it just allows us to see at a glance whether the user followed the guide correctly. I've previously not filled out all the boxes when it wasn't relevant e.g. #2972.

E.g. this = LGTM

gonX (approved):

PR #3001, opened 2023-11-15
Remove --settle and unnecessary second trigger

X9VoiD:
Fixes hanging issues on some misbehaving Linux systems.


gonX (commented):
Like mentioned on Discord, my stance on this is that we should still try to settle (e.g. timeout: `udevadm trigger && udevadm settle --timeout=15`). Maybe look at what other projects like sc-controller / openrazer etc does?

PR #3024, opened 2023-11-22
Tests: Add check for 0.6.x paths in configurations

gonX:
This is to avoid old syntax from 0.6.x sneaking in to master.


gonX:
This will be superseded in an upcoming PR by simply not allowing unknown properties.

PR #3025, opened 2023-11-22
Fix configs that still use 0.6.x syntax

gonX:
Fortunately it seems like the worst that we've been missing are just the buttons on the pen. The one missing aux identifier was not really an issue because it was already blank to begin with.

This fixes failing tests in #3024 and #3027

Does not need a configuration backport for 0.6.x


gonX:
Will be included in #3027 instead

PR #3026, opened 2023-11-22
Tests: Add Tablet config template matching

gonX:
This helps ensure that new configurations are correctly defined and aren't unintentionally missing fields (and therefore features)

This generates the JSONPaths expected in our configs by recursively iterating through the `TabletConfigurations` class

Consider this a discount version of JSON Schema validation - we do not validate types nor values here, just that the paths (keys) are declared.

Json.Net Schema exists and it does not seem difficult for us to add (you can generate the schema in code), so it is likely a cleaner (and more performant) alternative to this down the line. Regardless, this does still help future config PRs for now.

### DIFF
--- a/OpenTabletDriver.Tests/ConfigurationTest.cs
+++ b/OpenTabletDriver.Tests/ConfigurationTest.cs
@@ -1,9 +1,12 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


jamesbt365:
Couldn't we just put in a sample configuration and compare against that?

gonX:
> Couldn't we just put in a sample configuration and compare against that?

It was actually initially proposed like that but if we end up changing the code, we need to also be mindful of changing that file. This approach links it tighter to the code so that a simple variable renaming doesn't cascade fail the tests.

As the Json.NET Schema line in OP alludes to, that is the even better approach, since it allows us to basically just give the schema generator our `TabletConfiguration` class, and we magically get a JSON Schema back that can validated against the configuration files.

gonX:
Closing as this will be superseded in an upcoming PR with a much simpler approach

PR #3049, opened 2023-11-26
Lint configuration json files

### LINKED ISSUE #1689: Add configuration linter
## Description

A configuration linter would improve the configuration workflow to avoid cases like [this](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/1679#pullrequestreview-822969719) so that all configuration changes would have to clean in order to be merged.

Suggestions are welcome

gonX:
Fixes #1689

Features:

- Flattens trailing decimal point on the following Digitizer fields:
  - Width
  - Height
  - MaxX
  - MaxY
- Checks for end-of-line markers
- Checks for end-of-line at end-of-file
- Ensures 2-character indentation with spaces
- Probably some other "undocumented" features, since it checks strictly whether the external configuration file matches the result of doing deserialization -> serialization of the same json file.

A commit is included to make the new tests pass


OpenTabletDriver/Tablet/DigitizerSpecifications.cs
```diff
@@ -14,26 +17,65 @@ public class DigitizerSpecifications
...
         /// </summary>
         [Required(ErrorMessage = $"Digitizer ${nameof(MaxY)} must be defined")]
+        [JsonConverter(typeof(DecimalJsonConverter))]
         [DisplayName("Max Y")]
         public float MaxY { set; get; }
     }
+
+    public class DecimalJsonConverter : JsonConverter
+    {
+        public override bool CanRead => false;
```
X9VoiD, line 49: `Newtonsoft.Json` uses a fallback when the converter doesn't support reading?

jamesbt365 (commented):
Don't see anything immediately wrong with any of the configuration files, haven't sat down to check the code for anything funky yet.

OpenTabletDriver/Tablet/DigitizerSpecifications.cs
gonX: Yes, it uses their default converter if nothing else is defined.

My initial plan was just to handle all floats generically with this class, so it could be left inside the tests project, but due to (from what I can tell are) library limitations, I could either pick between that or being able to set indentation and then having to specify the converter on a per-object basis.

It's a longer explanation but there are multiple ways to serialize an object to JSON in `Newtonsoft.Json`, so it is likely I missed a better way to do this.

PR #3051, opened 2023-11-26
MacOS: Set mouse delta to move and drag event

### LINKED ISSUE #3048: Support drag to pan the scene around in Godot, Unity
### Description

I'm facing an issue when using my XP-Pen (Deco 01 v2) to work with Godot where I can not pan the scene by drag the pen around on MacOS Sonoma. This also happens in Unity.

Other applications like Krita, Aseprite, Gimp... works fine.

I can pan with my pen when using XP-Pen's stock driver. But the stock driver is unstable on my machine so it would be nice if I can use OTD instead. Thanks!

### Acknowledgements

chutneyio:
This PR fixes the issue where pen drag to pan does not work in some apps which requires mouse delta.
Fixes #3048

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/MacOSAbsolutePointer.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/MacOSAbsolutePointer.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Numerics;
 using OpenTabletDriver.Native.MacOS;
+using OpenTabletDriver.Native.MacOS.Input;
 using OpenTabletDriver.Platform.Display;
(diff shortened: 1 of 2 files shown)


X9VoiD (changes requested):
Test if this works on both absolute and relative output modes. This looks redundant to what this does:

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/5218f4313a34e26ef79671cbf50f58adec25ffee/OpenTabletDriver.Desktop/Interop/Input/Relative/MacOSRelativePointer.cs#L28-L33

And will likely break relative output mode. Probably only change `OpenTabletDriver.Desktop/Interop/Input/Absolute/MacOSAbsolutePointer.cs`

X9VoiD (approved):

PR #3072, opened 2023-12-03
Add extra string for Kamvas Pro 13 (2.5k)

### LINKED ISSUE #3069: Huion KAMVAS Pro 13 (2.5K) Missing Device String
### Steps to Reproduce

1. Connect Huion KAMVAS Pro 13 (2.5K) to computer
2. Wait for device to be detected
3. It never connects, it says it is missing devices.

### Expected Behavior

1. Connect Huion KAMVAS Pro 13 (2.5K) to computer
2. Wait for device to be detected
3. Device connects and works correctly.

### Observed Behavior

1. Connect Huion KAMVAS Pro 13 (2.5K) to computer
2. Wait for device to be detected
3. It never connects, it says it is missing devices.

### Tablet Device

Huion KAMVAS Pro 13 (2.5K)

### Diagnostic Information

The debug information just listed trying to detect my device against all the configurations, but I figured that was not worth putting here because I have already worked it out. I just don't know the way that your workflow works to commit the changes.

I changed [DeviceString in Kamvas Pro 13 (2.5k).json](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/da80f70893c01d407e4fa0633c4ea9508c77d5e4/OpenTabletDriver.Configurations/Configurations/Huion/Kamvas%20Pro%2013%20(2.5k).json#L30C15-L30C15) to `HUION_M213_\\d{6}$` instead (changing a single character 0→3) and it detected the device correctly. I also tried copying the existing block in `DigitizerIdentifiers` and changing only one of them, and it worked correctly after I built and ran it.

I found this out by placing a breakpoint at `OpenTabletDriver/Driver.cs:93` on the correct device and seeing where the logic didn't match up.

There are a few other problems I have currently, I will list them here, but they might belong better in separate issued, just tell me if that is the case, and I'll move them and include any diagnostic info that is relevant.

1. The screen output selector needs to be set to the size that encompasses both screens I have plugged in (4480x1600), which is the height of the tablet, but the witch of the tablet + my 1080p monitor. This is because if I have it set to only the size of one screen, then it squishes the tablet area so it only covers one half of one screen. This might be a wayland problem, but I don't see any other issues for it, so it might be specific to this device.
2. The tilt of the pen goes the opposite way it is meant to in Krita, I have not tried any other software, but there is no option to reverse it in there anyway, so I figured it was a driver problem.

System Information:
OS: Arch linux
Version: 0.6.3.0-3, git: da80f70893c01d407e4fa0633c4ea9508c77d5e4 manually built and run. The git version is what I used to modify it, but the bugs were on both.
Desktop manager: KDE Plasma, Wayland

### Acknowledgements

J05HM0N5TER:
My device was not detected using the existing string, this adds the one that my device has.

Fixes #3069

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/Kamvas Pro 13 (2.5k).json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/Kamvas Pro 13 (2.5k).json
@@ -27,7 +27,7 @@
       "FeatureInitReport": null,
       "OutputInitReport": null,
       "DeviceStrings": {
-        "201": "HUION_M210_\\d{6}$"
+        "201": "HUION_M(210|213)_\\d{6}$"
       },
       "InitializationStrings": [
         200


gonX:
We prefer that you change it to the suggestion by jamesbt365

jamesbt365:
you can improve that further by leaving the M outside the brackets and just putting the numbers within it.

J05HM0N5TER:
> you can improve that further by leaving the M outside the brackets and just putting the numbers within it.

Do you think it would be better to have `HUION_M21(0|3)_\\d{6}$`, or would it be unreadable at that point?

jamesbt365:
I think that either one of my suggestions is more readable, I think taking it any further makes it not as clear imo.

J05HM0N5TER:
Updated it. Figured I would just keep amending the existing commit, so I don't pollute the commit history. It mentioned so in your commit guidelines.

J05HM0N5TER:
Is there anything else that I need to do before this can be merged?

gonX (approved):

PR #3073, opened 2023-12-03
merge matching identifiers with different strings

jamesbt365:
I hope I didn't accidentally unsupport some tablets! Only targets the Huion directory.

One Huion H1060P variant gains tilt support, this is untested by most likely fine to do. I can do this to the other remaining H1060P identifiers too if this is desirable.


jamesbt365:
Will backport when I go through the rest of the directories and am sure I didn't break anything.

PR #3151, opened 2023-12-27
Create README_TW.md

stonkol:
This is the Traditional Chinese version of the README file


jamesbt365:
You need to attribute your README on every other README so that users can click to and from any README.

stonkol:
> You need to attribute your README on every other README so that users can click to and from any README.

Ok, I will do that.

gonX:
For another time, you can simply add (and then push) commits to your branch to add the requested changes. You don't need to create a new PR to modify your changes.

(new PR at #3152)

PR #3168, opened 2024-01-04
Update templates diagnostic step

jamesbt365:
OpenTabletDriver/opentabletdriver.github.io#152 should be fixed before this is merged, if the link changes that should also be updated on this pull request.

With the update to the logging directory users can now retrieve logs from a crash easier.

GitHub allows uploading of json files without renaming them to txt now.

I have not tested these templates updates yet and I'm bad at English so wording improvements are always welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -52,13 +52,12 @@ body:
       label: Diagnostic Information
       description: |
         Export diagnostics through the OpenTabletDriver GUI from menu <kbd>Help</kbd> → <kbd>Export Diagnostics</kbd>.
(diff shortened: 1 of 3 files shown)


jamesbt365:
One small issue
image
image


No idea how to prevent the newline without making the line stupidly long in the text editor.

gonX:
> One small issue image image
> 
> No idea how to prevent the newline without making the line stupidly long in the text editor.

You can likely put backslashes at the end of the line to prevent a line wrap

InfinityGhost (dismissed):

jamesbt365:
Decided to just make the line longer in the editor, it just gives a better final result for the user. Not sure how to get the newlines to behave here, escaping them doesn't work and I don't feel like going through yml spec until i do figure it out.

InfinityGhost (approved):

PR #3175, opened 2024-01-07
Backport CTH-300 config

Kuuuube:
Backport  #3174


jamesbt365:
Doesn't include the Tablets.md change but I have included it in #3372.

PR #3226, opened 2024-03-04
Use basic TabletReportParser on Huion 420

jamesbt365:
The Huion 420 has no need for using the UCLogicReportParser as its only uses that for auxiliary keys, as this tablet has none it should *probably* not be using this in the first place.

The main reason behind doing this is because when pressing the pen buttons it can potentially trigger auxiliary keys that don't exist.

image

Not sure if this is the most optimal solution to begin with but this tablet has no aux functionality and thus doesn't need this parser.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/420.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/420.json
@@ -21,7 +21,7 @@
       "ProductID": 110,
       "InputReportLength": 8,
       "OutputReportLength": null,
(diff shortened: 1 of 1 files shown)


gonX (approved):
The `UCLogicReportParser` does use `TabletReportParser` (well, `TabletReport`) when a report isn't detected as aux, so I think this is a good bet:
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/2d7d22905fd79806a0f8c73336b954cb991360a6/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicReportParser.cs#L7-L13

PR #3228, opened 2024-03-08
Add support for XP-Pen Gen2

Kuuuube:


jamesbt365:
You forgot the TABLETS.md

chenyanming:
Nice PR! I am using your PR to support XP-Pen Artist Pro 16 (Gen2), however, I found the X and Y positions would be more accuate if it takes tilt into account. I have successfully added the following change locally and the X/Y is more accuate now. Would you consider to add it?

@Kuuuube https://github.com/Kuuuube/OpenTabletDriver/pull/2

OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenTabletGen2Report.cs
```diff
@@ -0,0 +1,41 @@
...
+        public XP_PenTabletGen2Report(byte[] report)
+        {
+            Raw = report;
+
+            Position = new Vector2
+            {
+                X = Unsafe.ReadUnaligned<ushort>(ref report[2]) | report[10] << 16,
+                Y = Unsafe.ReadUnaligned<ushort>(ref report[4]) | report[11] << 16
+            };
+            Pressure = (uint)((Unsafe.ReadUnaligned<ushort>(ref report[6]) - 16384) | (report[13] & 0x01) << 13);
```
X9VoiD, line 18: Is `Pressure` really read like this? It kinda feels dodgy to me.

X9VoiD: Maybe it's supposed to be:

```suggestion
            Pressure = (uint)((Unsafe.ReadUnaligned<ushort>(ref report[6]) & 0x1FFF) | (report[13] & 0x01) << 13);
```

Kuuuube: It's essentially the same thing we do in XP_PenPressureOffsetTiltTabletReport but with an extra bit so it's doubled.

The tablet reports this when there is no pressure:
```
00 40
```
With 1 pressure it reports this:
```
01 40
```
When it gets to 4096 (or 4097?) it reports:
```
01 50
```
With max pressure it reports this:
```
FF 5F
```

The `40` in the second byte is just left there always. We could remove this bit differently but `- 16384` does the job.

Data capture: 
[tablet-data.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/15369388/tablet-data.txt)

X9VoiD: Looks right then. OTD really should start documenting these things so we won't lose track of how stuff came to be.

X9VoiD (approved):

PR #3229, opened 2024-03-08
Backport XP-Pen Gen2 support

Kuuuube:

PR #3233, opened 2024-03-09
Backport support for deco pro lw gen2

Kuuuube:
Depends on #3229


Kuuuube:
Combining backport for xp-pen gen2 support into #3229 to avoid making things more messy than needed.

Issue #3011, opened 2023-11-17
Change detection scheme


X9VoiD:
Using just the input/output/feature report length is not specific enough for devices that have all three looking similar.

Instead we should use HID report ID and usage ID.

Something like this in configurations will be enough to cover all possible variations of a tablet and still not be confused by report lengths.

```json
  "Attributes": {
    "HID_REPORT_ID": "0x07",
    "HID_USAGE_PAGE": "0xff0a",
    "HID_USAGE": "0x01",
    "USB_INTERFACE": "1",
  }
```

X9VoiD:
#3245 when merged provides the base for finally implementing this.

InfinityGhost:
> [#3245](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/3245) when merged provides the base for finally implementing this.

Issue wasn't linked, closing as it is merged.

jamesbt365:
This is not complete, the only part that is complete is matching via interface.

Kuuuube:
Closing this as it has been implemented mostly with #4836. Could be made a bit nicer but it is there.

PR #3238, opened 2024-03-11
Fix UC-Logic 1060N configuration

### LINKED ISSUE #3237: XP-Pen Star 03 / UC-Logic 1060N variant not detected
### Steps to Reproduce

Since linux Mint does not allow the configuration of multiple tablet profiles I installed OpenTabletDriver.

1. I installed your *.deb package
2. after rebooting I connected my tablet (Xp-Pen Star 03) - and yes, my model is supported on your device list
3. I launched OpenTabletDriver from the applications panel in Mint
4. OpenTabletDriver tells me: "Connecting to OpenTabletDriver deamon" but aborts after a timeout
5. Now I checked Mint's System settings > tablet and surprise! now, Mint is unable to see my tablet 

Before the installation of OpenTabletDriver my tablet was working fine and Mint detected my tablet and pen immediately, in addition, every time I connect my tablet, my mouse cursor disappears. 

### Expected Behavior

A working driver that doesn't mess with my mouse and gives me a way to setup multiple tablet profiles.

### Observed Behavior

I didn't do anything wrong with your installer - the installation of OpenTabletDriver is fool proof.
All I know is that OpenTabletDriver messed up my default tablet settings (deactivating wacom drivers) as described in your wiki:
```
Currently, two kernel modules are also unloaded as they interfere with OpenTabletDriver. These are:

wacom
hid_uclogic
```
and deactivating my damn cursor, which is even more upsetting!

Then I tried to install yet another time in the hope to get a "complete installation" of your driver. Nothing changed after rebooting.
Guys, you probably you need to check better how you write your installers, because I don't use a small insignificant distro with all kind of issues, so the fault is clearly on your side. Sorry to come up to ask this, but did you actually do some little check if your installer worked fine on major distros like Mint, **at least** ?

Since this installation is such a birth, I just ask you to tell me how to reactivate my wacom drivers so mint can see again my tablet and the uninstaller for OpenTabletDriver. 

### Tablet Device

XP-Pen Star 03

### Diagnostic Information

`lsusb` yes I can see my tablet:
`Bus 001 Device 015: ID 5543:0081 UC-Logic Technology Corp. TABLET 1060N`

there is nothing related to OpenTabletDriver in these folders: 
```
/dev/uinput #uinput does not exist
~/.config/
/etc/udev/rules.d/
~/.config/OpenTabletDriver 
```

here's the log from `~/.config/OpenTabletDriver`
```
[
{
  "Time": "2024-03-11T11:26:53.4500529+01:00",
  "Group": "Detect",
  "Message": "Searching for tablets...",
  "StackTrace": null,
  "Level": 1,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4672761+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet '10moon 1060N'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4679634+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Acepen AP 1060'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4680586+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Adesso Cybertablet K8'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4680782+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Artisul A1201'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4680831+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Artisul AP604'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4680927+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Artisul M0610 Pro'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4804868+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'FlooGoo FMA100'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.480502+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Gaomon 1060 Pro'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4805117+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Gaomon GM116HD'",
  "StackTrace": null,
  "Level": 0,
  "Notification": false
},
{
  "Time": "2024-03-11T11:26:53.4805202+01:00",
  "Group": "Detect",
  "Message": "Searching for tablet 'Gaomon GM156HD'",
  "StackTrace": null,
  "Level": 0,
  "Notification": fal
```

### Acknowledgements

jamesbt365:
Fixes #3237.

I don't think we had concrete evidence of this configuration working to begin with.

No verification was ever done on pressure and considering the age of this tablet I'm guessing it was plain wrong to begin with, so I've corrected the value to 2047. See #1037 for the original implementation.

Somewhere between the original pull request mentioned above and now the `OutputReportLength` was set to null, which is wrong because multiple interfaces have the `InputReportLength` of 8, which at a minimum breaks support for Linux and possibly MacOS.

Edit: looking into this further, this reverts changes from #1721, but without reverting this it introduces RNG into weather or not the tablet wants to work, but no logs were provided there, so I'm not sure what the actual approach should be here.

Edit2: looking into this further, they are forcing their tablet to use interface 0, which I don't believe is right? I've seen 2 tablets using the exact same configuration with the exact same device strings, unless this is a different variant i don't see why its using interface 0 and instead of interface 2, and why it needs null instead of 8.

Regardless using null is not always a brilliant idea....

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/UC-Logic/1060N.json
+++ b/OpenTabletDriver.Configurations/Configurations/UC-Logic/1060N.json
@@ -8,7 +8,7 @@
       "MaxY": 30480
     },
     "Pen": {
-      "MaxPressure": 8191,
+      "MaxPressure": 2047,
       "ButtonCount": 2
     },
     "AuxiliaryButtons": {
@@ -20,6 +20,7 @@
       "VendorID": 21827,
       "ProductID": 129,
       "InputReportLength": 8,
+      "OutputReportLength": 8,
       "ReportParser": "OpenTabletDriver.Configurations.Parsers.XP_Pen.XP_PenReportParser",
       "OutputInitReport": [
         "ArAC"
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -144,7 +144,7 @@
 | XP-Pen Deco mini7             |     Supported     |
 | XP-Pen Deco Pro LW Gen2       |     Supported     |
 | XP-Pen Deco Pro XLW Gen2      |     Supported     |
-| XP-Pen Star 03                |     Supported     |
+| XP-Pen Star 03                |     Supported     | Older variants may use the same configuration as the UC-Logic 1060N.
 | XP-Pen Star 03 Pro            |     Supported     |
 | XP-Pen Star 05 V3             |     Supported     |
 | XP-Pen Star G430S             |     Supported     |


jamesbt365:
The PR body is a mess at this point so to clarify what this actually does.

- It fixes the max pressure of the tablet, it previously was never verified.
- It has to remove support for a variant of this tablet with broken firmware, support currently cannot co-exist without changes to the detection scheme that are not mentioned in #3011 by only matching specific configurations when others fail.
- Auxiliary buttons may or not work consistently, I will note this on the TABLETS.md if its the case but it has not been mentioned with 2/3 users with this tablet (one has bugged firmware the other is from multiple years ago), so its impossible to say for sure if the 3rd user has broken hardware or its on our side.

I will rebase and undraft when I get clarification on if this change is okay.

Kuuuube (approved):

gonX:
Merged the 0.6.x one, so merging this as well.

PR #3245, opened 2024-03-12
Attribute overhaul

### LINKED ISSUE #2438: support for feature initialization delay lost after DI
### Acknowledgements

### Steps to Reproduce

Open up a configuration, add the attribute FeatureInitDelayMs and invoke detection, inits are not delayed.

### Expected Behavior

Be able to add feature init delay in tablet configurations

### Observed Behavior

This cannot be done because the code is gone.

### Tablet Device

N/A

### Diagnostic Information

N/A

jamesbt365:
Supersedes #3243.
Fixes #2438.

TODO:
- [x] Add attribute to force specific interface on Linux & unify attributes to "Interface".
- [x] Restore `FeatureInitDelayMs` attribute.
- [x] Add support for specific identifiers via `DigitizerIdentifiers`.
- [x] Document changes in the codebase.
- [x] Fix tests.

I have something I'm not the most sure on right now, should a method be implemented somewhere to get the combined attributes? Right now it has absolutely no demand and its not complex logic to begin with, and use for combined attributes only are required in one spot, so its not like any code duplication will happen if we don't implement one.

At this time I'm not going to add support for FeatureInitDelayMs or libinputoverride in specific device identifiers, this is not needed and just increases complexity, but implementing the above would amend that. But considering the usage of delay (only intuos3 and on ALL tablets regardless of identifier) there is absolutely no need to implement support into the identifiers too.

Reminder to submit a pull request for web once this is merged.

### DIFF
--- a/OpenTabletDriver.Analyzers.Tests/Verifiers/FileBasedAnalyzerConfigOptionsProvider.cs
+++ b/OpenTabletDriver.Analyzers.Tests/Verifiers/FileBasedAnalyzerConfigOptionsProvider.cs
@@ -103,7 +103,7 @@ namespace OpenTabletDriver.Analyzers.Tests.Verifiers
             return options;
         }
 
-        [GeneratedRegex(@"^(?<key>(?:\w|_|-|\.)+) = (?<value>(?:\w|_|-|\.)+)$", RegexOptions.Multiline | RegexOptions.NonBacktracking)]
+        [GeneratedRegex(@"^(?<key>(?:\w|_|-|\.)+) = (?<value>(?:\w|_|-|\.)+)\s*$", RegexOptions.Multiline | RegexOptions.NonBacktracking)]
         private static partial Regex ValuePair();
 
         private class CustomAnalyzerConfigOptions : AnalyzerConfigOptions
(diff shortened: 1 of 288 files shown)


gonX:
Probably a bit late that I thought of this, but wouldn't the attribute better be defined as a new property in the digitizer identifier?

jamesbt365:
> Probably a bit late that I thought of this, but wouldn't the attribute better be defined as a new property in the digitizer identifier?

This probably isn't the best case considering improving the detection scheme further would add extra attributes, putting extra filters in place. See #3011 for what else could be added in the future.

jamesbt365:
#3276 requires Interface 0, there are 3 interfaces all with the same input and output lengths.

jamesbt365:
Will rebase this PR and backport it when we close in on the next release and merge some configurations, its not fair on others to rebase 70+ configuration pull requests because i changed the configuration format.

jamesbt365:
I'm not actually sure if using the existing code is the best idea, looking at some of the comments internally i'm not sure the information is available for every identifier (whereas it is always in the device path) I know @X9VoiD implemented this stuff so I'd like his opinion on it.

jamesbt365:
Will revert the latest commit and put a note in to use existing stuff when its reliable with WinUSB and anything else we add support for later. Then I'll rebase and fix conflicts.

jamesbt365:
The remaining elements from #3011 are out of scope in this PR and will be handled later so that should be everything complete now (except a rebase) if I'm correct @X9VoiD?

X9VoiD:
We still need to not require the new Attribute field on all configs.

X9VoiD (approved):

PR #3206, opened 2024-02-10
Add UGEE S1060

kip93:
Adds support for the UGEE S1060. Had to add 2 extra aux buttons since this tablet has 12 but the code only parsed 10 of them.

Related issue: #3053

<details><summary>diagnostics</summary>

```json
{
  "App Version": "OpenTabletDriver v0.6.4.0",
  "Build Date": "02/10/2024",
  "Operating System": {
    "Name": "NixOS",
    "Version": "24.05 (Uakari)",
    "Attributes": {
      "BUILD_ID": "24.05.20240207.f8e2ebd",
      "ID": "nixos",
      "PRETTY_NAME": "NixOS 24.05 (Uakari)",
      "VERSION_CODENAME": "uakari",
      "VERSION_ID": "24.05",
      "KERNEL_VERSION": "6.1.77"
    }
  },
  "Environment Variables": {
    "USER": "kip",
    "TEMP": null,
    "TMP": null,
    "TMPDIR": null,
    "DISPLAY": ":0",
    "WAYLAND_DISPLAY": "wayland-0",
    "PWD": "/home/kip",
    "PATH": "/nix/store/hsssk3gp60r9xxaq2kpgp3nd10zilidq-dotnet-runtime-6.0.26/bin:/run/wrappers/bin:/home/kip/.nix-profile/bin:/nix/profile/bin:/home/kip/.local/state/nix/profile/bin:/etc/profiles/per-user/kip/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin"
  },
  "HID Devices": [
    {
      "DevicePath": "/sys/devices/pci0000:00/0000:00:14.0/usb1/1-4/1-4:1.1/0003:28BD:0938.0002/hidraw/hidraw1",
      "Manufacturer": "UGTABLET",
      "ProductName": "UGEE S1060",
      "SerialNumber": "000000",
      "FriendlyName": "UGEE S1060",
      "VendorID": 10429,
      "ProductID": 2360,
      "InputReportLength": 10,
      "OutputReportLength": 0,
      "FeatureReportLength": 0,
      "CanOpen": true,
      "DeviceAttributes": {
        "USB_INTERFACE_NUMBER": "1",
        "HID_REPORTS": "07:000D:0002"
      }
    },
    {
      "DevicePath": "/sys/devices/pci0000:00/0000:00:14.0/usb1/1-4/1-4:1.2/0003:28BD:0938.0003/hidraw/hidraw2",
      "Manufacturer": "UGTABLET",
[...]

Issue #3053, opened 2023-11-26
Support Request: for UGEE S1060 and S1060W


xKRISTOFx:
### Description

Those are analog tablets to UGEE S640. They use one driver from producer's page, only difference is that they are bigger and have 2 more action buttons. There also is version S640W, which as with bigger brother is wireless analog. I'm in need using Open source driver because producers driver doesn't support 2 tablet setup and I have 32:9 monitor, so I need two.

### Acknowledgements

- [X] I am on the latest version of OpenTabletDriver.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

jamesbt365:
We cannot support tablets without the relevant information. Please open another issue using the **configurations** template, not the enhancement template, or post the relevant information on that template here and I'll reopen this issue.

PR #3258, opened 2024-03-20
Adding UGEE S1060 and S1060W and Battery Report

Spencer-Sawyer:
Added support for the UGEE S1060W and S1060 as well as added a battery report which currently does nothing as I didn't see an easy way to tie it to anything useful. Still provides more information when debugging the tablet. 
Diag showing the tablet connects in both modes:
[Diag.json](https://github.com/OpenTabletDriver/OpenTabletDriver/files/14668027/Diag.json)


Spencer-Sawyer:
related PR: #3206 This is intended to supersede that PR as this config should be more complete.
Related issue #3053

jamesbt365:
I'm not sure we should have the S1060 and S1060W variants under the same configuration, if they are distinctly sold under different names, they should reflect so in OTD, this is how we do things with other configurations so it should be done here too.

TABLETS.md
```diff
@@ -1,5 +1,5 @@
 | Tablet                        |      Status       | Notes |
-| ---------------------------   | :---------------: | ----- |
+|-------------------------------| :---------------: | ----- |
```
jamesbt365, line 2: This shouldn't break anything but for what purpose?

Spencer-Sawyer: My bad, I thought I had removed all the weird things that Rider did.

TABLETS.md
```diff
@@ -63,6 +63,7 @@
 | UGEE M808                     |     Supported     |
 | UGEE S640                     |     Supported     |
 | UGEE S1060                    |     Supported     |
+| UGEE S1060W                    |     Supported     |
 | UGEE U1200                    |     Supported     |
```
InfinityGhost, line 67:
```suggestion
| UGEE S1060                    |     Supported     |
| UGEE S1060W                   |     Supported     |
| UGEE U1200                    |     Supported     |
```

OpenTabletDriver/Tablet/IBatteryReport.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    /// A report containing battery information.
+    /// </summary>
+    [PublicAPI]
+    public interface IBatteryReport : IDeviceReport
+    {
+        /// <summary>
+        /// Charge percentage of device
+        /// </summary>
+        uint ChargePercent { set; get; }
```
InfinityGhost, line 14: What is the range here? That should be documented.

OpenTabletDriver/Tablet/IBatteryReport.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <summary>
+        /// Whether or not the device is plugged in
+        /// </summary>
+        bool PluggedIn { set; get; }
```
InfinityGhost, line 18: Maybe this?
```suggestion
        bool Charging { set; get; }
```

InfinityGhost (changes requested):

PR #3262, opened 2024-03-22
Veikk A50 revisit

Spencer-Sawyer:
Added support for aux buttons to one of the existing Veikk parsers and shifted the A50 to use it instead. 

Also corrected a confusing naming convention. The V2 parser wasn't actually being used for any V2 tablets. I relabel it V1 because I think it's an older protocol.

I suspect that the other "skip byte tablets" that remain can also be migrated to this parser, but I don't want to touch them without confirmation that it would work correctly.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxV1Report.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxV1Report.cs
@@ -1,0 +1,32 @@
+using OpenTabletDriver.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Veikk
+{
+    public struct VeikkAuxV1Report : IAuxReport
+    {
+        public VeikkAuxV1Report(byte[] report)
+        {
+            Raw = report;
+            AuxButtons = new bool[]
+            {
+                // Buttons 4 and 6 could shadow each other in the case of more than one buttons being pressed at once.
+                // Button 6 would take precedence.
+                System.Array.IndexOf(report, (byte)0x3E) > 1,
+                System.Array.IndexOf(report, (byte)0x0C) > 1,
+                System.Array.IndexOf(report, (byte)0x2C) > 1,
+                System.Array.IndexOf(report, (byte)0x19) > 1 && report[1] == 0x00,
+                System.Array.IndexOf(report, (byte)0x06) > 1,
+                System.Array.IndexOf(report, (byte)0x19) > 1 && report[1] == 0x01,
+                System.Array.IndexOf(report, (byte)0x1D) > 1,
+                System.Array.IndexOf(report, (byte)0x16) > 1,
+                System.Array.IndexOf(report, (byte)0x2E) > 1,
+                System.Array.IndexOf(report, (byte)0x2D) > 1,
+                System.Array.IndexOf(report, (byte)0x30) > 1,
+                System.Array.IndexOf(report, (byte)0x2F) > 1,
+            };
+        }
+
+        public byte[] Raw { get; set; }
+        public bool[] AuxButtons { get; set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxV1Report.cs
```diff
@@ -0,0 +1,32 @@
...
+                System.Array.IndexOf(report, (byte)0x2C) > 1,
+                System.Array.IndexOf(report, (byte)0x19) > 1 && report[1] == 0x00,
+                System.Array.IndexOf(report, (byte)0x06) > 1,
+                System.Array.IndexOf(report, (byte)0x19) > 1 && report[1] == 0x01,
+                System.Array.IndexOf(report, (byte)0x1D) > 1,
+                System.Array.IndexOf(report, (byte)0x16) > 1,
+                System.Array.IndexOf(report, (byte)0x2E) > 1,
+                System.Array.IndexOf(report, (byte)0x2D) > 1,
+                System.Array.IndexOf(report, (byte)0x30) > 1,
+                System.Array.IndexOf(report, (byte)0x2F) > 1,
```
X9VoiD, line 25: Can I see what the raw data looks like? This doesn't seem like an efficient way of doing this, all of these are O(n) operations that could most likely be O(1) when done properly.

Probably do a screen record of tablet debugger or use its built-in tablet data recording tool. Record pressing these buttons one by one. [See here](https://opentabletdriver.net/Wiki/FAQ/General#appdata) where to find the `tablet-data.txt` if you used the built-in recorder.

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkV1ReportParser.cs
```diff
@@ -2,13 +2,16 @@
...
     {
         public IDeviceReport Parse(byte[] report)
         {
+            if (report[0] == 0x03)
+                return new VeikkAuxV1Report(report);
+
             return (report[1], report[2]) switch
             {
-                (0x41, byte x) when (x & 0xF0) == 0xA0 => new TabletReport(report[1..]),
+                (0x41, _ ) when (report[2] & 0xF0) == 0xA0 => new TabletReport(report[1..]),
```
X9VoiD, line 14: Also revert this change. This is done as an easy micro-optimization that the JIT compiler may miss.
```suggestion
                (0x41, byte x) when (x & 0xF0) == 0xA0 => new TabletReport(report[1..]),
```

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkAuxV1Report.cs
Spencer-Sawyer: It reports it as a list of bytes.  I can try to do something that's more efficient, but I didn't think that o(n) was that big of an issue with lists that were never more than 10 elements long.

When I tried to backport it to 0.6.x I can't seem to get it to work correctly and thus can't use the built in recorder from 0.6 is there another recorder that works with the master branch?

Spencer-Sawyer: https://github.com/OpenTabletDriver/OpenTabletDriver/assets/42276949/3e971c4f-594e-4790-a907-0d62aabc09e1

Spencer-Sawyer: Changed to be a single loop over the report that sets all the relevant aux bools.

X9VoiD: Yeah, it isn't much of an issue, but we'd like to save as much cycles as we can. OpenTabletDriver already have measurable negative effects to latency-sensitive games compared to other solutions so we'd like to not further regress that.

The new code is better, JIT can also optimize that switch into a jump table.

PR #3322, opened 2024-05-18
Add support for Trust Flex Design Tablet

jamesbt365:
Verification: https://canary.discord.com/channels/615607687467761684/1240317376101093446/1241474209163645000
Diagnostic: https://canary.discord.com/channels/615607687467761684/1240317376101093446/1241485624800972850

There is a quirk with the tablet itself (not a bug in OpenTabletDriver) that prevents pressure from being reported when pen buttons are pressed.

The lpmm of the tablet is around 80~ the specifications on the internet don't seem to agree on the size of the active area of this tablet, so it was measured by hand.


InfinityGhost (changes requested):
Seems to need a rebase to merge

InfinityGhost (approved):

PR #3408, opened 2024-08-02
Add support for Huion 1060 Plus

jamesbt365:
This is certainly an interesting one. It also seems identical to the Monoprice 10594. Both tablets require WinUSB, it just originally slipped past review on #2337. (the only notable difference is that this one has more auxiliary buttons)

https://discord.com/channels/615607687467761684/992875926654304426/992881131257069608
https://discord.com/channels/615607687467761684/992875926654304426/992880866302906489

I'm not sure all variants will have that string on index 6 for either of those tablets as we have seen some weird behaviour before, but we won't know for sure without tanking the possibly breaking change.

If we don't want to take any risks I can instead remove the string changes and simply add to the TABLETS.md that it'll detect as the monoprice and the last 4 auxiliary buttons won't be supported, but the perks of submitting this as is is that if it is breaking we will get information on a tablet that is otherwise problematic (there has been even more matches of this tablet falsely, so narrowing down false detections will help improve tablet support)

Old string dump is the second link __above__

Verification: https://discord.com/channels/615607687467761684/789348845372178482/1268664084740444303
Diagnostic: https://discord.com/channels/615607687467761684/789348845372178482/1268655991071969290

Can't find reasoning behind it but its already on the TABLETS.md


Kuuuube (dismissed):
Looks fine, if this does break detection for any versions of the monoprice tablet theres also a decent chance we weren't supporting them properly anyways.

Needs conflicts resolved though.

Kuuuube (dismissed):

Kuuuube (approved):

Issue #3403, opened 2024-08-01
OTD's linux native timer implementation constantly triggers ananicy-cpp to re-set rules for it


1Naim:
### Steps to Reproduce

1. Use a plugin that uses Linux timers (I used Hawku in this case)
2. Just move the cursor around
3. Check ananicy-cpp logs in journal (make sure `loglevel=debug` and `log_applied_rule=true`)
```bash
journalctl -b 0 | grep ananicy-cpp | grep Open
```

### Expected Behavior

https://paste.cachyos.org/p/759a856.txt <- This is a patched version of OTD that replaces librt-timer with libc-timerfd+epoll. Timer expiration is not signalled with a creation of a new thread with the latter implementation so OTD doesn't swamp ananicy-cpp with re-setting configs for it each time because of "unnecessary" new threads.

[This](https://github.com/1Naim/CachyOS-PKGBUILDS/blob/feat/opentabletdriver/opentabletdriver/otd-replacelibrt-timer-withlibc-timerfd.patch) is the patch that was provided to me by William Horvath (or better known as @spectator in the osu-linux community) which replaces librt-timer -> libc-timerfd+epoll. I have gotten permission from him to make an issue and share this patch, and I have also got word from him that he will be making a PR soon.

### Observed Behavior

https://paste.cachyos.org/p/759a856.txt <- This is OTD 0.6.4.0 stock with its librt-timer implementation. This is only a snippet of the whole log but basically because a new thread is created each time a timer expires, ananicy-cpp will detect that new thread and set rules for it. This is not ideal especially on older/less powerful CPUs where it will cause high CPU usage and cause performance regressions.

### Tablet Device

Wacom CTL-472

### Diagnostic Information

[otd-export-diag.json](https://github.com/user-attachments/files/16457465/otd-export-diag.json)

### Acknowledgements

- [X] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [X] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.
- [X] I am on the latest version of OpenTabletDriver.

### Additional Information
[...]

PR #3412, opened 2024-08-04
LinuxTimer: replace librt timer_ with libc timerfd_ + epoll

1Naim:
The old implementation signalled each timer expiration by creating new threads, which starves ananicy-cpp of time to keep up with any new processes started after OpenTabletDriver.

This conflict only occurred if there were any filters enabled which required use of the Linux timer, and the new implementation resolves it (along with being more efficient in general).

This change was originally made by @whrvt. I just rebased it onto master and I got his permission to open this PR under my name.

Fixes #3403

PR #3414, opened 2024-08-04
Use libc timerfd instead in LinuxTimer

hwsmm:
- Closes https://github.com/OpenTabletDriver/OpenTabletDriver/issues/3403
- An alternative of https://github.com/OpenTabletDriver/OpenTabletDriver/pull/3412

This PR uses simple `read` instead of `epoll` in #3412, and cleans up `LinuxTimer` a bit.

Not sure which one is better, but I'm just posting this PR because I finished the work lol


OpenTabletDriver.Desktop/Interop/Timer/LinuxTimer.cs
```diff
@@ -32,49 +27,74 @@ public void Start()
...
-                    Enabled = true;
+                    _timerThread = new Thread(() =>
+                    {
+                        while (enabled)
+                        {
+                            ulong timerExpirations = 0;
+
+                            if (TimerGetTime(_timerFD, ref timerExpirations, sizeof(ulong)) == sizeof(ulong) && enabled)
+                            {
+                                for (ulong i = 0; i < timerExpirations; i++)
```
X9VoiD, line 73: Will this not cause weird "catch-up" issues when system is too slow? Probably better to fire here only once than with too small of an interval.

hwsmm: I think that sounds better, it's now fixed.

X9VoiD (approved):

PR #2772, opened 2023-06-05
Migrate to Avalonia

X9VoiD:
> Don't mind the messy git history and weird build script changes, I'll clean it up before removing PR's draft status.

### Merge all three UX projects and rename to OpenTabletDriver.UI

UX is a misnomer. It should have been UI from the very beginning.

### Migrate from Eto to Avalonia

Sneak peek

image

The transparency is configurable.

image

> Don't mind "Random Stuff". It was used for debugging during the very early stages.

### Dark mode by default

Light mode not in scope of this PR

### Tablets Overview

https://github.com/OpenTabletDriver/OpenTabletDriver/assets/10338031/05d4bbd4-02ed-4587-a6fe-2f6f6276fede

> While writing this PR I accidentally triggered a "tablet duplication" bug from #2601 but it just so happened that that bug helps present the new tablet overview lol.

This overview paves the way for adding custom graphics for different tablets (not that we should implement it now) and even differentiate from display tablets. It will also be used for notifying users when a tablet is "known to be there but has problems" where pre-PR it would drop it and say "No tablet detected" but here they will remain but in a disabled state. See this toggle here.

image

Due to changes coming from #2601 it is now possible to get tablets with problems to remain "detected" but in a disabled state. Also allows for disabling input processing for certain tablets if the user wanted to.

### Auto-start on boot

Yep. It's finally happening. This option is only for windows for now and probably in the future as well.

### Daemon will no longer launch with UI

The plan is to use another means to launch the daemon. For example, we can use Task Scheduler on Windows to launch OTD.Daemon. This is to avoid situations where accidental closing or a crash of UI also closes Daemon.

## Remaining core features to implement

No merging without these.

- [x] Bindings
- [ ] Filters
- [ ] Tools

There are no plans of bringing back Configuration Editor, we're better off just improving the docs around creation of configs and parsers.

## Stretch goals

Things that can be done on a future PR

- [ ] Plugin Manager
- [ ] Diagnostics

PR #3427, opened 2024-08-13
[0.6.x] Show max recorded position in the Tablet Debugger.

### LINKED ISSUE #2433: Show max registered coordinates in tablet debugger
### Acknowledgements

### Description

This would speed up config creation a little bit, and ensure configs (X/Y values) are more likely to be correct

Here's what I imagine the current workflow is like:
- User needs help getting unknown/unsupported tablet supported
- Helper eventually gets endpoint connected with correct parser so that OTD tablet debugger shows data
- Helper now need max X and Y values reported by the tablet
- User is then tasked to manually read off the max X and Y values reported by the tablet debugger (which is reporting at higher rates than users should be expected to catch a glimpse of the value)

Alternatively, fixing #1657 would likely make this obsolete, as I imagine that would have an impact on workflow (replacing the sequence we currently do with tablet debugger)

silkneko:
Sorry if anything including naming convention is off, this is my first time touching C# let alone this project, please mention anything and I'll do my best to get things inline with current choices.

This pull requests currently targets `0.6.x`. I have noticed that `master` is a bit stalled right now and the entire UI is being replaced with #2772, so I won't be submitting an obsolete PR.

I wasn't really sure what to title the new `StackLayoutItem` so the name right now is a [bit redundant](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/ea5c3836f241c6fc16241c6b27e869adb3d750e3/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs#L127), but its also the only bit of data that currently benefits from having a field that shows the max.

I also thought about adding a button to reset the data, but considering switching tablet on the same debugger instance is an unusual scenario when using this feature, as typically the only reason to use it would be to support a new tablet? So I've not included one at this time.

Let me know on any changes or nitpicks, I'm not sure if I missed anything for this project and I'd rather not add any additional overhead from me trying to snipe some easy issues.

Fixes #2433 but solely on `0.6.x`.

### DIFF
--- a/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
+++ b/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.Linq;
+using System.Numerics;
 using System.Text;
 using Eto.Drawing;
 using Eto.Forms;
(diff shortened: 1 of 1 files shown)


gonX (commented):
Looks okay:
image

...until you get a fat tablet debugger parse and lines get hidden:
image

This is more of an issue with the tablet debugger window itself since fat parses could already do that, now it's just more apparent because a bunch of lines were stolen by this feature.

We should fix so that the tablet debugger reflows that stack above the fold, though probably not in this PR. But the feature works and the code LGTM, just want to hear what other people think based off of this.

PR #3433, opened 2024-08-21
docs: Simplify README_PTBR.md

meirochun:
Reduced verbosity in the README_PTBR.md by condensing the text and removing redundant information. The updated documentation is now more concise and easier to follow.

In few parts was changed the whole word "OpenTabletDriver" to the short way "OTD". Because in Brazil now is more often people saying the abbreviated version.

### DIFF
--- a/docs/README_PTBR.md
+++ b/docs/README_PTBR.md
@@ -4,7 +4,7 @@
 
 [English](../README.md) | [Español](README_ES.md) | [Français](README_FR.md) | [Deutsch](README_DE.md) | Português-BR | [Nederlands](README_NL.md) | [한국어](README_KO.md) | [Русский](README_RU.md) | [简体中文](README_CN.md) | [繁體中文](README_TW.md) | [Ελληνικά](README_EL.md) | [Magyar](README_HU.md)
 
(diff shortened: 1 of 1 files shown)


meirochun:
I will find someone to review the text.

jamesbt365:
> In few parts was changed the whole word "OpenTabletDriver" to the short way "OTD".

We try and avoid this in our official documentation, at least in practice and I personally would rather keep things in line with the rest of our docs. It doesn't really make that much sense to refer to it by an acronym in one localisation, its just generally unnatural.

n nearly every language, it is common for individuals to use abbreviations or acronyms. However, in more formal documentation, this practice is less frequent. I would prefer to maintain consistency here and keep it in line with the rest of the documentation.

meirochun:
That makes sense, I'll change it.

MrFelopes (changes requested):
At "OpenTabletDriver é um programa código aberto", there should be a "de" between "programa" and "código". It should be: "OpenTabletDriver é um programa de código aberto...".

MrFelopes (approved):
Now, it's good to go!

meirochun:
@jamesbt365 review done.

jamesbt365 (approved):

Issue #2124, opened 2022-03-14
Tablet Configurations should have an eraser specification.


jamesbt365:
## Description

The tablet configurations should have an eraser specification so that tablets without an eraser don't show this. However this could also interfere with tablets that support multiple pens, removing a possibility for a binding to be used.

PR #3438, opened 2024-08-25
Adding support for Turcom TS-6580. 

VoidJuiceConcentrate:
Adding support for Turcom TS-6580. Confirmed working in Windows 11.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/H610.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/H610.json
@@ -63,7 +63,7 @@
         ]
       }
   ],
-    "AuxilaryDeviceIdentifiers": [],
+    "AuxiliaryDeviceIdentifiers": [],
     "Attributes": {
       "libinputoverride": "1"
     }
--- a/OpenTabletDriver.Configurations/Configurations/Huion/H610.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/H610.json
@@ -11,28 +11,55 @@
         "MaxPressure": 2047,
         "ButtonCount" : 2
       },
-        "ActiveReportID": {
-          "Start": 64,
-          "StartInclusive": true,
-          "End": null,
-          "EndInclusive": false
-        }
+      "AuxiliaryButtons": null,
+      "Touch": null
       },
       "DigitizerIdentifiers" : [
       {
         "VendorID": 9580,
         "ProductID": 110,
-        "InputReportLength": 8,
+        "InputReportLength": 2,
         "OutputReportLength": null,
-        "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogicReportParser",
+        "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicReportParser",
         "FeatureInitReport": null,
         "OutputInitReport": null,
-        "DeviceStrings": {
+        "DeviceStrings": {  
+      "6": "Huion",
+          "121": "M508"
+        }
+      },
+      {
+        "VendorID": 9580,
+        "ProductID": 110,
+        "InputReportLength": 3,
+        "OutputReportLength": null,
+        "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicReportParser",
+        "FeatureInitReport": null,
+        "OutputInitReport": null,
+        "DeviceStrings": {  
       "6": "Huion",
           "121": "M508"
         },
         "InitializationStrings": [
-          "100"
+          100,
+          123
+        ]
+      },
+       {
+        "VendorID": 9580,
+        "ProductID": 110,
+        "InputReportLength": 8,
+        "OutputReportLength": null,
+        "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicReportParser",
+        "FeatureInitReport": null,
+        "OutputInitReport": null,
+        "DeviceStrings": {
+      "6": "Huion",
+      "121" : "M508"
+        },
+        "InitializationStrings": [
+          100,
+          123
         ]
       }
   ],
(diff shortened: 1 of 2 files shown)


TABLETS.md
```diff
@@ -21,6 +21,7 @@
 | Huion H320M                   |     Supported     |
 | Huion H420X                   |     Supported     |
 | Huion H430P                   |     Supported     |
+| Huion H610                    |     Has Quirks    | Buttons not currently assignable on Windows Platform. Application thinks there is an eraser on pen.
```
jamesbt365, line 24: This is not a Huion H610, this is a Turcom TS-6580, please mark it as so.

Buttons should be able to be resolved, please export diagnostics as previously stated.

OTD has no eraser spec, so they'll always show as bindable, there is no need for a note #2124.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
```diff
@@ -0,0 +1,70 @@
+{
+    "Name": "Huion H610",
```
jamesbt365, line 2: Ditto, see comment on TABLETS.md.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
```diff
@@ -0,0 +1,70 @@
...
+        "OutputInitReport": null,
+        "DeviceStrings": {  
+      "6": "Huion",
+          "121": "M508"
+        },
+        "InitializationStrings": [
+          100,
+          123
+        ]
+      },
```
jamesbt365, line 47: Both these identifiers are useless, only the bottom one is meaningful.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
```diff
@@ -0,0 +1,70 @@
...
+       {
+        "VendorID": 9580,
+        "ProductID": 110,
+        "InputReportLength": 8,
+        "OutputReportLength": null,
+        "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicReportParser",
+        "FeatureInitReport": null,
+        "OutputInitReport": null,
+        "DeviceStrings": {
+      "6": "Huion",
```
jamesbt365, line 57: Lint.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
```diff
@@ -0,0 +1,70 @@
...
+        },
+        "InitializationStrings": [
+          100,
+          123
+        ]
+      },
+       {
+        "VendorID": 9580,
+        "ProductID": 110,
+        "InputReportLength": 8,
```
jamesbt365, line 51: Are you sure this can be opened on Windows without WinUSB?

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
```diff
@@ -0,0 +1,70 @@
...
+      "6": "Huion",
+      "121" : "M508"
+        },
+        "InitializationStrings": [
+          100,
+          123
+        ]
+      }
+  ],
+    "AuxilaryDeviceIdentifiers": [],
```
jamesbt365, line 66: The configuration spec is different on master than it is on `0.6.x`.
```suggestion
    "AuxiliaryDeviceIdentifiers": [],
```

jamesbt365 (changes requested):
All configurations submitted must follow our [CONTRIBUTING.md](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/CONTRIBUTING.md) and thus must contain a diagnostic (help > export diagnostics)

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
VoidJuiceConcentrate: Nothing worked until I added these two identifiers.

jamesbt365: Export diagnostics.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
VoidJuiceConcentrate: image
image
image

jamesbt365: Just export diagnostics.

VoidJuiceConcentrate: [turcom-ts-6580.txt](https://github.com/user-attachments/files/16738647/turcom-ts-6580.txt)

jamesbt365:
This tablet is an absolute mess and unless you want to brute force this a package capture from the OEM driver is probably the best bet here.

There probably isn't any reason have 3 identifiers *required* for the functionality of the tablet. The story on Linux is much worse too because any interface can be opened so the duplicate `8` length will become a problem too, adding even more RNG to the picture. Probably causing problems until #3011 is resolved.

OpenTabletDriver.Configurations/Configurations/Huion/H610.json
jamesbt365: Looking back previously when you originally got help for this, they weren't required, this is likely user error as I doubt anything has changed since, won't know for sure without a capture of the OEM driver.

VoidJuiceConcentrate: Removing these two segments breaks compatibility. I'm adding the new identifiers to this and I'll see how that works with the tablet.  

Currently it's not feasible to do any USB packet capture from the original drivers, both versions I could get my hands on have less functionality and more issues than this "absolute mess" does on a Windows 11 system. 

Also, it's not exactly unheard of that a USB device might function different on Linux/Ubuntu than it does on Windows.

jamesbt365: Seeing what the OEM driver does to initialize the tablet is enough, it doesn't matter what features that driver has.

PR #3453, opened 2024-09-06
Fix new firmware deco pro gen2

Kuuuube:
So... it was not a good idea to use subtraction here. In newer versions of the xp pen firmware they removed the troublesome bit that was being subtracted (`00 40`) and that makes this parser overflow the pressure value.


jamesbt365:
Is it possible that this change breaks the older firmware?

Kuuuube:
No, the only reason the subtraction was there is to remove `00 40`. https://github.com/OpenTabletDriver/OpenTabletDriver/pull/3228#discussion_r1605963487

jamesbt365 (approved):

PR #3475, opened 2024-09-21
[Backport] Attribute overhaul

jamesbt365:
Attempts to backport #3245.

As a consequence of the diverging branches, history had to be killed, I also have not tested the changes on the backport yet.

Additionally, this PR fixes the linting on some of the tablets that had mangled formatting when they were backported.

It also fixes the name from #3285, as this did not have a `-` in the DTK-2200. The DTK-1660 is also like this, but as that config has been in a release already, changing it would "reset" the settings of users all for a minimal formatting change.


jamesbt365:
I tested the HID path on my PTZ-630 on Linux, seems to work.

jamesbt365:
Needs to take yet another fix that I found when reviewing #3289.

X9VoiD (approved):

PR #3480, opened 2024-09-23
Add MacOS configuration override directory

### LINKED ISSUE #3402: Add configuration override directory to macos
Not sure what location is best, but creating configurations on MacOS is challenging and currently requires a recompile on each change.

jamesbt365:
Fixes #3402.

Untested.


jamesbt365:
Can you do this without adding this code then? I think I got a user to try something before and we just couldn't get overrides to work. It was over a year ago at this point so I'm not sure if it actually is a problem or was just user error.

AkiSakurai:
Works on my machine

jamesbt365:
Tested with a user on discord, this isn't needed.

Issue #3435, opened 2024-08-22
Add support for Fansjoy S2 tablet


sekatol:
### Manufacturer

Fansjoy

### Product Name

FJ-S2

### Device Width

115mm

### Device Height

72mm

### Pressure Levels

8192

### Lines per Inch (LPI)

5080

### Additional Information

Without any driver, this tablet sends wrong cursor position to system.
Like this:
```
+------+                   +-------------+
|      |                   |             |
|      | --map-to-screen-> |             |
|      |                   +-------------+
+------+                       screen
 tablet
```
But it should be like this: (rotate 270 degree clockwise)
```
+----------+                   +-------------+
|          | --map-to-screen-> |             |
|          |                   |             |
+----------+                   +-------------+
   tablet                          screen
```

### Diagnostic Information

[diag.json](https://github.com/user-attachments/files/16703013/diag.json)

### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

InfinityGhost:
Can you provide some sort of proof this thing exists? A pretty thorough search doesn't show anything at all.

sekatol:
Sorry for the inconvenience. This tablet is from a small corporation in China, and I bought it from Taobao, an online shopping platform in China. I found the [official website](https://www.fans-joy.com/s) of Fansjoy through Bing Chinese. (Sorry for no translations.) The official website has an entry for this model, "S2数位板" for S2 tablet. The button below it is "了解更多" means "Learn more", but it is just a link to Taobao site, 😂. I copied information on that website:
- Pressure Level: 8192
- Resolution: 5080 LPI
- Read Frequency: 266 points/second
- Max Tilt Angle: 60 degree
- Response Height: 1 cm
- 4 customizable buttons
- A LED light
- Additional information: Press the first and fourth button simultaneously for 6 seconds to toggle orientation. 
- There are 2 buttons on the digital pen. The upper one is for the mouse middle button and the lower one is for the eraser.
- No need to charge the pen.

I don't know if there is a need to build a driver for this tablet, because there are really a few people using this model of tablet, so working on this tablet is time wasting.
Or, if possible, I will try to learn from the recent pull requests and contribute codes for this tablet, giving any guidances are the best.

jamesbt365:
[...]

PR #3484, opened 2024-10-02
Add support for Fansjoy S2

sekatol:
For #3435
Note:
There is a function to change orientation by pressing the 1st and 4th button simultaneously for 6 sec. But this action doesn't sending any report at all.
Orientation can be changed in OTD. So I think there's no need to add that function.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
+++ b/OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
@@ -2,8 +2,8 @@
   "Name": "Fansjoy FJ-S2",
   "Specifications": {
     "Digitizer": {
-      "Width": 115,
-      "Height": 72,
+      "Width": 160,
+      "Height": 100,
       "MaxX": 10500,
       "MaxY": 16800
     },
--- a/OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
+++ b/OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
@@ -1,0 +1,38 @@
+{
+  "Name": "Fansjoy FJ-S2",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 115,
+      "Height": 72,
+      "MaxX": 10500,
+      "MaxY": 16800
+    },
+    "Pen": {
+      "MaxPressure": 8191,
+      "ButtonCount": 2
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 4
+    },
+    "MouseButtons": null,
+    "Touch": null
+  },
+  "DigitizerIdentifiers": [
+    {
+      "VendorID": 11648,
+      "ProductID": 12307,
+      "InputReportLength": 14,
+      "OutputReportLength": 0,
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Fansjoy.FansjoyReportParser",
+      "FeatureInitReport": null,
+      "OutputInitReport": null,
+      "DeviceStrings": {},
+      "InitializationStrings": [],
+      "Attributes": {}
+    }
+  ],
+  "AuxiliaryDeviceIdentifiers": [],
+  "Attributes": {
+    "libinputoverride": "1"
+  }
+}
--- a/OpenTabletDriver.Configurations/Parsers/Fansjoy/FansjoyTabletReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Fansjoy/FansjoyTabletReport.cs
@@ -1,0 +1,33 @@
+using System;
+using System.Numerics;
+using OpenTabletDriver.Tablet;
+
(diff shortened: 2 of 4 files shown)


OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
```diff
@@ -0,0 +1,38 @@
+{
+  "Name": "Fansjoy FJ-S2",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 160,
+      "Height": 100,
+      "MaxX": 10500,
+      "MaxY": 16800
```
sekatol, line 8: Yes, the horizontal resolution is less than the vertical resolution. XD

OpenTabletDriver.Configurations/Parsers/Fansjoy/FansjoyTabletReport.cs
```diff
@@ -0,0 +1,32 @@
...
+                X = (report[5] << 8 | report[4]),
+                Y = 16800 - (report[3] << 8 | report[2])
+            };
+
+            Pressure = (uint)(report[7] << 8 | report [6]);
+
+            PenButtons = new bool[]
+            {
+                report[1].IsBitSet(2),
+                report[1].IsBitSet(1)
```
sekatol, line 23: Without any drivers, when the 1st pen button is pressed, the pen is used as an eraser. Now this function is not working with the default configuration. Can users change settings to use the eraser when the specific pen button is pressed?
Or should it be implemented in code?

sekatol:
Oh, I've forgotten upload the diagnostic file.
[diag3.txt](https://github.com/user-attachments/files/17266126/diag3.txt)

OpenTabletDriver.Configurations/Parsers/Fansjoy/FansjoyTabletReport.cs
gonX: In this case where it sounds like it's just a normal pen button, I personally don't think you should be setting eraser based off of that.

gonX:
Fix tests then I think we can likely merge this

OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
```diff
@@ -0,0 +1,38 @@
+{
+  "Name": "Fansjoy FJ-S2",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 165,
+      "Height": 91,
+      "MaxX": 10500,
```
gonX, line 7: This width combined with this `MaxX` is confusing me. Typically, tablets are 200LPI or 100LPI, and in very rare instances, the tablet maps to a square area (like 4096x4096 or 16384x16384). With the proposed settings, this would be a very unique tablet by having ~184 LPI horizontally, and ~63 LPI vertically.

Please double check the **active area** width/height with a ruler to ensure that you are within a couple of millimeters - that or you have a mistake in your parser.

gonX (commented):
If possible, please attach a tablet recording from the tablet debugger where you move to all 4 corners. That feature is only present on the 0.6.x branch, so you will have to port your changes to that branch. Fortunately, it should just be a line or 2 changed in the tablet json configuration.

OpenTabletDriver.Configurations/Configurations/Fansjoy/FJ-S2.json
sekatol: Sure, the LPI is weird...
I double checked the active area and ported changes to branch `0.6.x` and logged tablet recordings.
[tablet-data.txt](https://github.com/user-attachments/files/17801194/tablet-data.txt)

gonX:
I feel like it's a mistake to subtract what is the same const as MaxY - it very likely makes the parser unportable for other tablets using the same parser.

That leaves us with 2 options:
- Refactor the position parsing (at a quick glance, data recording look like it's 0,0 aligned with bottom-left rather than top-left as OTD expects)
- Merge it as-is and then when the next Fansjoy tablet shows up, we can finalize the parser format by then

I'll take a look at this one of the (hopefully) coming days and see if I can up with something smarter, otherwise I think we'll just go with option 2. CC @X9VoiD

gonX:
It seems to me like you're maybe parsing the tablet rotated?
(reports are Y then X, rather than the likely expected X then Y)

"top left", but "top right" rotated:
[0x0000, 0x41A0] = [0, 16800], you parsed [0, 0]

"top right", but "bottom right" rotated:
[0x2904, 0x41A0] = [10500, 16800], you parsed [10500, 0]

"bottom right", but "bottom left" rotated:
[0x2904, 0x0000] = [10500, 0], you parsed [10500, 16800]

"bottom left", but "top left" rotated:
[0x0000, 0x0000] = [0, 0], you parsed [0, 16800]

gonX (changes requested):

Also please ensure that you don't have commits that touch non-configuration files.

sekatol:
Continue at #4125

PR #3488, opened 2024-10-06
Add support for hiding the daemon console on Windows

BunnyTub:
Frees the console when the user specifies the added command line argument --hidden or -h. If the user is not running Windows, the console will not be freed (for now --- not sure how to add support for other platforms just yet).

This pull request also explicitly adds strings for options with strings to try and avoid unintended behavior.

### DIFF
--- a/OpenTabletDriver.Daemon/Program.cs
+++ b/OpenTabletDriver.Daemon/Program.cs
@@ -28,14 +28,15 @@ namespace OpenTabletDriver.Daemon
 
                 var rootCommand = new RootCommand("OpenTabletDriver")
                 {
-                    new Option(new[] { "--appdata", "-a" }, "Application data directory")
+                    new Option<string>(new[] { "--appdata", "-a" }, "Application data directory")
                     {
                         Argument = new Argument<string>("appdata")
                     },
-                    new Option(new[] { "--config", "-c" }, "Configuration directory")
+                    new Option<string>(new[] { "--config", "-c" }, "Configuration directory")
                     {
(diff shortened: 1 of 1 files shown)


BunnyTub:
This was written originally on OpenTabletDriver v0.6.4.0, then ported forwards.
[Program.cs of v0.6.4.0.zip](https://github.com/user-attachments/files/17268699/Program.cs.of.v0.6.4.0.zip)

gonX:
Just trying to understand what this is solving - you want the daemon window away from the taskbar?

Could a tray icon be negotiated instead? Not sure how the Windows API works in that regard, but I would prefer the daemon to minimize to tray rather than hiding entirely.

BunnyTub:
Sure, a tray icon could work to show the console again, but that's already the job of the UX for the GUI.
This completely hides the console (or rather, destroys it).

The Windows API is a little wonky to deal with, especially without something like a wrapper such as WPF or WinForms, and I'm hoping that this code can be modified (even if greatly) to be cross platform. A tray icon would be easier to implement on Windows though, at least from my perspective.

gonX:
This at the very least needs to not break builds on the other platforms. Converting to draft until that is fixed.

With that being said, I am not sure there is a huge interest for this yet.

gonX:
As in #3489:
> Sorry, but I don't think there is interest for this, and with no cross-platform solution in sight I am closing this PR.

(thank you for trying though, we really appreciate all PR's!)

BunnyTub:
No problem, it probably isn't needed anyways /shrug

PR #3489, opened 2024-10-06
Add support for hiding the daemon console on Windows (0.6.x)

BunnyTub:
This pull request is for the 0.6.x branch.

Frees the console when the user specifies the added command line argument --hidden or -h. If the user is not running Windows, the console will not be freed (for now --- not sure how to add support for other platforms just yet).

### DIFF
--- a/OpenTabletDriver.Daemon/Program.cs
+++ b/OpenTabletDriver.Daemon/Program.cs
@@ -1,9 +1,10 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.CommandLine;
 using System.IO;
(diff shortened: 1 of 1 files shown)


gonX:
Marking as draft for the same reason as my comments on #3488:
> This at the very least needs to not break builds on the other platforms. Converting to draft until that is fixed.
> 
> With that being said, I am not sure there is a huge interest for this yet.

Additionally, it seems that not all checks were run correctly here (only 1 was run). Not sure if this is a GitHub issue.

gonX:
Sorry, but I don't think there is interest for this, and with no cross-platform solution in sight I am closing this PR.

PR #3490, opened 2024-10-06
Prevent multiple exception dialogs from opening at once (spamming) (0.6.x)

BunnyTub:
Uses the **MessageBoxActive** bool to determine if there is an exception dialog already open.

This is to prevent cases where for example, a plugin throws an unhandled exception, or something goes horribly wrong in OpenTabletDriver, and exception windows are constantly shown. Things such as that can cause the system to slow down, and is definitely a problem, since not all windows can be closed quickly, or sometimes, even at all.

Although more than one exception won't be shown, they are logged.
There may be a better way to do this though.

### DIFF
--- a/OpenTabletDriver.UX/Extensions.cs
+++ b/OpenTabletDriver.UX/Extensions.cs
@@ -12,31 +12,41 @@ namespace OpenTabletDriver.UX
 {
     public static class Extensions
     {
+        private static bool MessageBoxActive = false;
+
         public static void ShowMessageBox(this Exception exception)
         {
+            if (MessageBoxActive)
+                return;
             Application.Instance.Invoke(() =>
             {
                 exception = exception.GetBaseException();
 
                 var dialog = new ExceptionDialog(exception);
+                MessageBoxActive = true;
                 dialog.ShowModal(Application.Instance.MainForm);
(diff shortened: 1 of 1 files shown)


BunnyTub:
image

OpenTabletDriver.UX/Extensions.cs
```diff
@@ -12,31 +12,41 @@ namespace OpenTabletDriver.UX
...
+                MessageBoxActive = true;
                 dialog.ShowModal(Application.Instance.MainForm);
+                MessageBoxActive = false;
             });
         }
 
         public static void ShowMessageBox(this CommonErrorData errorData)
         {
+            if (MessageBoxActive)
+                return;
```
gonX, line 35: Unless I'm missing something, shouldn't this be moved below the `Log.Write` statement below?

gonX (changes requested):
Looks alright other than that one thing.

OpenTabletDriver.UX/Extensions.cs
BunnyTub: Yes, you're right, fixed the oversight.

gonX (approved):

X9VoiD (approved):

PR #3503, opened 2024-10-24
Better time calculation for LinuxTimer

hwsmm:
This now looks much more saner...


gonX (approved):
I verified that the new code has the same result. LGTM.

PR #3522, opened 2024-11-15
Reduce bloat of tablet configuration JSONs

X9VoiD:


jamesbt365 (approved):

Kuuuube (approved):

gonX (approved):
Code seems good, didn't check jsons too in depth tho, but what I saw also made sense to me.

gonX:
Out of curiosity, what happens if a user uses a config with the old specifications? Will it still load?

X9VoiD:
It's only the tests that enforces that they be null, so whether the config defines nothing, null or is set to an empty array, they will still load.

gonX:
Good enough for me, merging.

PR #3551, opened 2024-11-24
Update softprops/action-gh-release to v2

Kuuuube:
Apparently softprops/action-gh-release v1 has been deprecated for over a year and might not even work on current github actions runners due to enforcing node 20.


gonX (approved):

gonX:
Merging, since we already merged the same change for 0.6.x I guess there's no harm in doing this one immediately either.

PR #3558, opened 2024-11-25
use GetExistingPathOrLast instead of GetExistingPath

AkiSakurai:
It is correct on the 0.6 branch.


jamesbt365 (approved):

gonX (approved):
Makes sense, will merge now

PR #2894, opened 2023-08-22
Support modifier click, double click, tilt and pressure for macOS

AkiSakurai:
Finishing  #2371
Fix #2871, #2291

Dedicated artist mode seems unnecessary for macOS as it does not change the behaviour like Windows Ink.

The tilt functionality has not been tested with a real device, so the scale normalisation may differ from that of the official driver.

PR #3559, opened 2024-11-25
[0.6.x] fix: double click in relative mode

AkiSakurai:
QueuePendingPosition expects relative coordinates in relative mode, but was incorrectly pulling absolute system coordinates. Now using proper (0,0) deltas for clicks.


InfinityGhost (approved):

gonX (approved):
I'll assume this is working, given you wrote most of the code. Otherwise we need to fix tests.

Nothing looks wrong with the code, so approving and merging.

/e: to note, the commit for `master` is 267e164b0402b25e7087ab3e6f686814a1837d66 in #2894

PR #3564, opened 2024-11-27
[0.6.x] Add options to disable pressure & tilt

AkiSakurai:
A separate option for disabling pressure and tilt might be better than duplicating them to create more combinations of modes.


Kuuuube:
What's the reasoning behind wanting this option? Considering #2894 mentions it not changing behavior.

AkiSakurai:
It doesn’t affect apps that don’t support pressure sensitivity. However, in apps that do, it’s not always possible to turn off pressure sensitivity to achieve a consistent stroke width. For example, I couldn’t find a way to do this in OneNote. I also remember seeing people who wanted to temporarily disable this feature.

OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -198,10 +198,22 @@ public Task SetSettings(Settings? settings)
...
+                    }
 
                     if (dev.OutputMode is RelativeOutputMode relativeMode)
+                    {
                         SetRelativeModeSettings(dev, relativeMode, profile.RelativeModeSettings);
+                        relativeMode.DisablePressure = profile.BindingSettings.DisablePressure;
+                        Log.Write(group, $"Pressure: {(profile.BindingSettings.DisablePressure ? "Disabled" : "Enabled")}");
+                        relativeMode.DisableTilt = profile.BindingSettings.DisableTilt;
+                        Log.Write(group, $"Tilt: {(profile.BindingSettings.DisableTilt ? "Disabled" : "Enabled")}");
+                    }
```
gonX, line 216: I think it would be better if we avoided repeating ourselves here, ie. refactor this into using a calling function that takes `dev.OutputMode` and `profile.BindingSettings`, and/or move the invocation to the `Set{Relative,Absolute}ModeSettings` functions.

Also curious what other people think - I'm not really requiring any changes here, just thought this bit of code looked unoptimal.

gonX (commented):
Generally LGTM, but not sure if the UI approach is correct here. Can you provide a screenshot of how this looks in the pen binding tab?

AkiSakurai:
> Generally LGTM, but not sure if the UI approach is correct here. Can you provide a screenshot of how this looks in the pen binding tab?

OpenTabletDriver.Daemon/DriverDaemon.cs
```diff
@@ -203,11 +203,20 @@ public Task SetSettings(Settings? settings)
...
+                        outputMode.DisableTilt = profile.BindingSettings.DisableTilt;
+                        Log.Write(group, $"Tilt: {(profile.BindingSettings.DisableTilt ? "Disabled" : "Enabled")}");
+                    }
+
+                    if (dev.OutputMode is IOutputMode iOutputMode)
+                    {
+                        iOutputMode.Tablet = tabletReference;
+                        var bindingHandler = CreateBindingHandler(dev, iOutputMode, profile.BindingSettings);
+                        SetOutputModeElements(dev, iOutputMode, profile, bindingHandler);
                     }
```
InfinityGhost, line 220: `DisablePressure` and `DisableTilt` should be added to the `IOutputMode` interface

InfinityGhost (dismissed):

OpenTabletDriver.Plugin/Output/OutputMode.cs
```diff
@@ -44,6 +44,16 @@ private set
...
+        /// <summary>
+        /// Whether to disable pressure
+        /// </summary>
+        public bool DisablePressure { set; get; }
+
+        /// <summary>
+        /// Whether to disable tilt
+        /// </summary>
+        public bool DisableTilt { set; get; }
+
```
gonX, line 56: Is this necessary now that it's also defined in `IOutputMode`? This class inherits from `IOutputMode` anyway.

gonX:
I'm going to move this to the 0.6.6 milestone since changes are still requested.

gonX:
Needs testing on Linux, will do this at some point.

There are merge conflicts, btw.

gonX (commented):
Pressure/tilt status is logged even when using output modes that don't implement pressure/tilt (like Absolute Mode on Windows or Linux). I feel this can be confusing for users using a mode that doesn't support this. I've attached a patchfile that fixes this.

[0001-daemon-Only-log-pressure-tilt-when-relevant.patch.gz](https://github.com/user-attachments/files/18497200/0001-daemon-Only-log-pressure-tilt-when-relevant.patch.gz)
(github didn't like the naked patch file)

The patch checks for `IPressureHandler`/`ITiltHandler` before spewing log output - this was annoying to fix given the current typing, since `Pointer` only exists in `AbsoluteOutputMode`/`RelativeOutputMode`, not in `OutputMode` itself.

gonX (dismissed):
LGTM, but this has conflicts and I would like to merge it. Do you want me to handle this or would you like to fix it yourself?

PR #3566, opened 2024-11-28
[0.6.x] Disable libinput smoothing for OpenTabletDriver virtual tablet

### LINKED ISSUE #2895: Discussion: Disable Artist Mode smoothing with libinput quirks file

gonX:
As discussed in issue #2895

Fixes #2895 for 0.6.x

Note that packaging maintainers will likely need to update packaging scripts accordingly, unless they use the generic source folder or binary tarball to base their install on.

Check:

- [x] Generic Tarball
- [x] AUR (needs to be reworked to just have OTD "install" instead of copying files manually)
- [x] Debian
- [x] Fedora
- others?

### DIFF
--- a/eng/linux/Debian/debian/opentabletdriver.install
+++ b/eng/linux/Debian/debian/opentabletdriver.install
@@ -11,3 +11,4 @@ usr/share/man/man8/opentabletdriver.8.gz
 usr/share/doc/opentabletdriver/LICENSE
 usr/share/pixmaps/otd.ico
 usr/share/pixmaps/otd.png
+usr/share/libinput/30-vendor-opentabletdriver.quirks
--- a/eng/linux/Generic/usr/share/libinput/30-vendor-opentabletdriver.quirks
+++ b/eng/linux/Generic/usr/share/libinput/30-vendor-opentabletdriver.quirks
@@ -1,0 +1,3 @@
+[OpenTabletDriver Virtual Tablets]
+MatchName=OpenTabletDriver*
+AttrTabletSmoothing=0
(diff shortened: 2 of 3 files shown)


jamesbt365:
Attempting to build with `./eng/linux/package.sh --package RedHat` fails.
```
error: Installed (but unpackaged) file(s) found:
   /usr/share/libinput/30-vendor-opentabletdriver.quirks

RPM build errors:
    cannot open Packages database in /var/lib/rpm
    Installed (but unpackaged) file(s) found:
   /usr/share/libinput/30-vendor-opentabletdriver.quirks
Build failed!
```

gonX:
AUR is working now and is ready for 0.6.5.0. Debian and Fedora still needs adjustments.

gonX:
Fedora and Debian hopefully fixed now

jamesbt365 (approved):

PR #3570, opened 2024-11-30
[0.6.x] Reduce bloat of tablet configuration JSONs

X9VoiD:
Backports #3522 to `0.6.x` branch.

For the cleanup, I used the following script:

```py
import json
import os
import sys

def process_file(filepath):
    if filepath.endswith(".json"):
        try:
            with open(filepath, 'r+') as f:
                data = json.load(f)

                def remove_redundant(obj):
                    if isinstance(obj, dict):
                        return {k: remove_redundant(v) for k, v in obj.items() if v is not None and v != {} and (not isinstance(v, list) or v)}
                    elif isinstance(obj, list):
                        return [remove_redundant(item) for item in obj if item is not None and item != {} and (not isinstance(item, list) or item)]
                    else:
                        return obj

                cleaned_data = remove_redundant(data)
                f.seek(0)
                json.dump(cleaned_data, f, indent=2)
                f.truncate()

        except (FileNotFoundError, json.JSONDecodeError) as e:
            print(f"Error processing {filepath}: {e}")
            return

    with open(filepath, 'r+', encoding='utf-8') as f:
        content = f.read()
        if not content.endswith('\n'):
            f.write('\n')

def process_directory(directory):
    for root, _, files in os.walk(directory):
        for filename in files:
            process_file(os.path.join(root, filename))

if __name__ == "__main__":
    if len(sys.argv) < 2:
        print("Usage: python cleanup.py <directory>")
        sys.exit(1)

    directory = sys.argv[1]
    process_directory(directory)
```

Will commit this to repo if desired.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
+++ b/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
@@ -15,22 +15,14 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 12
-    },
-    "MouseButtons": null,
-    "Touch": null
+    }
   },
   "DigitizerIdentifiers": [
     {
       "VendorID": 2290,
       "ProductID": 26641,
       "InputReportLength": 64,
-      "OutputReportLength": null,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.TenMoon.TenMoonReportParser",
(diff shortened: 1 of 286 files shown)


gonX (approved):
Checked code, LGTM.

PR #3583, opened 2024-12-04
Escape special characters and unicode from string reader output

### LINKED ISSUE #3582: String reader non-dumps should better show non-characters e.g. alternate hex representation
### Description

Sometimes, manufacturers like to throw in non-ASCII characters like nulls.

It would be nice if the string reader also(?) showed the string as hex values, or showed non-ASCII characters as hex.

Full string dump works fine as shown here:

image
(credit Kuuube)

### Acknowledgements

Kuuuube:
Fixes #3582 for 0.6.x, no use in adding this to master for now.

Should help when there are weird unprintable characters or some weirdness in the string.


gonX (approved):
LGTM assuming it's similar to the bottom field of this result (as was posted in Discord):
image
image

Kuuuube:
Yep. The bottom result is with `System.Text.Json.JsonEncodedText.Encode(s).ToString()`.

PR #3601, opened 2024-12-15
BindingSettings: Default pressure and eraser threshold to 1%

### LINKED ISSUE #3597: Set default pen and eraser threshold to 1%
Setting these to 0% can create a broken user experience for users of some tablets due to certain pens or tablets reporting >0 pressure constantly. Especially on linux where the pen tip getting "stuck pressed down" cannot be reset by clicking the mouse and is stuck forever unless the tablet sends a pressure value less than the threshold.

This should be a very small change, just setting `tp` and `ep`  to 1 here: https://github.com/OpenTabletDriver/OpenTabletDriver/blob/0.6.x/OpenTabletDriver.Desktop/Profiles/BindingSettings.cs#L11.

gonX:
This fixes instances of the default button being stuck on tablets with nonzero pressure

Some users would've experienced this as mouse buttons stopping working as soon as the daemon starts, especially if they use Absolute Mode on Linux

Fixes #3597 for 0.6.x

### DIFF
--- a/OpenTabletDriver.Desktop/Profiles/BindingSettings.cs
+++ b/OpenTabletDriver.Desktop/Profiles/BindingSettings.cs
@@ -8,7 +8,8 @@ namespace OpenTabletDriver.Desktop.Profiles
 {
     public class BindingSettings : ViewModel
     {
-        private float tP, eP;
+        private float tP = 1;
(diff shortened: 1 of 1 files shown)


Kuuuube (approved):
Looks good!

Issue #3606, opened 2024-12-16
[Linux] There is no obvious way to disable OpenTabletDriver for certain tablets


khalid151:
I have multiple tablets, some of which work better with OpenTabletDriver. I prefer using wacom drivers for wacom tablets.
There's no way (that I'm aware of) to use the two at the same time. There should be a way to blacklist certain tablets in OpenTabletDriver.

zxsa89675:
Hello, we tried to solve the issue.

This is what we did:

Added a blacklist feature to allow users to exclude specific tablets from being detected by OpenTabletDriver. This allows users to use other drivers (like Wacom) for certain tablets while still using OpenTabletDriver for others. Modified the Settings class to include a BlacklistedTablets property and updated the Driver class to respect this blacklist during device detection.

You can review changes in this commit: https://github.com/zxsa89675/OpenTabletDriver-OpenTabletDriver-3606/commit/e68c7c482d65d6807fd84cc5e30562b6a7c19700.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

jamesbt365:
You would be right that theres no way to disable OpenTabletDriver for certain tablets. As a workaround you can add a configuration override and remove the identifiers for the tablets you don't want OpenTabletDriver to detect.

As you are on linux create `~/.local/share/OpentabletDriver/Configurations`, find your configuration from [here](https://github.com/OpenTabletDriver/OpenTabletDriver/tree/0.6.x/OpenTabletDriver.Configurations/Configurations), copy it into that folder and empty the `DigitizerIdentifiers` and `AuxiliaryDeviceIdentifiers` (if present).

khalid151:
Great, that works, thanks.

gonX:
Related: #3664

gonX:
This is now provided on the `avalonia` branch (however, it is unfortunately broken, which is tracked in #3730)

gonX:
As a note for a future implementer: we probably want to disable it entirely by not even probing the device, to avoid device behavior changing. It will likely not be enough for handled tablets to simply not emit their output.

I'll take a quick look at seeing if this is possible to add for 0.6.7, otherwise it'll probably go to the next milestone.

daxbo:
[...]

PR #3607, opened 2024-12-16
Use `blacklist` to disable `wacom` and `hid_uclogic` modules

khalid151:
Simply installing OpenTabletDriver breaks wacom drivers on Linux. I had to waste hours trying to fix a non-existent issue thinking it was related to libinput just because I forgot I had OpenTabletDriver installed.

`install wacom /usr/bin/true` should **NOT** be a thing. It renders modprobe for loading drivers useless.

### DIFF
--- a/eng/linux/Generic/usr/lib/modprobe.d/99-opentabletdriver.conf
+++ b/eng/linux/Generic/usr/lib/modprobe.d/99-opentabletdriver.conf
@@ -1,2 +1,2 @@
-install wacom /usr/bin/true
-install hid_uclogic /usr/bin/true
+blacklist wacom
+blacklist hid_uclogic


jamesbt365:
OpenTabletDriver is supposed to be a replacement to the kernel drivers and isn't really supposed to be used alongside them.

If the kernel drivers weren't so nasty to OpenTabletDriver and didn't cause problems for users we would go back to the blacklist, but we moved to `install module /usr/bin/true` because in some systems it is not respected, which is undesirable for most people as they are trying to use OpenTabletDriver as a *replacement*.

`blacklist module` allows the system and other programs to load the module, not respecting what usually needs to be done for OpenTabletDriver to function, and the method we picked doesn't actually render the modules completely unloadable, you can still load them through `modprobe -i module`.

I don't really think there is much harm in leaving it as is because its supposed to be a replacement and a large majority of our userbase uses it as one.

I'm willing to hear other peoples opinions on this but prior to us using `install` instead of `blacklist` we would get multiple users a week in our support channels unable to figure out that the kernel modules are loading (because their DE/applications aren't respecting the blacklist) rendering OpenTabletDriver (the supposed replacement) to work.

khalid151:
> OpenTabletDriver is supposed to be a replacement to the kernel drivers and isn't really supposed to be used alongside them.

Which wouldn't really solve #3606, in case anyone would need to use different drivers for different tablets.

If it's better to stick with altering the install script, then it's much better to (at least) use `/usr/bin/false`, so it would be obvious there is something off while loading the module manually.

I still believe this isn't the right approach, especially that just installing otd breaks other drivers, even when otd isn't being used. Adding those rules manually makes much more sense.

I have few suggestions:
- Notify users with a warning when modules are loaded. It's fairly easy to check if those specific modules are loaded in `/sys/module/<module_name>`
- Remove module with otd service starts.

The easiest way to implement the latter is to add the following udev rules:
```
ACTION=="add", SUBSYSTEM=="input", ATTRS{name}=="OpenTabletDriver*", ATTRS{phys}=="", RUN+="/usr/bin/rmmod wacom"
ACTION=="add", SUBSYSTEM=="input", ATTRS{name}=="OpenTabletDriver*", ATTRS{phys}=="", RUN+="/usr/bin/rmmod hid_uclogic"

ACTION=="remove", SUBSYSTEM=="input", ATTRS{name}=="OpenTabletDriver*", ATTRS{phys}=="", RUN+="/usr/bin/modprobe wacom"
ACTION=="remove", SUBSYSTEM=="input", ATTRS{name}=="OpenTabletDriver*", ATTRS{phys}=="", RUN+="/usr/bin/modprobe hid_uclogic"
```

When OTD service starts, it creates a uinput virtual device that can matched with udev rules.

gonX:
> Notify users with a warning when modules are loaded. It's fairly easy to check if those specific modules are loaded in /sys/module/<module_name>

We already do this, IIRC

The module removal on udev load is somewhat smart though. My only concern with that approach is that OTD expects tablets to be uninitialized, and some tablets might stay initialized even when you `rmmod`.

Switching to `/usr/bin/false` might be an option but I would be afraid that this could break parts of user setups in surprising ways, like making the WM/DE unable to start. It is probably the better option of the 3 provided if it doesn't cause a bunch of side effects.

FWIW, I did contest the same usage when we originally implemented it (can't find the quote, sadly), but our driver conflict issues for official installed packages has dropped to 0 since, so I would be hesitant to change too much of this behavior.

gonX:
After further internal discussion we don't want to switch back a blacklist quite yet.

Closing to reduce confusion - please continue discussion in #3606. We can reopen this PR if consensus changes.

PR #3613, opened 2024-12-17
Add config json linting to 0.6.x

### LINKED ISSUE #3599: 0.6.x tests need to include json linting
We already have this for master and it should be a simple copy paste of `Configurations_Are_Linted` (https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/OpenTabletDriver.Tests/ConfigurationTest.cs#L321).

Issues such as #3598 can occur without this testing.

Kuuuube:
Fixes #3599. Prevents issues such as #3598.

99% copy paste from master and #3049.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
+++ b/OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
@@ -2,10 +2,10 @@
   "Name": "10moon 1060N",
   "Specifications": {
     "Digitizer": {
-      "Width": 254.0,
+      "Width": 254,
       "Height": 152.4,
-      "MaxX": 4095.0,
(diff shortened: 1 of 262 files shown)


gonX (commented):
I think this looks good but the are some check warnings - did we have this on the original PR as well?

Kuuuube:
The warnings are because the nullable property was already set in the project this was added to on master but it wasn't for 0.6.x. Shouldn't cause any harm to enable it here.

gonX (approved):
LGTM.

As for my earlier concern with check warnings - they're fine - looking into it it might actually help us uncover more null references in the long run. We should whack a mole the warnings ASAP (though not massively high priority) to help keep checks clean for future PR's.

PR #3614, opened 2024-12-18
[0.6.x] Fix a case nullref when In-Range Apply occurs

Mrcubix:
Reported in https://github.com/Mrcubix/Preset.Binding/issues/2#issuecomment-2516217342


gonX (approved):
LGTM

PR #3615, opened 2024-12-18
[0.6.x] Fix AppInfo desync in UX

### LINKED ISSUE #3575: UX-Side Plugin Manager, Preset Manager are not instantiated using daemon AppInfo, defaulting to local appdata
### Steps to Reproduce

1. Build 0.6.x (Either in IDE, or publish it),
2. Create a userdata folder next to the generated Daemon binary,
3. Move the UX away (in the case of running in the IDE, it's already done),
4. UX fetch AppInfo from Daemon,
5. UX proceed to use the Plugin Manager & Preset Manager instantiated with the old AppInfo.

### Expected Behavior

Instantiate a new Plugin Manager & Preset Manager or change the plugin directory when the daemon is connected.

### Observed Behavior

Plugin Manager, Preset Manager & other parts of the UX with old path are not updated, as a result Plugins & Presets will be loaded from the wrong folder in the UX 
(potentially being outdated, since from a separate install)

### Tablet Device

All

### Diagnostic Information

Daemon: [daemon.json](https://github.com/user-attachments/files/17981782/daemon.json)

image

### Acknowledgements

Mrcubix:
Attempt at fixing the Plugin Manager & Preset Manager being instantiated with the wrong AppInfo.

Closes #3575

### DIFF
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -76,7 +76,6 @@ namespace OpenTabletDriver.UX
         private const int DEFAULT_CLIENT_WIDTH = 960;
         private const int DEFAULT_CLIENT_HEIGHT = 760;
 
-        private TabletSwitcherPanel mainPanel;
         private MenuBar menu;
         private Placeholder placeholder;
         private TrayIcon trayIcon;
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -11,6 +11,7 @@ using Newtonsoft.Json.Linq;
 using OpenTabletDriver.Desktop;
 using OpenTabletDriver.Desktop.Diagnostics;
 using OpenTabletDriver.Desktop.Interop;
+using OpenTabletDriver.Desktop.Reflection;
 using OpenTabletDriver.Interop;
 using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Logging;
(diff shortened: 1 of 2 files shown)


Mrcubix:
image

Well that's an odd way for unit tests to fail

Mrcubix:
Might need to unload all plugins before re-instantiating the new manager

gonX (changes requested):
satisfy warnings (remove `mainPanel` definition in MainForm)

OpenTabletDriver.UX/MainForm.cs
```diff
@@ -369,6 +369,9 @@ private void HandleDaemonConnected(object sender, EventArgs e) => Application.In
             // Load the application information from the daemon
             AppInfo.Current = await Driver.Instance.GetApplicationInfo();
 
+            AppInfo.PluginManager = new DesktopPluginManager();
+            AppInfo.PresetManager = new PresetManager();
```
gonX, line 373: I don't know enough about this part of the code - shouldn't this be grabbing the manager from the daemon rather than making a new one instead?

Mrcubix: The UX instantiate its own Preset & plugin manager (in static properties).
The constructors of these objects use a path, that is only synchronized once the UX connects to the daemon.
This result in the Plugin Manager & Preset Manager, pulling data from the wrong path.

Mrcubix: By re-instantiating the object, you force them to use the new path provided by the daemon.

gonX: Ok, makes sense.

gonX (approved):

PR #3623, opened 2024-12-19
Ensure Store is properly set when opening advanced binding editor dialog

Mrcubix:
Imagine this scenario : 

The user spends some time changing settings for a binding (Scroll Binding for example)

image

Then they Apply those settings. Suddenly they realize they made a mistake, and need to change a settings, they open back the advanced binding editor to this : 

image

The settings are gone, they have to set everything back by hand.

That scenario has been me a dozen of times while testing as well as my users (and probably more users outside of just my plugins)


gonX (approved):
Can confirm it fixes the issue (tested with Linux Artist Mode dropdown).

Not sure if there are side effects - approving, but letting someone else merge.

gonX (approved):
After some further thought, this seems safe to merge to me now.

PR #3631, opened 2024-12-23
[0.6.x] Fix another case of Output Mode NullException when saving using the tablet

Mrcubix:
I've identified the core issue : The output mode is being used while it's being initialized.
`InputDeviceTree` should not call `Read()` on the output mode until it's ready.
This could be avoided by setting the OutputMode at a later time.

Now the real question is:  **How, has this not been caught in 5 minutes of testing.**

The other question is related to plugins : 

Some plugins fetch the output mode from the device itself, for diverse reasons.
Setting the output mode later would result in a failure to fetch such output mode (unless they check on every reports)

### DIFF
--- a/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
@@ -153,7 +153,7 @@ namespace OpenTabletDriver.Plugin.Output
             if (report is ITiltReport tiltReport && Pointer is ITiltHandler tiltHandler)
                 tiltHandler.SetTilt(tiltReport.Tilt);
             if (report is ITabletReport tabletReport && Pointer is IPressureHandler pressureHandler
-                && Tablet.Properties.Specifications.Pen != null)
+                && Tablet != null && Tablet.Properties.Specifications.Pen != null)
                 pressureHandler.SetPressure(tabletReport.Pressure / (float)Tablet.Properties.Specifications.Pen.MaxPressure);
 
             // make sure to set the position last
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
```diff
@@ -153,7 +153,7 @@ protected override void OnOutput(IDeviceReport report)
             if (report is ITiltReport tiltReport && Pointer is ITiltHandler tiltHandler)
                 tiltHandler.SetTilt(tiltReport.Tilt);
             if (report is ITabletReport tabletReport && Pointer is IPressureHandler pressureHandler
-                && Tablet.Properties.Specifications.Pen != null)
+                && Tablet != null && Tablet.Properties.Specifications.Pen != null)
```
gonX, line 156:
```suggestion
                && Tablet?.Properties?.Specifications?.Pen != null)
```
might as well do this

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -131,7 +131,7 @@ protected override void OnOutput(IDeviceReport report)
             if (report is ITiltReport tiltReport && Pointer is ITiltHandler tiltHandler)
                 tiltHandler.SetTilt(tiltReport.Tilt);
             if (report is ITabletReport tabletReport && Pointer is IPressureHandler pressureHandler
-                && Tablet.Properties.Specifications.Pen != null)
+                && Tablet != null && Tablet.Properties.Specifications.Pen != null)
```
gonX, line 134:
```suggestion
                && Tablet?.Properties?.Specifications?.Pen != null)
```
as above

gonX (commented):
Almost same fix done in #3691, so this LGTM (it is 1:1 after the included suggestions).

In general, enabling nullable for the entire project should help us see these more often, see #3687

OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
Mrcubix: `Tablet?.Overkill.Overkill.Pen != null`

gonX: Well, check warnings will continue if you don't do it this way. But approving anyway since the remaining check warnings will be fixed in #3691.

Mrcubix: What warning (there are none caused by me), also they don't happen locally, i have also noticed found many issues with C# & github's CI yesterday alone.
Btw you shouldn't make projects Nullabke if they are not built with it in mind.

PR #2063, opened 2022-02-12
Add support for Huion H642

jamesbt365:
Verification: https://discord.com/channels/615607687467761684/723399565780189234/942034357990850601

PR #3650, opened 2024-12-29
Add support for new firmware of RTS-300

jamesbt365:
Diagnostics: https://discord.com/channels/615607687467761684/615611007951306863/1322926818722054164
Verification: https://discord.com/channels/615607687467761684/615611007951306863/1322926361957896253

### DIFF
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -13,7 +13,7 @@
 | Gaomon PD1561                 |     Supported     |
 | Gaomon PD2200                 |     Supported     |
 | Gaomon S620                   |     Supported     |
-| Genius G-Pen 560              |     Supported     | Soft-buttons are bindable as aux buttons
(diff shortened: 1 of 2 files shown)


TABLETS.md
```diff
@@ -48,6 +48,7 @@
 | Huion Q11K                    |     Supported     |
 | Huion Q11K V2                 |     Supported     |
 | Huion RDS-160                 |     Supported     |
+| Huion RTS-300                 |     Supported     | Uses the same configuration as the Huion H642.
```
gonX, line 51: is this convention? I see the strings overlap - but would a separate config still not be preferable?

jamesbt365: A separate config without finding a way to identify them would cause a double detection and as they internally are the same tablet i don't think its a problem that they share.

gonX: oh, apparently it was always like this too (thanks james): #2063

gonX (approved):

PR #3656, opened 2024-12-29
nix: Update version to 0.6.5.0

gonX:
Also worth noting that the current way the NixOS actions are set up, updating `deps.nix` is seemingly impossible without having a working Nix install.


InfinityGhost:
For some reason despite allowing maintainer edits I can't push to your branch so here's the patch with updates

[flake-update.patch](https://github.com/user-attachments/files/18271874/flake-update.txt)

gonX:
> For some reason despite allowing maintainer edits I can't push to your branch so here's the patch with updates
> 
> [flake-update.patch](https://github.com/user-attachments/files/18271874/flake-update.txt)

Added. You forgot the git commit header so I modified attribution to you.

InfinityGhost (approved):

PR #3657, opened 2024-12-30
Fix tilt for Huion H1061P

ymanton:
Use UCLogicTiltReportParser instead of UCLogicReportParser.

Confirmed on my tablet that tilt works with this change.

Thanks.


jamesbt365 (approved):

gonX (approved):
Yeah I remember this conversation. LGTM.

PR #3658, opened 2025-01-02
Add wireless support for Wacom CTC-6110WL

nahkd123:
Add wireless (Bluetooth) support for Wacom CTC-6110WL. Wireless variant have product ID of 259 when connected using Bluetooth, comparing to product ID 258 when connected using cable.

The tablet when connected using Bluetooth always send a report every second while idling, which is why I changed `IntuosV3ReportParser` to prevent the cursor from teleporting to top-left corner.

Confirmed working with my CTC-6110WL on Windows 11. I don't have CTC-4110WL to test wireless.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/CTC-6110WL.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/CTC-6110WL.json
@@ -21,6 +21,15 @@
       "FeatureInitReport": [
         "AgI="
       ]
+    },
+    {
+      "VendorID": 1329,
+      "ProductID": 259,
+      "InputReportLength": 19,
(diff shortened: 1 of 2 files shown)


Kuuuube:
That report parser change looks sketchy and there might be a better solution. Send a recording of tablet data with pen out of range and in range from debugger. And since youre adding a new identifier a normal diag would be good too.

nahkd123:
Diagnostic info when the tablet is connected through Bluetooth: [diagnostic.json](https://github.com/user-attachments/files/18304705/diagnostic.json)

Tablet data recording: [tablet-data.txt](https://github.com/user-attachments/files/18304706/tablet-data.txt)

Kuuuube (approved):
Looks good. Confirmed this doesn't conflict with anything on the wired version and the DTH-135 (not supported yet) that also needs this parser reports `data[1] == 0x01` too.

PR #3662, opened 2025-01-06
Fix the problem debugging on Macos requiring app bundle

muchitto:
When debugging by launching the UX.Mac project in Rider, it throwed an error about needing to be an app bundle. This occured only when providing the platform with a string instead of the Eto.Platform class.

I replaced these with the actual platform instance.

### DIFF
--- a/OpenTabletDriver.UX.Gtk/LinuxApp.cs
+++ b/OpenTabletDriver.UX.Gtk/LinuxApp.cs
@@ -9,7 +9,7 @@ namespace OpenTabletDriver.UX.Gtk
     {
         private Process? _daemon;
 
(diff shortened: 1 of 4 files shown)


AkiSakurai:
You can also change the Exe path in the Rider configuration to point to the executable inside the app bundle.
Screenshot 2025-01-07

muchitto:
Sure. I think it's still better to not have to create a bundle just to debug?

jamesbt365:
Thanks for your contribution, however we have just feature locked the `master` branch and all contributions should target the `avalonia` branch instead, however because we switched the UI framework this PR doesn't apply there so I'll be closing this PR.

Thanks for this change regardless and we hope to see more from you in the future!

Issue #2992, opened 2023-11-07
Incorrect auxiliary button behavior for Wacom CTE-440


LostInSwiss:
### Steps to Reproduce

The Wacom CTE-440 has three auxiliary buttons on the tablet. The middle button is a scroll button. They almost work but not quite.

In the tablet debugger the last two values of the raw tablet data relate to these buttons but the behavior of the tablet report Aux Buttons is incorrect and cannot be used for binding.

Below is a table of the button, the last two hex digits of the tablet data, the mapping with the button and what is reported in AuxButtons and a comment on the strange behavioure.

Button, Hex value, AuxButtons
1 left button, 40,  [False,False,False,True] - This is the only button that works correctly. True when pressed, all clear to false when depressed
2 scrl up = 08, [True, False, False, False] - This sort of works but does not clear to FFFF when depressed
3 scrl down = 38, [True, True, True, False] - This incorrectly sets three buttons to true and does not clear them when depressed 
4 right button = 80, This button will only clear all values to False if one is stuck on True, it will not set any buttons to true.

The issue is inputs do not set buttons to true and one button sets three buttons to true. The resulting behavior is only the left aux button works correctly, scroll up only fires a button once (until cleared by another), scroll down fires three buttons at once (and wont clear on depress), and the right button does nothing at all.

While the UI does not show 4 aux buttons (it should) I was able to add 2 more by editing the json settings file.

### Expected Behavior

Each button should trigger a single aux button when pressed and clear the aux buttons to false when depressed.

### Observed Behavior

The issue is inputs do not set buttons to true and one button sets three buttons to true. The resulting behavior is only the left aux button works correctly, scroll up only fires a button once (until cleared by another), scroll down fires three buttons at once (and wont clear on depress), and the right button does nothing at all.

### Tablet Device

Wacom CTE-440

### Diagnostic Information

[diag.json](https://github.com/OpenTabletDriver/OpenTabletDriver/files/13273700/diag.json)

### Acknowledgements

- [X] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [X] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.
- [X] I am on the latest version of OpenTabletDriver.

[...]

PR #3673, opened 2025-01-12
Candidate fix for Wacom Graphire AUX buttons.

mjh65:
To address https://github.com/OpenTabletDriver/OpenTabletDriver/issues/2992

This fix has been tested with a CTE-430 (no aux buttons) and a CTE-440 (2 buttons and scroll wheel). Configuration for models CTE-630, ET-0405* have also been updated but not tested.

Mapping of scrollwheel to AuxButtons report has been removed, but affected raw bits are noted in the source.

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Wacom/Graphire/GraphireTabletReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Wacom/Graphire/GraphireTabletReport.cs
@@ -1,0 +1,48 @@
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using OpenTabletDriver.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Wacom.Graphire
+{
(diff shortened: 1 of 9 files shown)


OpenTabletDriver.Configurations/Parsers/Wacom/Graphire/GraphireTabletReport.cs
```diff
@@ -0,0 +1,48 @@
...
+            var auxByte = report[7];
+            AuxButtons = new bool[]
+            {
+                auxByte.IsBitSet(6),
+                auxByte.IsBitSet(7),
+            };
+
+            // wheel = report[7][5:3]
+
+            HighConfidence = report[1].IsBitSet(7);
```
gonX, line 37: Most Wacom tablets that set this also set hover distance, see #2556. Does Graphire not emit hover distance?

gonX (changes requested):

PR #3674, opened 2025-01-12
Add support for Gaomon M10K (2018) tablet

thegeek95:
Add support for Gaomon M10K (2018) tablet.
The 2018 version uses a productID of 111.
I have tested this config with the physical device and verified the MaxX, MaxY, and MaxPressure. via the tablet debugger.


jamesbt365 (commented):
Our CONTRIBUTING.md states that all configuration additions must include a diagnostic showing the tablet working with the config.

Please export diagnostics (help > export diagnostics) and upload that file here.

thegeek95:
[tabletDiagnostics_final.json](https://github.com/user-attachments/files/18390075/tabletDiagnostics_final.json)

Diagnostics attached.

jamesbt365:
Could we have a copy of the string dump too?

thegeek95:
[stringDump.txt](https://github.com/user-attachments/files/18390083/stringDump.txt)

I have attached the string dump. Most of it appears unreadable to me.

jamesbt365:
Not sure if everyone really wants this, but we already have variants of the 2018 model under the m10k configuration, to keep things simple (and so we don't have to backtrack and find out which is which, even though it doesn't matter) could we merge the configurations and you just put this identifier inside the m10k config?

gonX (changes requested):
The implemented tablet exactly matches the specifications of Gaomon M10K. You should include the identifiers specified here in that file instead.

gonX:
Also, configuration PR's needs to target the `avalonia` or `0.6.x` branch.

gonX:
`master` branch is EOL so I am going to close this stale PR.

PR #3678, opened 2025-01-14
Fix XP-Pen Deco 01 V2 (Variant 2) detection

tujh2:
Hi! Bought this tablet (XP-Pen Deco 01 V2) in 2024 and discovered that tablet area size is wrong (smaller). I'm playing osu!lazer that uses OpenTabletDriver and it has this issue too: https://github.com/ppy/osu/issues/21888

Also you can check my USB Strings on this screenshot.
telegram-cloud-photo-size-2-5204288070034257058-x

After my commit it detects my tablet as (Variant 2) configuration which has correct MaxX and MaxY.


jamesbt365:
You don't touch the 01 V2 config, just the Variant 2 config, adjusting the Variant 2 config won't make it stop detecting as just the 01 V2. 

Are you sure its not double detecting the tablet?

Even still, a diagnostic is a requirement for all configuration tweaks (as per the CONTRIBUTING.md), so please upload the file generated from help > export diagnostics.

tujh2:
> You don't touch the 01 V2 config, just the Variant 2 config, adjusting the Variant 2 config won't make it stop detecting as just the 01 V2.
> 
> Are you sure its not double detecting the tablet?

Actually I don't know. I guess it find only one configuration with this USB strings. 

> Even still, a diagnostic is a requirement for all configuration tweaks (as per the CONTRIBUTING.md), so please upload the file generated from help > export diagnostics.

Yep, there are two files - with fix and on current master branch.
[diagnostics(fix).txt](https://github.com/user-attachments/files/18416504/diagnostics.fix.txt)
[diagnostics(master-branch).txt](https://github.com/user-attachments/files/18416505/diagnostics.master-branch.txt)

jamesbt365:
Could you test on the latest version with a configuration override? I trust 0.6.x much over any build based on master, as master can be a bit unpredictable at times and isn't held to the same standard as 0.6.x is.

You can create a folder called `Configurations` next to the binaries and drop in any configuration files you like and it'll read those instead of the inbuilt ones. Once you get it working there could you export diagnostics.

jamesbt365:
As expected, this fix is insufficient and causes multiple instances of this tablet to be detected.

You will need to modify the 01 V2 configuration to not match your variant without making detection worse.

tujh2:
Good news. Found device string reader tool in OpenTabletDriver GUI. This is dump from my tablet. And found new device string at index 5! New fix should not affect on other configurations. But I guess there is not only 2023-year revision of this tablet, so i guess this will not fix all Deco 01 V2 tablets :)

[device-strings.txt](https://github.com/user-attachments/files/18417317/device-strings.txt)
[diagnostics(release-fix-config-v3).json](https://github.com/user-attachments/files/18417312/diagnostics.release-fix-config-v3.json)

tujh2:
also forgot to escape `.` character in json-string regex. (`.` means any character, so we need`\.`, also because of json we shold also escape `\`, so `\\.`). Fixed this too (it does not matter in general, but this would be correct)

tujh2:
Actually I think all new revisions after 2020-10-23 should use (Variant 2) configuration. So I added this via regex and I hope this will fix all revisions of Deco 01 V2

[diagnostics(release-fix-config-v4).json](https://github.com/user-attachments/files/18418720/diagnostics.release-fix-config-v4.json)

TecnoCreeper:
I just noticed that while this fixes the issue with the area, the pen buttons binding disappears from the GUI. The buttons are still correctly detected from the tablet debugger though.

mm-sorel:
Thank you, @tujh2. My XP-Pen Deco 01 V2 is now correctly identified as Variant 2, which fixed the issue I had; before this, I could only use a quarter of the tablet's active area.

Doesn't the regex modification precludes 01 V2 from matching with Variant 2? I understand it's yet untested with an actual physical device, but I'm not sure how someone with a working 01 V2 is going to find their way here to test it. In the meantime, all Variant 2 devices remain broken.

tujh2:
> Doesn't the regex modification precludes 01 V2 from matching with Variant 2?

I assume that all revisions after 2020-10-23 should use the (Variant 2) configuration, and I also assume that there are no revisions in 2020-11-xx and 2020-12-xx, but this is just a hypothesis.

gonX (changes requested):

tujh2:
> Configuration PR's should be targeted for `0.6.x` or `avalonia` branch.

updated the branch and target

TecnoCreeper:
> We'd likely at the very least need a diagnostics file and a string dump from both types of tablets.

I updated today to the latest release. Here my string dump and diagnostics with my `Deco 01 V2 (Variant 2)` that is not being correctly detected:

[dump.txt](https://github.com/user-attachments/files/22642541/dump.txt)
[diagnostics.json](https://github.com/user-attachments/files/22642542/diagnostics.json)

tujh2:
> It looks like #4293 has made some changes to the configurations of these tablets. Please double check that this PR is still needed - if not, sorry for missing your change :(

Yes, it now handles "2023-08-07_Release" (which is my tablet and @TecnoCreeper's as well).
It's being detected as Variant 2 on the 0.6.x branch. Closing this PR as fixed.

image

TecnoCreeper:
Maybe I misunderstood or did something wrong, but it's still not working for me.
I cloned the repo, checked out the 0.6.x branch, ran build.ps1 on Windows (I tried building on WSL but the gui app errored and crashed on startup), started the gui app; and the tablet still has the wrong area.

EDIT: I copy pasted the json files from #4293 to the Configurations override folder (using the stable release) and it works correctly.

mm-sorel:
> Yes, it now handles "2023-08-07_Release" (which is my tablet and @TecnoCreeper's as well).

Unfortunately, it seems I have a "2023-11-24_Release."


image


#4293 does not work for me, @gonX, but this PR did. Could it still be merged?

gonX:
Hello @mm-sorel feel free to open a new issue with your string, otherwise your issue will very likely be lost

PR #3681, opened 2025-01-16
Add linux support to Huion H610 Pro V3

jamesbt365:
As WinUSB changes the identifiers to 16, the current config doesn't work on other platforms than Windows.The correct interface is grabbed from the TABLETS.md, so this change shouldn't break anything.


gonX (approved):
LGTM

PR #3689, opened 2025-01-18
[0.6.x] Switch deps.nix to deps.json

InfinityGhost:


gonX (approved):
Seems harmless, so LGTM. Didn't inspect deps in depth, but this seems to only affect nix anyway.

PR #3690, opened 2025-01-18
[master] Switch deps.nix to deps.json

InfinityGhost:


gonX (approved):
LGTM, same comments as 0.6.x review (#3689)

Issue #3687, opened 2025-01-18
[0.6.x] Follow up on checks after Plugin subproject has nullable enabled


gonX:
#3613 added `nullable` for the Plugin subproject (later retracted in #3781). This adds a bunch of warnings when building in regards to maybe-null values. 

TODO: explain why its good to have it enabled

Reading:
https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references
https://learn.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies

Initial proposal of stages:
- PR that Cover all of the checks in Plugins subproject
- PR that enables nullable for other subprojects and/or everything in one go.
  - Implementers: try to keep the PR's separate between csproj's, to simplify reviewing.

The milestone is loose, but would be nice to have in for at least 0.6.6, if not 0.6.5.1.

Will update with more at a later point - comments or questions are welcome.

PR #3691, opened 2025-01-19
[0.6.x] Address nullability warnings

gonX:
Addresses some aspects of #3687

Some Resharper checks annotations sneaked in - this was unintended but seems harmless to keep in

This PR is exclusively intended to address the current nullability warnings thrown on PR's. Further code quality changes (if any) will go in a separate PR.

Could technically throw almost all labels on given how much code this PR touches, but the intention is that the code should behave (almost) exactly the same as before - in cases where it doesn't behave the same as before, it is unintentional and should be fixed before merge.

### DIFF
--- a/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
@@ -12,12 +12,12 @@ namespace OpenTabletDriver.Plugin.Output
     public abstract class AbsoluteOutputMode : OutputMode
     {
         private Vector2 min, max;
-        private Area outputArea, inputArea;
+        private Area? outputArea, inputArea;
 
         /// <summary>
         /// The area in which the tablet's input is transformed to.
         /// </summary>
-        public Area Input
+        public Area? Input
         {
             set
             {
@@ -30,7 +30,7 @@ namespace OpenTabletDriver.Plugin.Output
         /// <summary>
         /// The area in which the final processed output is transformed to.
         /// </summary>
-        public Area Output
+        public Area? Output
         {
             set
             {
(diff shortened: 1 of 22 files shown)


OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs
```diff
@@ -117,8 +117,10 @@ protected static Matrix3x2 CalculateTransformation(Area input, Area output, Digi
         /// Transposes, transforms, and performs all absolute positioning calculations to a <see cref="IAbsolutePositionReport"/>.
         /// </summary>
         /// <param name="report">The <see cref="IAbsolutePositionReport"/> in which to transform.</param>
-        protected override IAbsolutePositionReport Transform(IAbsolutePositionReport report)
+        protected override IAbsolutePositionReport? Transform(IAbsolutePositionReport? report)
         {
+            if (report == null) return null;
```
gonX, line 122: the real fix is probably to make `report` arg not nullable? still needs to keep null return though

OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
```diff
@@ -53,6 +54,8 @@ public float Frequency
             set
             {
                 this.frequency = value;
+                if (Scheduler == null) return;
```
gonX, line 57: this logic change seemed harmless to avoid NRE, in the very unlikely edge case that frequency is set before a scheduler is available

OpenTabletDriver.Plugin/Output/OutputMode.cs
```diff
@@ -104,18 +106,18 @@ public virtual TabletReference Tablet
             get => this.tablet;
         }
 
-        public virtual void Consume(IDeviceReport report)
+        public virtual void Consume(IDeviceReport? report)
         {
             if (report is IAbsolutePositionReport tabletReport)
                 report = Transform(tabletReport);
-            if (report != null)
+            if (report != null) // TODO: remove this null check, make sure null check is performed outside
```
gonX, line 113: was hesitant to remove this null check in case output mode plugins relied on this behavior, can remove it upon advisory

gonX:
needs to have another pass over, some things were probably made nullable that don't need to yet

gonX:
Going to move this to the backlog because it doesn't effectively change much and I think the time fixing this up is better spent elsewhere. People can use this PR as an inspiration of where we might need to do some null checks, but some of them might not be necessary.

PR #3695, opened 2025-01-19
Fix eraser tool state getting stuck in libinput

### LINKED ISSUE #3534: pen eraser stays enable
### Steps to Reproduce

OTD v0.6.4.0

open xournalpp
use the back of the pen (eraser)
try using the pen to write, it doesn't work and switch back to the eraser (binded to move in the video below)

https://github.com/user-attachments/assets/f43d364d-c763-47c8-b574-5dcfeb4c6729

### Expected Behavior

I expect the pen not to switch to the eraser

### Observed Behavior

I've tried closing xournalpp but it gets stuck with the eraser and I have to restart OTD for it to work again.

I've tried using 0.7 but I get an error about parsing. Should I open another issue for this ?

### Tablet Device

Wacom PTH-660

### Diagnostic Information

[tablet-data.txt](https://github.com/user-attachments/files/17837497/tablet-data.txt)

### Acknowledgements

Kuuuube:
Fixes #3534 

Tested on Arch with `libinput 1.27.1-1`. Haven't checked master to see if the issue is there as well.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -91,6 +91,9 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
         public void SetPosition(Vector2 pos)
         {
+            // Reset opposite tool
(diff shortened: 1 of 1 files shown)


gonX:
I don't like this approach*, so I've made a competing PR at #3700.

*to explain, behavior seems less defined when you send a "tool X is not present also tool Y is present" packet every time, rather than just sending "tool X is not present" once when we know that tool Y will be coming up.

PR #3698, opened 2025-01-21
OutputMode: Refactor `Elements`

gonX:
This cleans up the code to be slightly more intuitive. It's pretty difficult to read longer negated booleans in if statements, at least IMO. I tested with both pre- and post-transpose filters, as well as async filters.

This seems to slightly change behavior - previously, link/unlink operations only ran if there was something to link/unlink, but the end result seems to be exactly the same. **Please try to break this.**.

I originally had an `if (ElementsAsPipeline.Count > 1)` (since `this` is always in the list) in place before the link/unlink operations to match the original behavior, but I don't think the original behavior was intended, so I kept this out.

Another change that this PR does is that previously, an `IPipelineElement<IDeviceReport>` plugin could exist that was neither pre- nor post-transpose (but instead `None`) which would break the pipeline linkage, since the elements would be set, but no links would need to be set (`Elements` > 1 but no Post/Pretranspose plugins, so no linkage performed), and yet the passthrough was disabled, so `entryElement` ended up being null, effectively unsubscribing the `PipelineManager` (`OutputMode`). This PR solves that (but it wasn't really an issue to begin with, since the Binding Handler is a post-transpose plugin).

(not sure which label to file this under)

TL;DR:
Cleaner code, slight behavior change which I think is more intuitive, end effect should be no different (but please try to prove me wrong)

### DIFF
--- a/OpenTabletDriver.Plugin/Output/OutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/OutputMode.cs
@@ -41,8 +41,11 @@ namespace OpenTabletDriver.Plugin.Output
             get => this.passthrough;
         }
 
-        protected IList<IPositionedPipelineElement<IDeviceReport>> PreTransformElements { private set; get; } = Array.Empty<IPositionedPipelineElement<IDeviceReport>>();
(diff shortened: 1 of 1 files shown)


Kuuuube (approved):
Seems solid from my testing. I don't see any way this would break anything either.

PR #3699, opened 2025-01-21
[06x] Daemon: Properly output filters to log in development edge case

gonX:
This has no real change for releases, but helps with debugging without the bindingHandler in place since it counts as a filter.

## Pre-PR, if you have the binding handler removed:

Plugins do not show up in the filter when using exactly 1 plugin

## Post-PR, if you have the binding handler removed:

Plugins show up in log output regardless of filter count, unless no filters are enabled.

---

Mostly harmless code change, so targeting v0.6.5.1


Kuuuube (approved):
Looks good

PR #2503, opened 2022-11-19
Linux Artist Mode: Ignore proximity

gonX:
During behavior matching validation with the stock Wacom driver it was revealed that going to the edge of proximity has no influence on the values sent to userspace; "pen out" only occurs when it actually leaves the digitizer area.

This has the side effect of improving usable hover distance on affected tablets

This supersedes 1a09505e4cb1e1f633fe63e2a14d0c8965556663 from #1984

(please test, the changes were simply rebased to the current codebase)

PR #3703, opened 2025-01-22
Fix incorrect usage of proximity in libinput

Kuuuube:
Proximity in libinput is not the same as proximity in OTD. Libinput proximity is moreso equivalent to an OTD OutOfRangeReport (https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#handling-of-proximity-events). Libinput ignores all reports that do not have proximity set to 1 (true).

The default state for proximity is set (correctly) to true. Since OutOfRangeReport already runs the reset, SetProximity does not need to fire on an out of range report.

Users who need reports to be ignored when out of proximity (by OTD's definition) can use a plugin such as Hover Distance Limiter.


gonX:
As discussed on Discord I would rather just fix the improper implementation as was done for master in #2503. Closing in favor of a backport of that.

PR #3704, opened 2025-01-22
[06x] evdev: Dispose device when no longer needed

### LINKED ISSUE #3055: Driver does not disconnect old output modes when switching output mode
### Steps to Reproduce

(easiest to reproduce on Linux, since I don't know how to see device endpoints on Windows)

- Change output mode to something other than what you were using (e.g. Absolute Mode -> Artist Mode)
- Apply (or save)

### Expected Behavior

`libinput list-devices` does not contain the now-unused devices

### Observed Behavior

`libinput list-devices` still contains the now-unused OpenTabletDriver endpoints, e.g.:
```
$ libinput list-devices | grep -i opentabletdriver
Device:           OpenTabletDriver Virtual Artist Tablet
Device:           OpenTabletDriver Virtual Tablet
```

Worth noting that the "old" device is reused if output modes are switched back, so this is a low priority issue.

### Tablet Device

PTH-851

### Diagnostic Information

on request, but it should be doing it on current master (ac5d79c)

### Acknowledgements

gonX:
This adds IValidatableDevice to almost all cached DesktopInterop classes, to ensure that they remain usable when disposed.
(advice on how to better do this would be appreciated, it ended up more ugly than I liked - can I avoid adding an interface?)

Fixes #3055

I think we also had an issue about keys being held when tablet is disconnected - can't find it, but this should fix at least mouse left click being held when device is disconnected.

Post-PR only the Virtual Keyboard remains connected, unsure how to dispose of it correctly.

### DIFF
--- a/OpenTabletDriver.Benchmarks/Output/NoopAbsoluteMode.cs
+++ b/OpenTabletDriver.Benchmarks/Output/NoopAbsoluteMode.cs
@@ -16,6 +16,8 @@ namespace OpenTabletDriver.Benchmarks.Output
             {
                 Position = pos;
             }
+
+            public bool HasValidDevice() => true;
         }
     }
 }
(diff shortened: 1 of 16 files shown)


gonX:
Converting to draft because I'm sure there's got to be a better way. IDisposable + dependency injection feels weird.

gonX:
This doesn't look like the best approach when we use DI. We should be disposing of it in DI instead.

PR #3705, opened 2025-01-22
Remove proximity handling from linux artist mode

Kuuuube:
Backport of [#2503](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/2503)

Fixes improper usage of proximity in libinput reporting.

Proximity in libinput is not the same as proximity in OTD. Libinput proximity is moreso equivalent to an OTD OutOfRangeReport (https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#handling-of-proximity-events). Libinput ignores all reports that do not have proximity set to 1 (true).

Users who need reports to be ignored when out of proximity (by OTD's definition) can use a plugin such as Hover Distance Limiter.


gonX (approved):
LGTM (if checks pass)

PR #3709, opened 2025-01-23
Avalonia fix tests

gonX:
This adds an `IAppInfo` interface to allow DI to properly inject `AppInfo` - this wasn't working properly before.

It also re-adds the plugin version compatibility test, which is likely going to be obsolete later. The test should be removed or changed if we end up using another method to decide on plugin compatibility.


gonX:
Ty for added changes @jamesbt365

PR #3714, opened 2025-01-23
[Avalonia] Fix daemon not launching (Update the rest of the daemon to properly use `IAppInfo`)

gonX:
I was likely misunderstanding the intention here when implementing IAppInfo, but daemon is not launching without this.


gonX:
Merging because it's needed to continue

PR #3718, opened 2025-01-24
[Avalonia] driver: Fix NRE for tablets with no auxiliary device identifiers

gonX:
Code was expecting a bit too much that all configs have auxiliary identifiers.


gonX:
Drafting, fix looks incorrect now that I'm debugging.

jamesbt365 (approved):

PR #3720, opened 2025-01-24
Add support for variant of Huion Q630M

### LINKED ISSUE #3464: Issue with Huion Q630M Device Name in Configuration
### Steps to Reproduce

Connect Huion Q630M tablet to the system.
Run OpenTabletDriver's diagnostic function.
Check the device name fields in the output compared to the configuration file.

### Expected Behavior

The tablet should be recognized without issues.

### Observed Behavior

The tablet is not properly recognized due to a possible mismatch between the ProductName and the Name field in the configuration file.

### Tablet Device

Huion Q630M

### Diagnostic Information

I am experiencing an issue with my Huion Q630M tablet while using OpenTabletDriver. Although this device is listed as supported, I have encountered problems when trying to configure it.

According to the diagnostic output, the device fields are as follows:

```
"ProductName": "Huion Tablet_Q630M",
"SerialNumber": "20200922",
"FriendlyName": "Huion Tablet_Q630M"
```
However, in the source code, the configuration file for the Huion Q630M lists the device name as:
```
"Name": "Huion Q630M"
```
It seems that there is a mismatch between the actual ProductName field ("Huion Tablet_Q630M") and the name in the configuration file ("Huion Q630M"). This might be the reason the driver is not properly recognizing the tablet.
[3.json](https://github.com/user-attachments/files/17004769/3.json)

### Acknowledgements

jamesbt365:
I don't have verification, but as the user has not complained or done anything else on that post, its safe to say that it works fine. And if it doesn't work properly then we can always revert if someone complains.

Fixes #3464, string dump and diag is in that issue.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/Q630M.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/Q630M.json
@@ -22,7 +22,7 @@
       "InputReportLength": 12,
       "ReportParser": "OpenTabletDriver.Configurations.Parsers.Huion.GianoReportParser",
       "DeviceStrings": {
-        "201": "HUION_T216_\\d{6}$"
+        "201": "HUION_T(216|22b)_\\d{6}$"
       },
       "InitializationStrings": [
         200


gonX (approved):
Tests pass which means I am happy. LGTM.

PR #3727, opened 2025-01-24
Cap pressure sliders at 99

### LINKED ISSUE #3654: Pressure sliders should probably cap out at 99 instead of 100
### Description

Currently, you can set the Tip/Eraser Pressure to 100, which will effectively disable clicks from ever propagating. Either the logic for this needs to change (so that 100% pressure still counts as a press), or the daemon and/or GUI should limit values to the range of 0-99.

Disabling clicks from propagating is not a necessary feature - you can just unbind the feature. While OTD (or at least v0.6.5.1) has pressure remapping and there is no other way of disabling the tip binding, #3564, when merged, allows users to disable this again.

### Acknowledgements

jamesbt365:
fixes #3654 for Avalonia.

### DIFF
--- a/OpenTabletDriver.UI.Library/Views/TabletView.axaml
+++ b/OpenTabletDriver.UI.Library/Views/TabletView.axaml
@@ -267,7 +267,7 @@
                   Grid.Column="1"
                   Grid.Row="0" />
               <tc:RangedInput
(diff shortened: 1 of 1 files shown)


gonX (approved):
Probably correct, will test with the other changes

PR #3729, opened 2025-01-24
[Avalonia] daemon: Wrap diagnostics call in Task.Run to satisfy async

gonX:
One less warning, and diag export still works.


gonX:
Quick merging because this should be harmless

Issue #3538, opened 2024-11-23
Non-tagged builds should set a VERSION_SUFFIX based on git info when possible


gonX:
### Description

Currently (at least on `0.6.x`), diagnostic and GUI title, when built via `./build.sh` (on Linux), does not display a version suffix.

Packaging scripts currently have support for setting a version suffix with `VERSION_SUFFIX`:
> ```
> $ grep -R VERSION_SUFFIX
> <snip>
> eng/windows/package.ps1:    "/p:VersionSuffix=$env:VERSION_SUFFIX"
> eng/lib.sh:VERSION_SUFFIX=${VERSION_SUFFIX:-}
> eng/lib.sh:OTD_VERSION="${OTD_VERSION_BASE}${VERSION_SUFFIX}"
> eng/lib.sh:    /p:VersionSuffix=${VERSION_SUFFIX}
> ```

My proposal involves:
- If git root is **not** found or `git` does not exist:
  - Do not set `VERSION_SUFFIX`
- If git root is found and `git` exists:
  - Use `git` to describe the version (`git describe --long --tags --dirty` or similar)

The solution is likely just to set `VERSION_SUFFIX` to the output of the following command, if `git` and `sed` exists:
```sh
git describe --long --tags --dirty | sed -E 's/^v((.\.)*.)-(.*)$/\3/'
# output:
#    187-g7a60a541-dirty
```
on my machine on `0.6.x`, on 7a60a541a (current `0.6.x`) with some file changes, the output is as above

TL;DR: `./build.sh` from `0.6.x` branch results in diagnostics and GUI title reading "0.6.4.0" - it should instead use git tagged versions. Build system (`eng/lib.sh`) already supports use of the env var `VERSION_SUFFIX`.

### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.
- [X] I am on the latest version of OpenTabletDriver.

gonX:
Fixed in `avalonia` branch with #3791, changing issue to track `0.6.x` specifically.

PR #3745, opened 2025-01-25
[06x] eng/lib.sh: Autodetect VERSION_SUFFIX when unset

gonX:
Detects correct version suffix based on git tags

Solves #3538 for 0.6.x

- [x] Linux Build
- [ ] Windows Build

Fixing Windows build is out of my skillset for now.

**Has bugs:**
- Assumes we're in a git root
- Assumes tags are exactly e.g. `v0.7.0.0` - does not support `v0.7.0.0alpha` or similar

### DIFF
--- a/eng/lib.sh
+++ b/eng/lib.sh
@@ -9,6 +9,25 @@ set -eu
 
 VERSION_SUFFIX=${VERSION_SUFFIX:-}
 
+# if unset, autodetect suffix from git
+if [ -z "$VERSION_SUFFIX" ]; then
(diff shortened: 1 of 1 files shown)


gonX:
Gonna move this to the next milestone instead. The change is mostly useful for development builds, and we get minimal diagnostics from development build on the 0.6.x branch.

gonX:
There are some bugs that need fixing with this implementation - a better one is done in #3791 which I expect to be backported to the stable (`0.6.x`) branch at some point.

PR #3750, opened 2025-01-25
[06x] RelativeOutputMode warn on reset spam

### LINKED ISSUE #1947: Relative mode no movement with very low reset times.
## Description

When reset time is set to 00:00:00 the cursor doesn't move, this should be either warned about or something should be put into place so that super low times cannot be allowed, see  [here](https://discord.com/channels/615607687467761684/615611007951306863/931609986394628116).
## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | N/A
| OpenTabletDriver Version | N/A
| Tablet                   | N/A

gonX:
Fixes #1947 for 0.6.x. Will warn the user if 10 consecutive resets have been hit. Unsure if this should error instead.

Alternative fixes could have been calculating a ratio, maybe >10%? It would help detect resets times that are marginally too low, especially in the case of 200hz tablets that alternate between 4 and 6ms reports (as seen in 200Hz USB1.1 Wacoms).

~~Enabling data validation in the UI seems like a larger undertaking: https://gist.github.com/cwensley/8348615c67618dc46102~~ We can actually just check in `ApplySettings()` if needed. Propagating it back into the UI is the hard part.

### DIFF
--- a/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
@@ -18,6 +18,10 @@ namespace OpenTabletDriver.Plugin.Output
         private Vector2 lastReadPos;
         private bool outOfRange;
 
+        // for handling detection of low resetTimes
+        private uint _resets, _passes;
+        private bool _warnedBadResets = false;
+
         /// <summary>
         /// The class in which the final relative positioned output is handled.
         /// </summary>
@@ -42,6 +46,7 @@ namespace OpenTabletDriver.Plugin.Output
         }
 
         private float rotation;
+        private TimeSpan _resetTime;
 
         /// <summary>
         /// The angle of rotation to be applied to the input.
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -84,7 +104,16 @@ public override void Read(IDeviceReport deviceReport)
...
                     outOfRange = true;
                     lastTransformedPos = null;
+                    _resets++;
                 }
+                else _resets = 0;
+
+                if (!_warnedBadResets &&
+                    (_warnedBadResets =
+                        _resets > 10))
+                    Log.Write("RelativeOutputMode",
```
jamesbt365, line 114: ditto
```suggestion
                    Log.WriteNotify("RelativeOutputMode",
```

OpenTabletDriver.Plugin/Output/RelativeOutputMode.cs
```diff
@@ -59,7 +64,22 @@ public float Rotation
...
         /// The delay in which to reset the last known position in relative positioning.
         /// </summary>
-        public TimeSpan ResetTime { set; get; }
+        public TimeSpan ResetTime
+        {
+            set
+            {
+                if (value == TimeSpan.Zero)
+                {
+                    Log.Write("RelativeOutputMode", "Reset time cannot be 0", LogLevel.Error);
```
jamesbt365, line 73: Probably a good idea to send a system notification to the user
```suggestion
                    Log.WriteNotify("RelativeOutputMode", "Reset time cannot be 0", LogLevel.Error);
```

gonX:
Apparently saving with a reset time of 0 resets your entire config, this will need fixing before merge.

gonX:
Thinking further about it, I am just going to tear out the zero resettime logic. We can revisit this later - the PR is already a significant improvement over existing behavior.

jamesbt365 (approved):

PR #3757, opened 2025-01-26
Revert Huion 420 parser change

Kuuuube:
Some Huion 420 tablets send something that breaks TabletReportParser. Revert of https://github.com/OpenTabletDriver/OpenTabletDriver/commit/2bfe9183f65d5840e5e2f0ac551075e2cee5a3aa.

Verification https://discord.com/channels/615607687467761684/615611007951306863/1328244031796285481


jamesbt365 (approved):
Would be nice to get some logging in place to see why this happens in the first place as the UCLogicReportParser is just the TabletReportParser with aux.

What black magic is going on here?

gonX (approved):
LGTM - I remember the change and was opposed to it, let's see if we can fix this properly instead.

PR #3761, opened 2025-01-27
[06x] Fix support for pressure sensitivity in Adobe software on macOS

AkiSakurai:
Adobe software requires specific Wacom information to recognize a tablet. We set this regardless, so events from the tablet get recognized by Adobe software.


gonX (approved):
Seems reasonable enough. Do you want this in for the bugfix release (0.6.5.1)?

AkiSakurai:
Yes, that works.

gonX:
Almost forgot about this one - it didn't make the release candidate but it will be going in 0.6.5.1 anyway.

PR #3763, opened 2025-01-27
Fill Empty screen with a Label when a tablet isn't detected or selected

Mrcubix:
When a tablet is connected & selected, Profile is always non-null.
A null profile probably either means a tablet isn't selected, or none are available.

image


Mrcubix:
Additionally, i could disable the tablet & Output Mode dropdowns if neccessary

Mrcubix:
Turns out there is already an existing Placeholder when no tablets are detected or selected, now setting the tabControl selected index approprietely to show that instead of the output mode screen.

gonX (approved):
Code doesn't look super silly, any weirdness should be getting caught in the upcoming rc anyway.

PR #3764, opened 2025-01-28
[06x] Fix: Prevent proximity event on mouse up for macOS

AkiSakurai:
Resolves an issue where a proximity event is incorrectly sent on mouse up, causing Wacom Inkspace app to fail in stopping the stroke after mouse release.


gonX (approved):
This probably makes sense. Targeting for 0.6.5.1.

PR #3769, opened 2025-01-29
Yank support for 10moons 1060N

jamesbt365:
Unfortunately, we have to remove support for this tablet as its hindering OpenTabletDriver's development and support for new tablets.

Parsing was originally broken because the user had a broken tablet, but this tablet is weird and hard to parse to begin with, it lies about every single spec so its just not something we can reliably have in the driver right now.

When people submit changes that fix the parsing, we can add this config file back as long as we change its identifiers to detect ONLY this tablet and not others.

There are numerous tablets that detect under this config that really shouldn't:

https://discord.com/channels/615607687467761684/723399565780189234/1328233971959595078 
https://discord.com/channels/615607687467761684/723399565780189234/1284314755699441758 
https://discord.com/channels/615607687467761684/723399565780189234/1270893651052068917 
https://discord.com/channels/615607687467761684/723399565780189234/1261241219674279937 
https://discord.com/channels/615607687467761684/723399565780189234/1228668011867668542 
https://discord.com/channels/615607687467761684/723399565780189234/1192868825952161833


OpenTabletDriver.Configurations/Configurations/10moon/1060N.json
```diff
@@ -1,44 +0,0 @@
...
-    "AuxiliaryButtons": {
-      "ButtonCount": 12
-    }
-  },
-  "DigitizerIdentifiers": [
-    {
-      "VendorID": 2290,
-      "ProductID": 26641,
-      "InputReportLength": 64,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.TenMoon.TenMoonReportParser"
```
gonX, line 23: Looks like this is the only config that was using this parser. Might as well remove that too.

jamesbt365: I guess, parser is the reason this tablet hardly work on the driver anyway so this can be yanked too as it doesn't work either.

Resolved.

gonX (approved):
LGTM now

Kuuuube (approved):

PR #3771, opened 2025-01-29
[Avalonia] Cleanup workflows

gonX:
This splits the build process into a matrix using composite actions. This should allow us to support additional packages or architectures by only adding to the matrix.

Some other notable changes:

- `actions/setup-dotnet` is now done in a composite action, this keeps usage DRY (though it does not cover everything, like .NET SDK's installed via `dnf` or `apt` invocations)
  - This can likely later be improved with another composite action that grabs the desired SDK version from OpenTabletDriver project files and then using this number in both the new composite action as well as in package manager invocations.
- Goes away with the custom invocation of `dotnet publish` and instead uses our build scripts
  - Our own build scripts use the same arguments as what we did pre-PR, hence why this PR does not "reimplement" the arguments
- Tags artifact names with `VERSION_SUFFIX` - this now helps everyone from testing the wrong packages because everything was essentially named 'opentabletdriver x64.zip'
- Udev rules runs as its own action, instead of as a part of the Linux build process
- Removed some useless `name` declarations, especially for checkouts
- Artifact uploads will now error if they aren't found (wasn't really a problem except for in dev situations)

Changing `release.yml` (for publishing new releases) is currently out of scope - testing it is difficult and I would rather do that in a later follow-up PR after this PR's changes have been tested for a bit.

Some notes to consider:
- The "git safe directory" call checks if it's container "fedora" and installs git using `dnf` prior to running it, ideally we should assume the container we're using already has a git. This can be fixed by using a custom container image.
- There's a TODO for supporting release/debug builds - currently it always builds in release mode (we also did this before). Unsure if we ever want to build in debug mode, so I didn't implement this fully. Line numbers should still be present in stack traces.
- Debian build likely doesn't pick up the VERSION_SUFFIX properly, meaning UI and daemon doesn't display the correct subversion. To avoid PR scope creep this should be solved outside of the PR.
- Windows build has been mostly untouched. This will be solved in a later PR.

Used documentation:
[docs.github.com - Running variations of jobs in a workflow](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow)
[docs.github.com - jobs.<job_id>.strategy](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategy)
idea for undefined container means no container used: https://github.com/orgs/community/discussions/46071
[docs.github.com - `jobs.<job_id>.steps[*].shell` bash works on Windows runners](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell)
[docs.github.com - Composite Actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action)
[docs.github.com - Accessing contextual information about workflow runs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs)
[docs.github.com - Reusing workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)


gonX:
Gonna clean this up a bit - notably, make the build output locations a bit more consistent across the actions

gonX:
Got a soft approve on Discord from jamesbt365, just going to run with it and see what happen.

PR #3773, opened 2025-01-30
[Avalonia] eng/macos: Disable Console, update UX -> UI

gonX:
MacOS package wasn't working without this (dunno if it is now, but seems necessary regardless)


gonX:
Merging because #3771 needs it

PR #3775, opened 2025-01-30
[Avalonia] Add arm64 builds

gonX:
Extends #3771

Adds arm64 as an example


gonX:
Closing, it was mostly a proof a concept. This might officially come later

PR #3777, opened 2025-01-30
[06x] workflows/dotnet: Run workflow on branch pushes too

gonX:
This ensures that newly merged changes still compile successfully


gonX:
Likely not important for `master` (feature freeze and all that) so just gonna remove the forward label now that #3787 exists.

PR #3779, opened 2025-01-30
[Avalonia] update executable name on macos

AkiSakurai:


gonX (approved):
LGTM. Didn't know we had had these.

PR #3782, opened 2025-01-31
[06x] workflows/release: Move NuGet deploy to its own step

### LINKED ISSUE #3617: Separate NuGet deploy workflow from artifact builds
### Description

Currently, the NuGet deployment workflow is run as a part of a tagged release together with artifact generation (building the releases). It should likely be run as its own action, in case it errors (like it did with v0.6.5.0) and we want to rerun it without affecting existing artifacts.

### Acknowledgements

gonX:
This should make re-running failed NuGet deploys easier

Fixes #3617 for 0.6.x

Changes are difficult to test so we'll have to keep an eye on whether this works properly when release happens.
The idea is that NuGet deploy should only be possible with a tagged release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,8 +66,18 @@ jobs:
           name: OpenTabletDriver.osx-x64.tar.gz
           path: ./dist/macos/*.tar.gz
 
+  nuget:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jamesbt365 (approved):

gonX:
Removing forward port tag because it is being done in #3791

Issue #3780, opened 2025-01-31
Unify build scripts across distros


gonX:
Currently, our build scripts duplicate a bunch of code. A minimal version of our BASH build scripts has been ported to Powershell to facilitate building on Windows - however this indirectly duplicates a significant amount of code across languages.

Additionally, our MacOS packaging script suffers the same issue in that it duplicates a lot of code between it and the Linux build.

Put briefly, the following tasks should be solved:

- [ ] BASH build pipeline needs to support Windows, and discourage/deprecate usage of Powershell scripts (`./build.ps1` and `eng/windows/package.ps1`)
  - Can keep Powershell scripts in the repo for compatibility for the rest of 0.6.x, but I hope to have them removed by 0.7.x unless a good reason to keep them pops up.
- [ ] `./eng/{linux,macos}/package.sh` should be unified and made crossplatform
  - Can still keep these `package.sh` files for backwards compatibility (e.g. by having them call a crossplatform-compatible packager), just avoid duplicating code needlessly.
- [ ] `./build.sh` should be crossplatform
  - macOS currently not supported
- [ ] `./build.sh` should support arguments defining which OS to target for
  - Can probably assume a lot by using the .NET RID (`linux-x64`, `osx-arm64`, etc.)

There seems to be consensus on the developer team that most Windows developer installs has BASH in some sort of capacity. This sounds to me like we can safely get rid of the Powershell scripts for building Windows, and move them to using our existing BASH build scripts instead.

Additionally, it is worth deduplicating code between MacOS and Linux builds, as the scripts are mostly identical, and both run in BASH.

gonX:
Fixed as well as we can with #4179 and #3791

Please see those pull requests for implementation details.

PR #3785, opened 2025-01-31
[Avalonia] eng/lib.sh: Autodetect VERSION_SUFFIX when unset

gonX:
Detects correct version suffix based on git tags

Port of #3745

Windows build is not relevant (yet) because the plan is to phase out the PowerShell scripts with #3780


gonX:
Merging because I want this in #3771

PR #3786, opened 2025-01-31
[Avalonia] Standardize release asset naming

gonX:
Forward port of #3762


gonX:
Quick merging because to ensure inclusion with #3771

PR #3787, opened 2025-01-31
[Avalonia] workflows/dotnet: Run workflow on branch pushes too

gonX:
Avalonia forward port of #3777

This ensures that the newly merged changes still compile successfully


gonX:
Quick-merging to allow consistent testing on Avalonia branch

PR #3788, opened 2025-02-01
[06x] Fix memory leaks on macOS

AkiSakurai:
The .NET runtime do not enables the autorelease pool by default.

This causes memory leaks in long-running processes that call Objective-C functions relying on the autorelease pool for memory management.

Although CGEventPost is not an Objective-C function, it internally depends on CFAutorelease to release locally created objects.

### DIFF
--- a/OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
+++ b/OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
@@ -6,6 +6,7 @@
     <NoWarn>VSTHRD100; VSTHRD101; VSTHRD110; VSTHRD200</NoWarn>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TieredPGO>true</TieredPGO>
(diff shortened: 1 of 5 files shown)


OpenTabletDriver.Daemon/OpenTabletDriver.Daemon.csproj
```diff
@@ -6,6 +6,7 @@
     <NoWarn>VSTHRD100; VSTHRD101; VSTHRD110; VSTHRD200</NoWarn>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TieredPGO>true</TieredPGO>
+    <AutoreleasePoolSupport>true</AutoreleasePoolSupport>
```
gonX, line 9: For anyone else concerned, this should only affect MacOS: https://learn.microsoft.com/en-us/dotnet/core/runtime-config/threading#autoreleasepool-for-managed-threads

gonX (approved):
LGTM

gonX:
Merging so that people have a chance to test before we do a release candidate.

PR #3790, opened 2025-02-01
Update to .NET 9.0

user202729:
Adapt from https://github.com/OpenTabletDriver/OpenTabletDriver/pull/3369

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
     <VersionBase>0.7.0.0</VersionBase>
-    <FrameworkLTS>net8.0</FrameworkLTS>
-    <FrameworkLatest>net8.0</FrameworkLatest>
+    <FrameworkLTS>net9.0</FrameworkLTS>
+    <FrameworkLatest>net9.0</FrameworkLatest>
     <LangVersion>preview</LangVersion>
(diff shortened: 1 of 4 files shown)


jamesbt365:
Please take a look at our CONTRIBUTING.md, PRs are not to be made directly to `master` at the moment.

gonX:
Please retarget this to the `avalonia` branch specifically. `0.6.x` will be staying on LTS.

Mrcubix:
No point in using .NET 9, as it doesn't bring anything of value to the table & isn't LTS.

user202729:
What happened in my case is Arch Linux latest version is net9, and putting target 8.0 wouldn't make it run under 9.0 for some reason.

I'll see what options are available.

jamesbt365:
> What happened in my case is Arch Linux latest version is net9, and putting target 8.0 wouldn't make it run under 9.0 for some reason.

Because dotnet runtime is not backwards compatible, but the *upstream* arch dependency issue has been solved if you update your system packages, so you can install 8 alongside 9.

user202729:
Should be https://archlinux.org/packages/extra/x86_64/dotnet-runtime-8.0/ https://archlinux.org/packages/extra/x86_64/dotnet-sdk-8.0/ I guess. I could try to improve `build.sh` or something to tell the user to install appropriate packages, but that's a different topic.

PR #3791, opened 2025-02-01
[Avalonia] Unify packaging build scripts

### LINKED ISSUE #3780: Unify build scripts across distros

gonX:
Primary function of the PR is code deduplication in build scripts and being able to have version name in Windows artifacts, e.g. `OpenTabletDriver-0.7.0.0-1227-g311b404a_win-x64`

Unified into Linux BASH build scripts:
- [x] MacOS
- [x] Windows

Fixes #3780:
- [x] BASH build pipeline needs to support Windows, and discourage/deprecate usage of Powershell scripts (`./build.ps1` and `eng/windows/package.ps1`)
- [x] `./eng/{linux,macos}/package.sh` should be unified and made crossplatform
- [x] `./build.sh` should be crossplatform
  - Cannot autodetect runtime/OS - see below. Always defaults to `linux-x64` if runtime is unspecified.
- [x] `./build.sh` should support arguments defining which OS to target for
  - can do `windows`, `mac` and `linux` to default to the x64 target for their respective OS. It's just an argument wrapper for `--runtime win-x64` etc.

Release CI results can be seen on my own fork:

Release CI: https://github.com/gonX/OpenTabletDriver/actions/runs/13103820441
Fork release: https://github.com/gonX/OpenTabletDriver/releases/tag/v0.7.0.0alpha

Known bugs/defects:
- Does break backwards compatibility for non-Linux packaging
  - Windows and MacOS building should be able to just `./build.sh` to start building.
- Seems impossible to make `./build.sh` autodetect platform
  - We can't use `dotnet --info` to `grep` for the RID. This still returns a distro-specific RID as of .NET 8.0 which doesn't work on Arch
- Debian build likely doesn't use VERSION_SUFFIX properly
  - This is caused by `VERSION_SUFFIX` not being sent to the Debian build process. It's building from source. It works with Fedora because we autogenerate those parts of the build assets.


gonX:
Tests passing, gonna merge..

PR #3796, opened 2025-02-04
[0.6.x] Improve config dir detection in tests

gepbird:
Previously some tests failed with `dotnet test OpenTabletDriver.Tests --configuration Release --runtime linux-x64`, because the current working directory was `bin/Release/linux-x64` directory instead of the usual `bin/Release`, and it couldn't find the configuration directory in the `../../../..` path.

This PR prioritizes reading the config directory from an environment variable already used in other files.

### DIFF
--- a/OpenTabletDriver.Tests/ConfigurationTest.cs
+++ b/OpenTabletDriver.Tests/ConfigurationTest.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 using DiffPlex.DiffBuilder;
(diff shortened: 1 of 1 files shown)


jamesbt365:
Hey, you should give our CONTRIBUTING.md a read, you should not be PRing to `master` at this time. Either target our stable branch (0.6.x) or our UI rewrite (avalonia)

gepbird:
Sorry about that, I only read the [Code Contribution](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/master/CONTRIBUTING.md?plain=1#L81-L82) section where it states that I should PR to master.

At the top I see it mentioning the 0.6.x branch, I'll target that in a sec.

gonX (commented):
Assuming the intent is to include support of the `OTD_CONFIGURATIONS` environment variable, wouldn't it be more interesting to instead change the existing `ConfigurationDir` variable to
```
Environment.GetEnvironmentVariable("OTD_CONFIGURATIONS")
                ?? Path.Join(ConfigurationProjectDir, "Configurations");
```

This would make this change a one, maybe 2 line change instead of this.

gonX:
You can alternatively use the [[CallerFilePath] attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerfilepathattribute?view=net-9.0) on a helper function to help figure out the root directory for the Configurations project.

gepbird:
> You can alternatively use the [[CallerFilePath] attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerfilepathattribute?view=net-9.0) on a helper function to help figure out the root directory for the Configurations project.

This is almost exactly what I was looking for, great find and thanks!

> Assuming the intent is to include support of the OTD_CONFIGURATIONS environment variable

My main intent is to make the tests pass when called with the extra argument `--runtime linux-x64` that we use for every [dotnet package in nix](https://github.com/NixOS/nixpkgs/blob/fa35a3c8e17a3de613240fea68f876e5b4896aec/pkgs/build-support/dotnet/build-dotnet-module/hooks/dotnet-check-hook.sh#L54-L68). I'll remove the env variable in favor of `CallerFilePath`, but we can keep it as a fallback if you think it could be useful later.

> This would make this change a one, maybe 2 line change instead of this.

I agree, I mostly changed it because those variables didn't have any other references and it was subjectively more readable to me.

gonX:
I liked the idea of also supporting `OTD_CONFIGURATIONS` but this is fine as well.

gonX (approved):

PR #3808, opened 2025-02-16
[06x] Fix occasional sticky modifier keys issue on macOS

AkiSakurai:
This commit addresses an occasional issue on macOS where modifier keys (e.g., Shift, Control) could appear "sticky" in certain scenarios. The problem occurred because OpenTabletDriver had to fetch the active modifier key status before sending a pointer event. If a modifier key was released after the status was checked but before the event was posted, the application would receive outdated modifier information, causing the sticky key effect.

To resolve this, we now rely on a system-level mechanism to include the event flags (including modifier key states) directly in the event. This approach is more reliable than manually fetching the modifier status, as it avoids the race condition entirely.


Kuuuube (approved):
Looks good.

Additional confirmation this works: https://github.com/OpenTabletDriver/OpenTabletDriver/issues/3956#issuecomment-3110131055.

Issue #3811, opened 2025-02-18
Unable to increase software tablet area height past tablet active area height when using rotation


Pickelgold:
### Steps to Reproduce

1. Turn your tablet 270 degrees on your desk.
2. In the OpenTabletDriver GUI, set the rotation to 270 degrees.
3. In OTD GUI, set the tablet area height to a value greater than your tablet's active area height but less than or equal to your tablet's active area width. For example, my CTH-480 has an active area of 152x95mm, so I set my tablet area height to 120mm.
4. In OTD GUI, set the tablet area width to a value less than or equal to your tablet's active area height so that it fits on your tablet. For example, I set my area width to 80mm.
5. You should see the area change within OTD GUI. Make sure that the area fits on your tablet's active area. Apply and save the change.
6. Close OTD GUI and reopen it. The tablet's area height (previously set to 120mm) has reverted to 95mm, even though 120x80mm area will fit on the tablet due to the 270 degree rotation.

### Expected Behavior

The tablet area would have stayed on 80x120mm, which would have fit on my tablet due to the 270 degree rotation. This is important as it would make it easier to have a multi-monitor setup with one monitor on top of the other where each monitor has a reasonable tablet area.

### Observed Behavior

The tablet area reverts to the tablet's active area height, i.e. 80x95mm, after apply and saving the change.

### Tablet Device

Wacom CTH-480

### Diagnostic Information

[diagnostics.json](https://github.com/user-attachments/files/18835554/diagnostics.json)

### Acknowledgements

- [x] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [x] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [x] I have searched the existing issues and this new issue is not a duplicate of any.
- [x] I have written a concise and meaningful title.
- [x] I am on the latest version of OpenTabletDriver.

Kuuuube:
Right click your area and turn off `Lock to usable area`.

Pickelgold:
Thank you, this solved the problem. I'll close this issue but if it is unintended behavior someone else can reopen it.

PR #3815, opened 2025-02-22
Only clamp width and height in `Lock to usable area` if rotation is 0 or 180

Kuuuube:
These make it annoying for users that rotate their area in some cases. Eg. #3811

In theory we could implement a way to shrink the area by exactly how much is needed to fit inside the tablet's usable area for any given rotation. But this feels really strange in practice like I'm fighting OTD to set the area and it's "randomly" changing things that weren't intended to be changed. So I think it's better to just let the user figure it out themselves.

The important thing with `Lock to usable area` is that users can easily set their area offsets exactly to the edges.

### DIFF
--- a/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
+++ b/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
@@ -252,10 +252,13 @@ namespace OpenTabletDriver.UX.Controls.Output
 
                 if (fullBounds.Width != 0 && fullBounds.Height != 0)
                 {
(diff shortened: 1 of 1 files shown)


gonX (changes requested):
This is just different from current behavior. I would prefer something that considers rotation, but I realize the general rotation handling isn't great either.
Rotating e.g. 45 degrees does not constrain the area properly even when moving around.

gonX:
After further discussion on Discord the PR should instead disable the behavior on non-symmetric rotations, e.g. if rotation is 0 or 180, clamp dimensions, otherwise do not clamp.

gonX (approved):
Tested working, good change.

PR #3837, opened 2025-03-08
Fix an exception when unplugging a tablet then closing a binding editor

Mrcubix:
The ItemSource is now null, so attempting to access it cause the exception.

### DIFF
--- a/OpenTabletDriver.UX/Controls/Generic/GeneratedItemList.cs
+++ b/OpenTabletDriver.UX/Controls/Generic/GeneratedItemList.cs
@@ -98,6 +98,10 @@ namespace OpenTabletDriver.UX.Controls.Generic
 
         protected virtual void Insert(int index)
         {
+            // Prevent an exception when a tablet is disconnected & a binding editor is closed
+            if (ItemSource is null || index >= ItemSource.Count)
+                return;
+
             var itemBinding = new DelegateBinding<T>(
                 () => ItemSource[index],
                 (v) => ItemSource[index] = v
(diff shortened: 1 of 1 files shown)


Mrcubix:
Encountered this a while back while testing plugins, didn't bother til i got that exception again.

OpenTabletDriver.UX/Controls/Generic/GeneratedItemList.cs
```diff
@@ -98,6 +98,10 @@ protected virtual void HandleCollectionChanged(object sender, NotifyCollectionCh
 
         protected virtual void Insert(int index)
         {
+            // Prevent an exception when a tablet is disconnected & a binding editor is closed
+            if (ItemSource is null || index >= ItemSource.Count)
```
gonX, line 102: the latter check (`index >= ItemSource.Count`) will definitely help in avoiding some exceptions but I feel like asking for an index that is higher than the count should warn the developer instead?
```suggestion
            Debug.Assert(index >= ItemSource.Count);
            if (ItemSource is null)
```

Alternatively you can make it fall back to the max, but that may have other side effects:
```suggestion
            index = (index >= ItemSource.Count) ? ItemSource.Count - 1 : index
            if (ItemSource is null)
```

(the suggested code might not compile or may have an off-by one, please test manually first)

Mrcubix:
Turns out it wasn't as straight forward, the core issue is from the binding.
This fixed the issue, but is it proper : 
```cs
var itemBinding = new DelegateBinding<T>(
    () => ItemSource != null ? ItemSource[index] : default!,
    (v) => 
    {
        if (ItemSource != null) 
            ItemSource[index] = v; 
    }
);
```

gonX (approved):
LGTM

PR #3839, opened 2025-03-12
Add getter for Settings instead of whatever in the hell that was

Mrcubix:
What the hell is that.
Who the hell thought this was a good idea.
How did whoever merged that PR thought it was okay.
Many questions, yet my 5 remaining cells cannot answer.


gonX (approved):
Makes sense to me

PR #3919, opened 2025-06-14
Update Gaomon 1060 pro maximum pressure level to 16k

daominh89:
My gaomon tablet has support for 16k pressure and the official website clearly says so too
image

Changes the string value MaxPressure to 16383

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/1060 Pro.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/1060 Pro.json
@@ -8,7 +8,7 @@
       "MaxY": 31750
     },
     "Pen": {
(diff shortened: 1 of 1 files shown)


jamesbt365:
This is not mergable and is a breaking change, not every tablet has this value so you can't just change it.

Find a difference in the strings of the old 8k and the new 16k and compare against that, using a new configuration file

daominh89:
I rebuilt it and compared the change in krita and fire aplaca, the pressure sensitivity works as I expected.

jamesbt365:
It works on *your* tablet, other peoples tablets that still have a max of 8k will only get *half* the pressure range, we can't break it for some people just to fix it for others.

PR #3964, opened 2025-07-25
Update Huion HS611 configuration

kostek001:
Small update to Huion HS611 configuration, specifying interface number

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/HS611.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/HS611.json
@@ -42,6 +42,7 @@
     }
   ],
   "Attributes": {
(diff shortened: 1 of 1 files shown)


kostek001:
Without specifying this OTD will randomly connect to one of the 3 interfaces of the tablet.

jamesbt365:
Export diagnostics

gonX:
Change makes sense to me based on the diagnostics. Will see what our config maintainer thinks.

Kuuuube (approved):
This seems fine. There is not full detail on the interface from the previous person with this tablet but they do also have two 12 endpoints so it must've worked by rng from them.

No details at all for where the first identifier came from.

PR #3999, opened 2025-08-26
[06x] Enhance handling of invalid settings

gonX:
This PR changes the behavior when encountering invalid settings.

This drops usage of `Setting`'s `Recover(FileInfo, Settings)` which was deemed to be mostly confusing in place of moving the file elsewhere.

Pre-PR:

- `master`/`avalonia` branch configurations are incorrectly parsed, causing weird issues (exception on startup with tablet plugged in, stack overflow when plugging in tablet while daemon active)
- Configs that wouldn't deserialize without an exception would still be attempted to be "populated" on top of a default configuration as introduced in #699 (which seems less relevant with #2808 in place)

Post-PR:

- Settings file now has a `Revision` field matching `avalonia`
- `avalonia` branch configurations are detected and rejected (by matching value `0.7.0.0` for `Revision` field in settings), improvements for this are welcome.
- Rejected/invalid configs are now moved to a new file (currently `settings_bak-${currentUTCtimeInEpoch}.json`)

TL;DR: Bad configs would sometimes magically work before. Now known bad configs never work and will be rejected.

The primary purpose of this PR is to ensure that the `opentabletdriver-git` AUR package can be migrated to 0.6.x without causing havoc in behavior

Please test thoroughly - this will likely go in 0.6.5.2

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -294,16 +294,36 @@ namespace OpenTabletDriver.Daemon
 
             var settingsFile = new FileInfo(AppInfo.Current.SettingsFile);
 
-            if (settingsFile.Exists &&
-                Settings.TryDeserialize(settingsFile, out var settings) &&
-                settings.Revision != "0.7.0.0")
+            if (settingsFile.Exists)
             {
-                await SetSettings(settings);
+                if (Settings.TryDeserialize(settingsFile, out var settings) &&
+                    settings.Revision != "0.7.0.0")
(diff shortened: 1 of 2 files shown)


Kuuuube:
Initial testing on this looks good. Handling of totally corrupt settings file, handling of an incorrectly written json file, handling of incorrect types in fields, and handling an avalonia settings file.

I'll test actually installing this over a master install from the aur later.

Kuuuube (approved):
Should be all good as far as I can tell. No problems testing ontop of a real `opentabletdriver-git` install from the aur.

PR #4000, opened 2025-08-27
[06x] DriverDaemon: Create settings file when none are detected

gonX:
Does not create a settings file if no tablet is detected. This is to prevent new users missing out on the Setup Wizard in case they start the driver with the tablet not being ready yet, e.g. misconfiguration or just "checking out the driver" before getting the hardware.

Probably fixes some existing issue


Kuuuube (approved):
Looks good. Worth noting this does not create a settings file if OTD is started with no tablets and then a tablet is plugged in.

PR #4006, opened 2025-08-28
[0.6.x] Actually implement MaxSupportedDriverVersion

Mrcubix:
This property was part of the plugin-repo table, but was never actually implemented.
This PR aims at implementing it and making it work


gonX:
Yeah I don't think this can be used in the daemon to begin with - plugins are not loaded with metadata present, if I remember correctly. Therefore I think this PR makes sense.

Kuuuube (approved):
Looks good, works. Testing branches with versions set for `Preset Bindings`: 
0.6.5.0 https://github.com/Kuuuube/Plugin-Repository/tree/test-max
0.6.5.1 https://github.com/Kuuuube/Plugin-Repository/tree/test-max-0.6.5.1
999.8.9.1 https://github.com/Kuuuube/Plugin-Repository/tree/test-max-greater

PR #4007, opened 2025-08-28
Fix Plugin manager Metadata Desynchronization

### LINKED ISSUE #720: OTD won't correctly refresh plugins when uninstalling and reinstalling plugins with the same filenames
## Description

OTD won't correctly refresh plugins when uninstalling and reinstalling plugins with the same filenames

To reproduce: 
1. Manually install any plugin under the same foldername and filename as a plugin in the plugin repo
(it can be an entirely different plugin with entirely different metadata as long as the foldername and filename are the same)
2. Uninstall the plugin with the plugin manager
3. Reinstall the plugin with the plugin manager

OTD now still shows the old plugin that isn't installed anymore and does not show the newly installed plugin

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System | Win10
| Software Version | 0.5.1 and 9ed806996b66fff557d9491f3a155a960c2768b7

Mrcubix:
This fixes an issue where even after updating the plugin, it wouldn't save the metadata to a file & the UX wouldn't load the updated metadata.

Closes #720

### DIFF
--- a/OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
+++ b/OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
@@ -49,7 +49,7 @@ namespace OpenTabletDriver.UX.Windows.Plugins
             );
 
             var updateableBinding = new DelegateBinding<bool>(
-                () => IsUpdated,
+                () => IsOutdated,
                 addChangeEvent: (e) => MetadataChanged += e,
                 removeChangeEvent: (e) => MetadataChanged -= e
             );
@@ -233,7 +233,7 @@ namespace OpenTabletDriver.UX.Windows.Plugins
             this.ParentWindow.Enabled = true;
         }
 
-        private bool IsUpdated => GetUpdatedMetadatas(PluginMetadataList.Repository, Metadata, CurrentDriverVersion).Any();
+        private bool IsOutdated => GetUpdatedMetadatas(PluginMetadataList.Repository, Metadata, CurrentDriverVersion).Any();
 
         private static IEnumerable<PluginMetadata> GetUpdatedMetadatas(PluginMetadataCollection repo, PluginMetadata Metadata, Version CurrentDriverVersion)
         {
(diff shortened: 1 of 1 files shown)


gonX (dismissed):
Code LGTM, will wait for @Kuuuube to test if this works as requested

Kuuuube (dismissed):
Looks like this is fetching the wrong metadata (not checking the supported versions?). Tested installing chatter exterm and got a version with metadata set at 0.5.0.0.

Kuuuube:
Gave it another test. Clicking the install button for a plugin in the plugin repo doesn't seem to do anything.

Updating a plugin gives `Attempted to load the plugin x when it is already loaded.` The old plugin is still held onto.

Uninstalling and installing a plugin does seem to work fine here without a restart.

Test case for updating: [Hover Distance Limiter.zip](https://github.com/user-attachments/files/22070027/Hover.Distance.Limiter.zip)

Mrcubix:
1. Can't reproduce the issue,
2. The message is normal as OTD Load the plugins multiple times, every single time you install or update one.

Mrcubix:
Try this one out, foubt it will be much different.

Kuuuube:
It looks mostly good now but the UX filters list (probably tools and output modes too?) doesn't update. Plugins install just fine. Plugins update fine on the daemon's side, files get replaced without any issue.

Test case to easily see this: 
[Hover Distance Limiter.zip](https://github.com/user-attachments/files/22080649/Hover.Distance.Limiter.zip)

`Hover Distance Limiter` with a radial follow dll inside it. When updated it should refresh the filters list to show hover distance limiter instead of still showing radial follow. If you only test this with a plugin update that doesn't add/remove/rename filters this pr should work fine in it's current state.

Mrcubix:
https://i.imgur.com/nbQDgoo.png
image

Doesn't happen on my side, list is updated.

Kuuuube:
Alright, I'll give it another test later and approve if I don't see any regressions. The list not refreshing isnt a regression since it was already broken anyways.

Mrcubix:
`AppInfo.PluginManager.GetChildTypes<TSource>()` returns the type, meaning it's not removed from somewhere

PR #4010, opened 2025-08-30
[06x] Adaptive Bindings

### LINKED ISSUE #2510: Contextual/Adaptive Bindings
As outlined in this [comment](https://github.com/OpenTabletDriver/OpenTabletDriver/pull/2505#issuecomment-1321199536), the effect of a binding should stay true to its name. Many users are confused when they realize that most pointer-related bindings are not actually compatible to other output modes. #2505 solves it in Linux by mapping pen tip and buttons event to mouse clicks. However, we do not want this behavior.

> Labeling both as a "mouse binding" is not right as it would cause confusion later on when a pen tip or button event is handled differently from an actual mouse click/button event.

What I propose instead is to have a set of "defaults of the defaults" binding that works whatever output mode is chosen, with the output mode's chosen function of the binding, without stating what the binding is for exactly.

- Adaptive pen tip
- Adaptive pen button 1
- Adaptive pen button 2
- Adaptive pen eraser tip

The underlying binding that these adaptive bindings invoke should also be exposed where supported.

| Adaptive | Windows Ink | Linux Artist Mode | Absolute Mode / Relative Mode
| --- | --- | --- | --- |
| Adaptive: Pen tip | Windows Ink: Pen tip | Linux Artist Mode: Pen tip | Mouse: Left click
| Adaptive: Pen button 1 | Windows Ink: Pen button 1 | Linux Artist Mode: Pen button 1 | Mouse: Right click
| Adaptive: Pen button 2 | Windows Ink: Pen button 2 | Linux Artist Mode: Pen button 2 | Mouse: Middle click |
| Adaptive: Pen eraser tip | n/a (automatically handled) | n/a (automatically handled) | n/a

This would give the user explicit control whether they want an adaptive binding that "just works" for any setup, or if they want a specific function from a specific output mode. In the future we may make use of the information that the binding is used on a foreign output mode to "spawn" the correct one and use it so the binding will still work well.

`IMouseButtonHandler` can be used as a base to implement the adaptive bindings.

gonX:
This adds a new type of bindings: Adaptive Bindings

Fixes #2510 for 0.6.x

# Description

See the issue for what "adaptive means", but the gist of it is that tip binding means left click, and barrel buttons send the signal that equates to a right or middle click respectively. E.g. on Linux Artist Mode a right click is `BTN_STYLUS2`.

These new bindings are intended to be used when switching between output modes. These functionally replace the Linux Artist Mode bindings. No more juggling profiles for default use cases when switching between output modes :tada: 

Additionally, daemon defaults now chooses these adaptive bindings as seen here:
Screenshot from 2025-08-30 at 03_01_27 946971502

# Technical

A new interface `IPenAction` has been added to avoid reusing `IMouseHandler` as originally suggested in the issue we're solving. Without that change, mouse bindings could still be picked up. To keep compatibility with existing output modes implementing `IMouseHandler`, this PR also decides that the first tablet pen button correlates to right click, and the 2nd to middle click.

Please test thoroughly on non-Linux platforms. The 3 primary modes (artist, absolute, and relative) have been tested on Linux, but a double check is always handy.

Plugin developers making their own output modes may want to add support for these bindings (e.g. Windows Ink). The adaptive bindings supports output modes that implement either `IPenActionHandler` or `MouseButtonHandler`, and are injected via DI.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -396,6 +396,9 @@ namespace OpenTabletDriver.Daemon
             if (pointer is IMouseButtonHandler mouseButtonHandler)
                 bindingServiceProvider.AddService(() => mouseButtonHandler);
 
+            if (pointer is IPenActionHandler penActionHandler)
+                bindingServiceProvider.AddService(() => penActionHandler);
+
             var tip = bindingHandler.Tip = new ThresholdBindingState
             {
                 Binding = settings.TipButton?.Construct<IBinding>(bindingServiceProvider, tabletReference),
--- a/OpenTabletDriver.Desktop/Binding/AdaptiveBinding.cs
+++ b/OpenTabletDriver.Desktop/Binding/AdaptiveBinding.cs
@@ -93,6 +93,7 @@ namespace OpenTabletDriver.Desktop.Binding
             { "Eraser", PenAction.Eraser },
             { "Button 1", PenAction.BarrelButton1 },
             { "Button 2", PenAction.BarrelButton2 },
+            { "Button 3", PenAction.BarrelButton3 },
         };
 
         private static string ActionToString(PenAction button) =>
--- a/OpenTabletDriver.Desktop/Binding/AdaptiveBinding.cs
+++ b/OpenTabletDriver.Desktop/Binding/AdaptiveBinding.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using OpenTabletDriver.Plugin;
+using OpenTabletDriver.Plugin.Attributes;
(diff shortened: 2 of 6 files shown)


gonX:
Rebased a comment out and force pushed. Also hopefully fixed the linting issue (and why doesn't Rider auto fix these for me??)

Kuuuube:
Looks like this is working perfectly on Linux. Still have Windows and MacOS to test.

Kuuuube:
Seems all good on Windows too. As mentioned, this does not work with Windows Ink plugin currently but there are no extra side effects. Exact same behavior as before this PR.

Kuuuube:
Confirmed working on MacOS by @TheSevenPens. No problems on all output modes and pen buttons.

Kuuuube (approved):

PR #4013, opened 2025-08-31
[0.6.x] Move addition of Daemon service to LoadPlugins

Mrcubix:
Didn't notice that `IDriver`'s instance was inserted in `LoadPlugins`.
Moved the addition of the daemon service there to maintain consistency.


gonX (approved):
Effectively no change but it's now together with the other injection, cleaning up the code. LGTM.

Issue #2957, opened 2023-10-07
Disconnecting a tablet device mid-drag causes the drag event to get stuck.


Kodehawa:
### Steps to Reproduce

Refer to: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1576

This was confirmed to be behavior caused by OTD itself.

### Expected Behavior

Drag gets released and the windows can be interacted with again.

### Observed Behavior

Compositor-dependent. On KWin you can't interact with only XWayland windows, on GNOME you have a midly-frozen desktop, on wlroots you're forced to ttyout and either kill otd daemon or reboot.

### Tablet Device

Wacom CTL-480

### Diagnostic Information

[otd-diag.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/12838476/otd-diag.txt)

### Acknowledgements

- [X] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [X] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

gonX:
Can confirm this happens on both master (35601541fb) and 0.6.3.0

PR #4017, opened 2025-08-31
[06x] Resolution Change Handler

### LINKED ISSUE #1143: Area maps incorrectly to the screen after disconnecting one monitor in a multi-monitor scenario
## Description
If I boot up my PC with 2 monitors attached, later disconnect one, the area mapped on the screen will mess up.
Same if I boot up with 1 monitor and attach another. 
For example, the tablet area will map to both screens side by side.

doing `systemctl --user restart opentabletdriver` fixes the issue for the time being.
OTD GUI also requires a restart to refresh the list of displays in the "set to display".

would be nice if this was automated somehow

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System | Arch Linux |
| Software Version | rolling release, everything is always latest |
| Tablet           | XP-PEN G430 |

### LINKED ISSUE #3055: Driver does not disconnect old output modes when switching output mode

gonX:
Monitor Layout changes (such as monitor addition/removal, or other layout/dimension change) is now handled

# Features Added

- Resetting the Daemon services without restarting the entire daemon
- Polling for resolution changes
- Correctly disposing services provided via Dependency Injection

# Issues Solved

- Fixes #1143
- Fixes #3055
- Likely provides a workaround for #3171 (re-applying settings should re-enable the tablet)
- Probably #2957 
- more?

# Known Defects that already existed, but may be more obvious now:

- Output mode coordinates are hardcoded and not assigned to a monitor, which means that the resolution change event may not remap the tablet as intended on some setups. But in cases of simply adding or removing a monitor that is rightwards or downwards from the origin, it should work as intended.

# Known Defects introduced by this PR

- `DeviceReader`'s may throw `ObjectDisposedException` on daemon reset, but this causes a harmless debug log message, as this exception is already specifically handled.

# Needs fixing before merge:

- [x] UI state on internal daemon restart is not properly synchronized (can be confusing for end users, but seems fixable)
  - Basically, the UI likely doesn't know that it needs to reset its internal state
- [x] Selected tablet is reset on resolution change, breaking UX
- [x] Suggested debounce of resolution changed event (only process the resolution change once readings stabilize), solved for example by ensuring that the desktop layout is "unchanged" following a resolution change event, before firing the internal reset.

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -51,8 +51,7 @@ namespace OpenTabletDriver.Daemon
             {
                 if (args.Additions.Any())
                 {
-                    await DetectTablets();
-                    await SetSettings(Settings);
+                    await ResetDaemon();
                 }
             };
 
@@ -88,9 +87,15 @@ namespace OpenTabletDriver.Daemon
                     return;
 
                 Log.Write(nameof(DriverDaemon), "Sleep detected...", LogLevel.Info);
+                await ResetDaemon();
+            };
+            return;
+
+            async Task ResetDaemon()
+            {
                 await DetectTablets();
                 await SetSettings(Settings);
-            };
+            }
         }
 
         public event EventHandler<LogMessage>? Message;
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -241,12 +241,12 @@ namespace OpenTabletDriver.Daemon
                     Log.Write("Settings", "Failed to apply settings. Attempted recovery. Some settings may have been lost.", LogLevel.Error, true);
                 }
 
-                Resynchronize?.Invoke(this, EventArgs.Empty);
                 return Task.CompletedTask;
             }
             finally
             {
                 InjectServices();
+                ForceResynchronize();
             }
         }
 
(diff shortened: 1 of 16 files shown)


Kuuuube (dismissed):
Gave this a quick test. When applying settings with multiple tablets plugged in, the ui switches to the first tablet. Huge problem for multi tablet use.

gonX:
This is ready to review again. After further consideration, I don't think a debounce event is needed, so I've checked it off on the task list. The code currently checks every 10 seconds:
https://github.com/OpenTabletDriver/OpenTabletDriver/pull/4017/files#diff-c3f59a7930111d1bad88529a695667c936a3092f271d2f1bf54d4dfe142f4cf1R18

gonX:
I think we're going to take a different direction with handling the internal services. The resolution change handler code itself can be reused later.

PR #4019, opened 2025-09-01
Cleanup Temp before plugin install

Mrcubix:
This partially fixes the issue mentionned in https://github.com/OpenTabletDriver/OpenTabletDriver/pull/4007#issuecomment-3242695614

Turns out the Temp Directory isn't cleaned up before a plugin install, resulting in more files being installed than intended.

Depends on #4007 being merged beforehand.
(for complete testing purpose)

### DIFF
--- a/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs
+++ b/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs
@@ -191,6 +191,8 @@ namespace OpenTabletDriver.Desktop.Reflection
 
             await metadata.DownloadAsync(sourcePath);
 
+            sourceDir.Refresh();
+
             var context = Plugins.FirstOrDefault(ctx => ctx.Directory.FullName == targetDir.FullName);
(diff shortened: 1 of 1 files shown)


Mrcubix:
How in the fuck does this even happen

Mrcubix:
Would it be possible to step by step debug & get the value of `sourceDir` & `targetDir` with a breakpoint at line 189 ?

https://github.com/Mrcubix/OpenTabletDriver-Except-I-Lied/blob/7f1c81683bca510f7c799a47b2c33cc61605125e/OpenTabletDriver.Desktop/Reflection/DesktopPluginManager.cs#L180-L190

I just cannot reproduce this on my side.

Kuuuube:
Just got to debugging this myself and I can replicate the exact same issue with `Exists`.

Tested with `Directory.Exists(source.FullName)` instead and that does work.

Mrcubix:
It seems that .NET on Linux specifically requires devs to call `Refresh` if the directory is modified outside of the instance.
(For example by using Directory.Create, Directory.Delete or Directory.Move)

Mrcubix:
Seems to be fixed afaik now with a call to Refresh

Kuuuube (approved):
Looks good. Works now.

PR #4047, opened 2025-09-07
Add CI workflow to verify new tablet names in PRs

### LINKED ISSUE #4045: Add check to config PRs for correct naming between TABLETS.md and tablet config name
### Description

Ideally the check extracts the added name(s) and tries to match the TABLETS.md name to the name defined in the added tablet configuration. The idea is to avoid fixup PR's like #4042.

### Acknowledgements

epsilorne:
Closes #4045.

Adds a basic CI workflow -- activated by any PR -- that checks if there were changes to the tablet configs (e.g. a modified or new config), and if there were, makes sure its `Name` exists (exactly) in `TABLETS.md`. This would prevent a tablet config with a mismatched `TABLETS.md` entry or vice versa.

### DIFF
--- a/.github/workflows/verify-tablet-names.yml
+++ b/.github/workflows/verify-tablet-names.yml
@@ -1,0 +1,33 @@
+name: Verify Correct Tablet Names
+
+on:
+  pull_request:
+
+jobs:
+  verify-names:
+    runs-on: ubuntu-latest
+    name: Verify Tablet Names
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get changed tablet configs
+        id: delta
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c
+        with:
+          files: OpenTabletDriver.Configurations/Configurations/**/*.json
+          separator: ","
+
+      - name: Check changed configs match TABLETS.md
+        if: steps.delta.outputs.any_changed == 'true'
+        env:
+          TABLETS: TABLETS.md
+          CONFIGS: ${{ steps.delta.outputs.all_changed_files }}
+        run: |
+          IFS=","
+          for CONF in $CONFIGS; do
+            NAME=$(cat $CONF| jq -r '.Name')
+            if ! grep -q "| $NAME " $TABLETS; then
+              echo "Failure: '$NAME' was not found in $TABLETS - was it spelt correctly?"
+              fi
+          done


.github/workflows/verify-tablet-names.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get changed tablet configs
+        id: delta
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c
+        with:
+          files: OpenTabletDriver.Configurations/Configurations/**/*.json
+          separator: ","
```
epsilorne, line 19: Required because some configs have spaces in their name, which would break the loop later on

gonX:
This looks good, but it would be nice if there was an option to run this manually as well.

You can see an example of how this is done here:
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/09fef9c41ca98a840162ba67bbba8962a7580a4c/.github/workflows/manual-pr-labeler.yml

.github/workflows/verify-tablet-names.yml
```diff
@@ -0,0 +1,33 @@
...
+        if: steps.delta.outputs.any_changed == 'true'
+        env:
+          TABLETS: TABLETS.md
+          CONFIGS: ${{ steps.delta.outputs.all_changed_files }}
+        run: |
+          IFS=","
+          for CONF in $CONFIGS; do
+            NAME=$(cat $CONF| jq -r '.Name')
+            if ! grep -q "| $NAME " $TABLETS; then
+              echo "Failure: '$NAME' was not found in $TABLETS - was it spelt correctly?"
```
gonX, line 31: We use US English:
```suggestion
              echo "Failure: '$NAME' was not found in $TABLETS - was it spelled correctly?"
```

epsilorne:
Have applied suggestions as per the code review, and also switched to a native git/bash implementation instead

.github/workflows/verify-tablet-names.yml
```diff
@@ -0,0 +1,32 @@
+name: Verify Correct Tablet Names
+
+on:
+  workflow_dispatch:
+    inputs:
+      pull_request_ids:
+        required: true
```
gonX, line 7: You need to handle the defined inputs somewhere otherwise you might as well remove them (regardless the input key doesn't make logical sense here)
```suggestion
```

see https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch

.github/workflows/verify-tablet-names.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Check changed configs match TABLETS.md
+        env:
+          TABLETS: TABLETS.md
+        run: |
+          IFS=$'\n'
+          mapfile -t configs < <(git diff --name-only --diff-filter=AM HEAD^ HEAD | grep '^OpenTabletDriver\.Configurations/Configurations/.*\.json')
```
gonX, line 23: You need to `mapfile` differently if the workflow is run manually, otherwise `git diff` will not fill

There are ways to check whether it's a manual run or not, again see docs: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch

gonX (changes requested):
Thank you very much for updating the PR as requested. I unfortunately have some further changes I'd like implemented.

epsilorne:
All good - I've switched to just the `push` event as discussed in Discord

.github/workflows/verify-tablet-names.yml
```diff
@@ -0,0 +1,28 @@
+name: Verify Correct Tablet Names
+
+on:
+  push
```
gonX, line 4: I'm really sorry but I missed the part where you asked if we wanted to have `pull_request` removed as well:

```suggestion
  push:
  pull_request:
```

This should be the final change that is necessary, then we can merge.

PR #4049, opened 2025-09-07
[0.6.X] Order Filters & Tools by Name

### LINKED ISSUE #1972: Filters are no longer ordered
## Description

Previously (0.6.0, 0.6.0.1) filters were listed alphabetically. On 0.6.0.2 they are now randomized like they were in <=0.5.3.3. 

I suspect this was caused by #1939.

## System Information:

| Name             | Value |
| ---------------- | ----- |
| Operating System         | Win10
| OpenTabletDriver Version | 0.6.0.2

Mrcubix:
Closes #1972

Also fixes a bug with the selection not being set back to the previously selected plugin.

### DIFF
--- a/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
+++ b/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
+using System.Reflection;
 using Eto.Drawing;
 using Eto.Forms;
 using OpenTabletDriver.Desktop;
@@ -79,12 +81,15 @@ namespace OpenTabletDriver.UX.Controls
         private void RefreshContent()
         {
             var types = AppInfo.PluginManager.GetChildTypes<TSource>();
+            var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.Name)]);
+
+            var oldSelected = sourceSelector.SelectedItem;
+            var newSelected = sortedTypes.FirstOrDefault(t => t.FullName == oldSelected?.FullName);
 
             // Update DataStore to new types, this refreshes the editor.
-            var prevIndex = sourceSelector.SelectedIndex;
-            sourceSelector.SelectedIndex = -1;
-            sourceSelector.DataStore = types;
-            sourceSelector.SelectedIndex = prevIndex;
+            sourceSelector.SelectedItem = null;
+            sourceSelector.DataStore = sortedTypes;
+            sourceSelector.SelectedItem = newSelected;
 
             this.Content = types.Any() ? mainContent : placeholder;
         }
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
```diff
@@ -79,12 +81,15 @@ protected virtual void OnStoreCollectionChanged()
         private void RefreshContent()
         {
             var types = AppInfo.PluginManager.GetChildTypes<TSource>();
+            var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.Name)]);
+
+            var oldSelected = sourceSelector.SelectedItem;
+            var newSelected = sortedTypes.FirstOrDefault(t => t.FullName == oldSelected?.FullName);
```
Kuuuube, line 87: I think `FullName` is not  the name of the filter (as defined with `[PluginName("whatever")]`) presented to the user but the filter class. Causes some odd sorting in some cases.

Kuuuube: This seems to make it sort by name correctly:
```cs
var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.GetFriendlyName())]);
```

OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
```diff
@@ -79,12 +81,15 @@ protected virtual void OnStoreCollectionChanged()
         private void RefreshContent()
         {
             var types = AppInfo.PluginManager.GetChildTypes<TSource>();
+            var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.Name)]);
```
Kuuuube, line 84: Suggestion if you want to apply this:
```suggestion
            var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.GetFriendlyName())]);
```

OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
gonX: After some further conversation on Discord I think it makes sense to do this change as well, as not all plugins have sane class/type naming:
image

Mrcubix: This will lead to some heavy overheads, Reflection isn't free.

gonX: If the function is only called when opening the plugin manager, that's fine. We can fix it in a later version if we start having complaints about this being slow - also it will be in a release candidate, so hopefully any issues along those lines gets caught during that.

Mrcubix: It's currently being called multiple times for the same objects & never cached, the impact of using GetFriendlyName is visible, even on my system.

Mrcubix:
The caching issue is tracked in #4050

gonX (approved):
LGTM, tested working by compiling this branch

PR #3022, opened 2023-11-21
Backport tablet support to 0.6.x

Kuuuube:

PR #4058, opened 2025-09-08
[06x] Backport: Candidate fix for Wacom Graphire AUX buttons.

### LINKED ISSUE #2992: Incorrect auxiliary button behavior for Wacom CTE-440

gonX:
Backport of #3673 with necessary namespace adjustments applied, and `NearProximity`/`HighConfidence` removed as hover distance was not included in the original PR. The latter concern has been pointed out in the original PR.

Original commit description:

> This fix has been tested with a CTE-430 (no aux buttons!) and a CTE-440 (2 buttons and scroll wheel). Models CTE-630, ET-0405* have also been updated but not tested.
> 
> Mapping of scrollwheel to AuxButtons report has been removed.

Fixes #2992 for 0.6.x - will go in 0.6.6 if we can find testers, otherwise it'll go in shortly after releasing 0.6.6 so people can test via the git branch (or this action build).

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-430.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-430.json
@@ -20,7 +20,7 @@
       "VendorID": 1386,
       "ProductID": 19,
       "InputReportLength": 8,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Bamboo.BambooReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Graphire.GraphireReportParser",
       "FeatureInitReport": [
         "AgI="
       ]
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-440.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/CTE-440.json
@@ -23,13 +23,13 @@
       "VendorID": 1386,
       "ProductID": 21,
       "InputReportLength": 9,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Bamboo.BambooReportParser"
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.Graphire.GraphireReportParser"
     },
     {
       "VendorID": 1386,
       "ProductID": 21,
       "InputReportLength": 8,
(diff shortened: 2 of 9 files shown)


gonX:
It seems hacky to set hover distance to 0, but if you think that works that's OK too.

Kuuuube:
It is. But it's better than omitting the data. Unless you want to make the change to separate hover distance and nearproximity (high confidence) like master has.

gonX:
Ok, did some further investigating, and the backport of #2907 was seen in #3022 and also has `HoverDistance` set to 0. Will change this PR to also do that.

Kuuuube (approved):

Kuuuube:
Confirming this works on my ET-0405-U.

PR #4063, opened 2025-09-09
[06x] Backport: Add CI workflow to verify new tablet names in PRs

gonX:
Backport of #4047


gonX:
I think I'll go a different route with this and create a script for this - it should also simplify CI runner usage so that it properly supports workflow_dispatch

PR #3056, opened 2023-11-27
Add Gaomon M5 support

gonX:
Freshens up stale PR at #2831

Fixes #2655

No verification, so setting as draft until tested

PR #2831, opened 2023-07-04
Add support for M5

telecomadm1145:
Device strings:[M5 Strings.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11944698/M5.Strings.txt)
Diagnostics:[M5 Diagnostics.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11944703/M5.Diagnostics.txt)

Works for my tablet:

PR #4087, opened 2025-09-20
[06x] Backport: Add support for Gaomon M5

gonX:
Backport of #3056


Kuuuube (approved):
Looks good.

Noting that the original pr creator (#2831) found specs for the digitizer maxes in the official driver's config files rather than testing in the OTD debugger.

PR #4097, opened 2025-09-21
[06x] Daemon/GetMatchingDevices: Filter devices without matching feature/output report lengths

### LINKED ISSUE #2129: Disregard endpoints that doesn't satisfy length requirements of output/feature report inits
## Description

Disregard endpoints that doesn't satisfy length requirements of output/feature report inits.

For example, if a config states that it needs to send an output/feature report init of an arbitrary length, then OTD should only attempt to match endpoints that supports at least that arbitrary length. It doesn't make sense to not do this.

gonX:
Fixes #2129 for 0.6.x

This should make interface selection more predictable

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -174,8 +174,8 @@ namespace OpenTabletDriver
                    where identifier.InputReportLength == null || identifier.InputReportLength == device.InputReportLength
                    where identifier.OutputReportLength == null || identifier.OutputReportLength == device.OutputReportLength
                    where identifier.FeatureReportLength == null || identifier.FeatureReportLength == device.FeatureReportLength
-                   where device.FeatureReportLength >= (identifier.FeatureInitReport?.Count ?? 0)
-                   where device.OutputReportLength >= (identifier.OutputInitReport?.Count ?? 0)
+                   where device.FeatureReportLength >= (identifier.FeatureInitReport?.Max(x => x.Length) ?? 0)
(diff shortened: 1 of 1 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -173,6 +173,8 @@ private IEnumerable<IDeviceEndpoint> GetMatchingDevices(TabletConfiguration conf
                    where device.CanOpen
                    where identifier.InputReportLength == null || identifier.InputReportLength == device.InputReportLength
                    where identifier.OutputReportLength == null || identifier.OutputReportLength == device.OutputReportLength
+                   where device.FeatureReportLength >= (identifier.FeatureInitReport?.Count ?? 0)
+                   where device.OutputReportLength >= (identifier.OutputInitReport?.Count ?? 0)
```
Kuuuube, line 177: The number of featureinits or outputinit shouldn't matter, it would be the length of the longest init. But I'm also not convinced the concerns in #2129 are even valid.
OTD inits are often clipped to be small by removing trailing zeroes which are ignored by the tablet anyways. And while these inits are usually short, oem drivers often send enormous inits to the device and then go read off an endpoint that is smaller.
And we have configs currently that would violate this. Since inits don't go to a specific endpoint it seems strange to filter like this.

gonX: Oh you are making a good point. It should be getting max length of any of the elements, not the amount of arrays.

As for the issue itself your comments are valuable there too - maybe the change doesn't make sense as you mention.

gonX:
Closing PR as linked issue has been decided as being invalid.

PR #4098, opened 2025-09-21
[06x] tests: Add check for Tablet Config names in tablets.md

gonX:
This adds a dedicated script to check for tablet names present in configs being absent from TABLETS.md

Should only check changed files on pull requests, but a `workflow_dispatch` handler was added to be run on-demand for all configuration files.

Can also be run locally with `./tests/test_tabletconfigs_present_in_tabletsmd.sh`

Logic stolen from OpenTabletDriver/OpenTabletDriver#4047

Alternative to #4063

Fix before merge:
- [x] Test for this PR should not test all configs. Logic needs to be double checked.

### DIFF
--- a/.github/workflows/shell-tests.yml
+++ b/.github/workflows/shell-tests.yml
@@ -15,3 +15,25 @@ jobs:
       - name: Test paths
         run: |
           ./tests/test_labeler_yaml.sh
+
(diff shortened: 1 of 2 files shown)


Kuuuube (approved):
Looks good, works.

PR #4101, opened 2025-09-21
[0.6.x] Allow matching FeatureReportLength

Kuuuube:
Noticed that this isn't allowed to be matched by configs for some reason. While no current configs use it, it seems reasonable to have in here and I imagine it wouldn't be any issue for performance for configs that have null for it anyways.

This value is output in diag and passed around everywhere already just for some reason not allowed in configs.


gonX (approved):
That's a weird oversight. LGTM

PR #4126, opened 2025-09-25
Add memoriam tab on show instead of setting visible

Kuuuube:
MacOS does not actually hide tabs when they are set to `Visible = false`, only the tab content is hidden.


gonX (approved):
LGTM

PR #4132, opened 2025-09-27
[06x] {Abs,Rel}OutputMode: Force GC collect

gonX:
Does an aggressive GC collect on pen out reports, otherwise an optimized collect

Leaving as draft since this needs testing on low end PC's, it might be worse


gonX:
Moving this to next release instead, since this is a "nice to" have and will definitely take a few hours to research better approaches for.

Worth noting that this PR did not help the person that I initially made this test for.

PR #4172, opened 2025-10-01
Windows artist mode

Kuuuube:


Mrcubix (commented):
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/2594564ae698615ba0d348a4c94e14e0158a9583/OpenTabletDriver.Native/Windows/PenPointer/PointerDevice.cs#L66-L69

This **will** throw an exception once in a while even if the input is valid.
Speaking from experience from experimenting with the API.

Kuuuube:
Putting this aside for now unless there's any great ideas on how to move forward. I cannot find a way to send subpixel values (available in the api as `himetric`). This is very important for getting proper precision for drawing.

I put some tests trying to get himetric working here https://github.com/Kuuuube/OpenTabletDriver/tree/windows-artist-mode-debug.
Potentially helpful reference for `GetPointerDeviceRects`: https://gist.github.com/fredemmott/c6949970ba25babc283fbe2434899b5d.

gonX:
Removing this from the 0.6.7 milestone given the points above. Thanks for the attempt, it would've been really cool if it worked!

Kuuuube:
Closing as this is now covered by the Windows Pen Pointer plugin here: https://github.com/Kuuuube/VoiDPlugins/tree/master/src/OutputMode/WindowsPenPointer. And available on the plugin repo.

Aforementioned issues are still present. If a way to resolve them is discovered, this can be reconsidered for a merge directly into the driver.

PR #4174, opened 2025-10-01
Run dotnet format

Kuuuube:
These two files not being formatted correctly makes running `dotnet format` on other prs annoying.


gonX (approved):
LGTM. Matches my `dotnet format` diff

PR #4190, opened 2025-10-04
[06x] github/pr-labeler: Add support for 'packaging' label

gonX:
Stops packaging changes from being labeled as `meta`. I don't think we had this label when the labeler structure was written.


gonX:
Removed `pack_local.sh` since a test was complaining about the file not existing.

PR #4197, opened 2025-10-05
[0.6.x] Replace bad subtraction with bitmask in XP-Pen offset parsers

Kuuuube:
I am hereby banning subtraction in parsers. (seriously though this is the second time this has happened and it wasn't my fault this time lol)

There are two variants of these tablets that send either "dirty" pressure bytes of "clean" pressure bytes:
`Dirty`: 0 pressure `00 20`, 8191 pressure `FF 3F`
`Clean`: 0 pressure `00 00`, 8191 pressure `FF 1F`
(note that these are unaligned, the byte order is reversed for parsing)

The lingering `00 20` causes the pressure in the dirty variant to need to be 8192 less. But doing this with subtraction causes the clean variant to underflow the uint.


gonX (approved):
Makes sense to me

PR #4199, opened 2025-10-05
Add new variant for Monoprice 10594

Kuuuube:
Verification: https://discord.com/channels/615607687467761684/615611007951306863/1424211292633305149
Diag: [geese diagnostics.json](https://github.com/user-attachments/files/22705123/geese.diagnostics.json)
String dump: [string dump.txt](https://github.com/user-attachments/files/22705124/string.dump.txt)

Not adding this onto the existing identifier with string check 6 since string 121 is empty on this variant. Not adding it on to string 6 and allowing string 121 to be empty on the existing identifier since that could potentially have unintended side effects.


gonX (approved):
LGTM

PR #4202, opened 2025-10-06
Fix libinput eraser not setting sometimes due to not flushing and sending two conflicting tool reports

Kuuuube:
Libinput gets angry sometimes when the pen tool is unset and the eraser tool is set in the same sync leading to some wonky behavior.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -118,6 +118,7 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 
             // unset opposite tool (in case tablet never sends us Reset/OutOfRange)
             Device.Write(EventType.EV_KEY, currentTool, 0);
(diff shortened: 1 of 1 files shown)


gonX (approved):
LGTM but may cause some issues if distance handling ends up being implemented due to the ordering in the output modes, since this PR flushes after any distance related stuff has been emitted:

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/b674de586857eebfe235978e392a26f696f6b353/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs#L138-L145

We may want to consider refactoring `EvdevVirtualTablet` entirely in the future, so that we don't actually write anything out before calling Flush(), instead using a queue per flush (which the flush then clears).

This should allow us to know to unset the tools (e.g. nullable private instance field `_lastTool` set in last invocation of flush) we know we've sent as being present and then flush unneeded tools as necessary, before continuing on with the pipeline.

IIRC, historically, the code used to flush on SetPosition, so the current design is mostly a relic of the past.

TL;DR: Accepting with a reminder to self to refactor this code to "assemble" the packet in `Flush()` instead.

PR #4210, opened 2025-10-07
[06x] CLI: Add plugin handling

### LINKED ISSUE #4209: Plugin Install / removal from file via OpenTabletDriver.Console
### Description

Would be a way to expand on the CLI experience.
Would also be useful in cases where users are running on an admin account.

Could also support drag & dropping files on Windows (just looking at the args and figuring out if it's a zip, dll or tar file)

### Acknowledgements

gonX:
Adds the following:

- `InstallPlugin` takes a fully qualified path
- `UninstallPlugin` takes a Plugin directory name
- `ListPlugins` lists the plugin directories

Fixes #4209 but does not implement the drag'n'drop functionality (yet)

### DIFF
--- a/OpenTabletDriver.Console/Program.Commands.cs
+++ b/OpenTabletDriver.Console/Program.Commands.cs
@@ -299,6 +299,18 @@ namespace OpenTabletDriver.Console
             await Driver.Instance.DetectTablets();
         }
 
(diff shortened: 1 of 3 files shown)


Kuuuube (approved):
Seems fine.

PR #4285, opened 2025-11-07
Add bottom padding on filters & tools list

Mrcubix:
From suggestion on discord : https://discord.com/channels/615607687467761684/723399565780189234/1436035336819052617

> Add an empty box below the last item in the plugins list, so that, if you scroll fully downwards, you have no trouble clicking on the last item

This is for unix-based platforms only as their users are the only ones known to be affected by the issue.

Before : 

image

After : 

image

### DIFF
--- a/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
+++ b/OpenTabletDriver.UX/Controls/PluginSettingStoreCollectionEditor.cs
@@ -81,10 +81,10 @@ namespace OpenTabletDriver.UX.Controls
         private void RefreshContent()
         {
             var types = AppInfo.PluginManager.GetChildTypes<TSource>();
-            var sortedTypes = new ReadOnlyCollection<TypeInfo>([.. types.OrderBy(t => t.GetFriendlyName())]);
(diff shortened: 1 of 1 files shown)


gonX (changes requested):
LGTM but I think there may be a mistakenly changed line

gonX (changes requested):
It looks like the latest commit (47c652db49) accidentally also removed the change needed for this change to properly work?

I suggest rebasing the commit history into a single commit once you fix that, though I won't require it.

gonX (dismissed):
LGTM.

This has an interesting side effect where the default plugin selected in the filter list is the null option. This means this helps mitigate #4161.

Kuuuube (changes requested):
This is kinda jank if you have a large filter list since it will scroll you to the bottom by default due to the `null` option being selected by default. Would like this to select the first filter on the list like it does currently.

Tested on Linux.

Kuuuube:
This somehow makes the scrollbar disappear on the filters list now. ????????????

eto truly is fascinating.

Notably if i click around a bunch after awhile it will randomly appear. Unsure of what causes it.

Kuuuube:
Also loads in with the first item not visible besides 1 pixel:
image

Kuuuube:
Moving this off the 0.6.7 milestone since this has become more messy than expected.

Issue #1664, opened 2021-11-25
absolute position for touch


UltimateBoi:
## Description

Is there anything that lets me use absolute position with touch on my touchpad? right now it only works on the pen.

Mrcubix:
State what tablet you're using, but i'm pretty sure this has already been added in pre-release.

UltimateBoi:
CTH-670

InfinityGhost:
> State what tablet you're using, but i'm pretty sure this has already been added in pre-release and master.

Depends on what they're actually looking for. This sounds like they want touch input to be interpreted as position data in which to move the cursor.

UltimateBoi:
> > State what tablet you're using, but i'm pretty sure this has already been added in pre-release and master.
> 
> Depends on what they're actually looking for. This sounds like they want touch input to be interpreted as position data in which to move the cursor.

Basically it clicks on the spot on the screen without me having to actually drag my finger across the pad it just clicks on the position when I touch. Its already getting this data but its not implemented yet. Just like absolute position for the pen but for touch basically.

Espionage724:
I would like to see this! I would like to use my finger, instead of a stylus to have the mouse move in absolute mode. If I tap the bottom-right of my tablet, the cursor should jump there.

gonX:
The feature is planned, its somewhat on the backlog but we are of course taking PR's

Mrcubix:
Update: Touch support is available via the `Enhanced Output Modes` plugin in Plugins > Plugin Manager, comes with customizable settings for each tablets in the `Filters` tab under `Touch Settings`. 

On top of adding support for touch, I've also taken a large amount of time & rewrites to implement non-native touch gestures,
results may vary from one person to another, but if you want them, try them & open a issue if you find anything weird or straight up bad. Check out `Touch Gesture` in the Plugin Manager.

Both of those have separate versions for 0.5.x & 0.6.x .

I'm still learning how to do UX & how to properly make that backend part so expect some issues, hopefully nothing critical.

Touch support in itself could be made available OOB in OTD, but that would require a lot of config & some UX changes (And Eto.Forms makes me crazy).

jamesbt365:
Click "show plugin wiki" in the plugin manager.

Mrcubix:
I primarily use it on Linux, on a raspberry Pi 4 connected & wirelessly.

PR #4327, opened 2025-11-30
Touch support

gonX:
Works on Linux and only 1 finger is tracked

Currently replicates Wacom behavior, which means that only relative mode touch is supported. Adding absolute mode (fixing #1664) is likely very easy, just tell uinput that the device is a direct device. However, requesting this behavior as the user seems difficult.

Before undraft:

- [ ] Add multitouch
- [ ] Split into meaningful commits
- [ ] Double check all changes

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-451.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-451.json
@@ -21,6 +21,12 @@
       "Buttons": {
         "ButtonCount": 1
       }
+    },
+    "Touch": {
+      "Width": 157.48,
+      "Height": 98.425,
+      "MaxX": 4095,
+      "MaxY": 4095
     }
   },
   "DigitizerIdentifiers": [
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-651.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-651.json
@@ -21,6 +21,12 @@
       "Buttons": {
         "ButtonCount": 1
       }
+    },
+    "Touch": {
+      "Width": 223.52,
+      "Height": 139.7,
+      "MaxX": 4095,
+      "MaxY": 4095
     }
   },
   "DigitizerIdentifiers": [
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
@@ -14,6 +14,12 @@
     "AuxiliaryButtons": {
       "ButtonCount": 8
     },
+    "Touch": {
+      "Width": 325.12,
+      "Height": 203.2,
+      "MaxX": 4095,
+      "MaxY": 4095
+    },
     "Wheel": {
       "StepCount": 71,
       "IsRelative": false,
--- a/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
+++ b/OpenTabletDriver.Configurations/Configurations/Wacom/PTH-851.json
@@ -14,12 +14,6 @@
     "AuxiliaryButtons": {
       "ButtonCount": 8
     },
-    "Touch": {
-      "Width": 325.12,
-      "Height": 203.2,
-      "MaxX": 4095,
-      "MaxY": 4095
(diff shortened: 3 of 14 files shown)


joaofrgomes:
Could macOS somehow be added to the list? macOS 28 is potentially going to nerf compatibility of Rosetta 2 with x86 device drivers (its extended support is officially meant for gaming only, so whether it can be coaxed to also work with those is anyone's guess), and leave an entire class of Wacom users in the dust.

We have maybe two more years of reasonable, fully-featured device compatibility (when macOS 29 is released and macOS 27 ceases getting security updates) via [the official drivers from Wacom as patched by nickthedude](https://github.com/thenickdude/wacom-driver-fix/releases), but already at the cost of not getting feature upgrades on our OSes; after that, either OTD fully supports these models and takes those crusty old drivers' place or into the trash cans or Craigslists and eBays of this world our Wacom tablets go, which is precisely what this project is meant to avoid.

joaofrgomes:
Update: ConsumerTouchDriver support on macOS 27.0 Golden Gate GM release (build 26A428) is absolutely **abysmal**. It can get the job done in an emergency, if my Magic Mouse runs out of battery or something, but for daily usage it's just atrocious. I no longer get smooth mouse tracking anymore, it's extremely difficult to get single-finger taps to register, etc., as if the driver wasn't properly polling the tablet or something.

While I will obviously file this under Feedback Assistant as something Apple should look into, as it's an unacceptable regression of something that should still be supported under Rosetta 2, it's clear that nickthedude's patched installer and Wacom's official Intel drivers can't even be relied upon in the short term, thus leaving OTD as the only sensible solution going forward.

Any updates on this front?

gonX:
@joaofrgomes This pull request is for providing an actual touch implementation for Linux specifically, so that native touch events (like 2 finger scrolling) work.

I don't have access to a macOS device, so creating a proper touch implementation for that will be difficult until I/someone else finalizes this PR so that we have a known good reference.

With that being said, OpenTabletDriver already largely processes touch for many tablets correctly, so it would be theoretically possible for a plugin to abstract touch reports back to cursor events, if the Enhanced Output Modes plugin does not already support this.

joaofrgomes:
The thing is, EnhancedOutputModes is not supported on macOS. Wacom Pen&Touch users that for some reason prefer Macs (which, for tablet users, is a disproportionate number) are effectively being ignored and left in the dust. Also, even pen support is extremely jittery on Golden Gate, so we're basically being asked between using older, unsafe OSes, chucking our tablets in the trash or… switching to Asahi Linux? I mean…

Mrcubix:
This following comment is out of context of this PR.

Enhanced Output Mode Does not depend on a specific OS, only the additional output modes, such as artist mode (Linux Only), Windows Ink & Vmulti (Windows Only).

Absolute Mode in Enhanced Output Mode just makes use of the existing MacOS pointer.
And that was confirmed to work for the few MacOS users that didn't despawn after i asked them for feedback.

Additionally, some users were able to get the beta Touch-Gestures working ( after applying some MacOS codesign bull for the UX ).

Note that this plugin does **NOT** send native touch events to the OS, i wouldn't even touch the MacOS API with any hair anyway 
( Even if i came to somehow own a Mac )

joaofrgomes:
Fair enough. I don't know what sort of weird voodoo happened in my system, but after swapping USB cables and ports, uninstalling the official driver, installing OTD, uninstalling OTD and reinstalling the official driver, it automagically came back to life, so I'm guessing that I can hold off on changing my setup for a bit.

At the moment, I have a project to finish and a thesis to write, but I'd love to help y'all getting this running and fully supporting touch tablets on the Mac. What do you reckon is the best venue to engage, the Discord server?

Mrcubix:
If you are an actual dev (and not just spamming some LLM without the knowledge), you have a touch tablet, and have enough sanity to work with whatever API Apple provides (Which probably does not even exist right now), you can prob join the discord if you have some question you want answered quicker.

Most of the work necessary, will be in actually implementing the proper API calls for creating the touch contacts events and sending them to the OS.

You also need to take into account, that Apple maybe hasn't provided any API or elements that would allow devs to inject touch inputs.

From what i understand tho, Apple working on touch related features is a recent development, so the tools to inject touch events may not be available yet.

PR #4361, opened 2025-12-10
Update to .NET 10.0

### LINKED ISSUE #4773: Update to dotnet 10.0
### Description

Support for dotnet 8.0 is getting dropped among modern Linux distributions, like for example Ubuntu 26.04 which only have dotnet 10.0 (and possibly later).

That makes it harder, if not impossible, to use OTD in the future.

I'd rather use OTD than the default Ugee driver for my device, so please consider adding support or moving to newer dotnet releases.

### Acknowledgements

gonX:
Like #3368 but for .NET 10.0

Also adds a `global.json` to properly define our `rollForward` approach. This should make building on CI more predictable in case actions change (or we move provider). A bit overly defensive, but it seems recommended that a `global.json` file is used when CI is involved with the codebase according to its page: https://learn.microsoft.com/en-us/dotnet/core/tools/global-json

Targets a later release while any potential upstream bugs are ironed out, and users can use this as a patch if they need (or want to test) 10.0 before we start supporting it.

My testing has been "It builds and the daemon indeed moves my cursor", so YMMV

### DIFF
--- a/.github/actions/build-bash/action.yml
+++ b/.github/actions/build-bash/action.yml
@@ -13,12 +13,12 @@ runs:
     - name: Install required packages for Debian
       if: inputs.package-name == 'Debian'
       shell: bash
-      run: sudo apt update && sudo apt install -y debhelper dotnet-sdk-8.0 build-essential
+      run: sudo apt update && sudo apt install -y debhelper dotnet-sdk-10.0 build-essential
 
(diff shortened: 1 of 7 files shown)


gonX:
Tests failing as this needs to also update GitHub actions definitions with .NET 10.0

Kuuuube (approved):

PR #4378, opened 2025-12-13
Add support for UGEE UE12

caterpillar-1:
- Verification: locally on my machine. More information could be provided if devs would like to check further. 
- Diagnosis: [diag-ugee-ue12.json](https://github.com/user-attachments/files/24139086/diag-ugee-ue12.json)
- Stylus only
- It seems that the buttons on the tablet are implemented as HID keyboard. Here's the output of `hid-recorder /dev/hidraw2`: [tablet-keys.txt](https://github.com/user-attachments/files/24139097/tablet-keys.txt). The proprietary driver may configure keys for different use with another usb endpoint. Here's the output of `hid-decode /dev/hidraw5`: [vendor.txt](https://github.com/user-attachments/files/24139107/vendor.txt)

working-with-xournalpp

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/UGEE/UE12.json
+++ b/OpenTabletDriver.Configurations/Configurations/UGEE/UE12.json
@@ -1,0 +1,30 @@
+{
+  "Name": "UGEE UE12",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 256.32,
+      "Height": 144.18,
+      "MaxX": 32767,
+      "MaxY": 32767
+    },
+    "Pen": {
+      "MaxPressure": 16383,
+      "ButtonCount": 2
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 8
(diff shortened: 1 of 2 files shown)


gonX:
Tablet buttons not working probably means you've attached to the wrong interface, or are missing `AuxiliaryDeviceIdentifiers`. The `MaxX`/`MaxY` also makes me think you're attaching to the non-vendor interface.

caterpillar-1:
There're 3 USB endpoints / HID interfaces on the digitalizer. Two can be recognized by kernel driver while one is "vendor specific". Now the config chooses the latter one, with which MaxX/MaxY turn out different. Is this the same situation as other devices?

@gonX check this: 
[diag-ugee-ue12-fix.json](https://github.com/user-attachments/files/26133146/diag-ugee-ue12-fix.json)

Now both pen and buttons on the tablet can be detected on my side. @amarru9675 can you try out this patch on your device?

PR #4563, opened 2026-02-21
06x nullability goal

gonX:
Depends on changes in #4562

Can be undrafted when linked PR is merged


gonX:
Superseded by #4779 as it doesn't depend on another PR (and has been rewritten to be more explicit in only adjusting behavior related to nullability)

PR #4607, opened 2026-02-25
Use concrete type where possible for improved performance

gonX:
Kinda free change - don't use interface types when you can use the class directly, as this causes an interface lookup on use.

Functions that have had their types changed has also had relevant function usages updated (such as `List.Count` -> `Array.Length`)

Also changed the ReSharper severity in editorconfig to a warning

Aside from binding lookup in the binding handler, none of the changed functions are in any hot paths, so only initialization should have a measurable performance change.

I tested most of the code paths, including the X11 changes.

At least 1 of these changes are seen in #4562

---

A note on `IReadOnlyDictionary` that was used in `GeneratedControls.cs`: The closest concrete type I could find would've been `FrozenDictionary`, but looking at benchmarks made by others, performance is exclusively improved on key- or value-only operations, as internally a `FrozenDictionary` allocates key- and value-only arrays to accelerate these types of operation, but the allocation cost means you would also have to read a significant amount of times from the dictionary (and also to be key- or value-only operations) before you'd even start to see the performance tradeoff over time.

Additionally, it seemed that the intention was to signal to developers that the dictionary shouldn't be changed, but I don't think we explicitly need to discourage that.

So I'd rather not waste power initializing a FrozenDictionary on every UX launch, so this just gets converted to a normal `Dictionary`.

The other changes should be self-explanatory. `GetXRandrDisplays` has been converted to an array that gets initialized by the count size, rather than a dynamically allocated list.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -36,6 +36,8 @@ dotnet_diagnostic.CA2216.severity = warning
 dotnet_diagnostic.CA1822.severity = warning
 # use compile-time regex
 dotnet_diagnostic.SYSLIB1045.severity = warning
+# use concrete types when possible for improved performance
+dotnet_diagnostic.CA1859.severity = warning
 
 # IDE-only inspections (Rider, etc.)
 resharper_can_simplify_dictionary_lookup_with_try_get_value_highlighting=warning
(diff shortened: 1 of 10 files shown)


gonX:
This needs a rebase now that the multi-wheel PR has been merged. Will convert to draft

gonX:
Rebased

Kuuuube (approved):

PR #4625, opened 2026-03-01
Refactor Linux native methods

gonX:
This PR aims to refactor the OpenTabletDriver.Native Linux API to harden against developer mistakes. External function signatures have been double checked, and used types in function signatures have been improved (e.g. using handles instead of raw `IntPtr`s) or signatures have been improved (e.g. `ref` -> `in` where the destination function only needed `in` behavior) to help the compiler optimize more.

A breakdown of all the changes are described below:

- Adds editorconfig TODO for not making P/Invoke functions public
  - Reasoning: Potential plugin API security issue and it also makes the plugin API unclear (e.g. timers should not be used by plugins)
  - To reduce the scope of this PR this isn't entirely enabled yet, but is implemented for some of the native Linux operations.
- Adds editorconfig TODO for using `[LibraryImport]` over `[DllImport`]
  - As it builds marshalling code compile-time instead of at runtime, reducing (likely static) runtime overhead.

- Rewrites Linux native methods to use `[LibraryImport]`

- Rewrites native udev/uinput, XLib, XRandr, and Linux timer related stuff to use built-in handle wrapper classes
  - Related classes are rewritten to use these handles

- Added wrapper functions for Linux native unsafe contexts, so that the consumer (e.g. `EvdevVirtualTablet`) does not have to be unsafe itself.

- Drops usage of enabling event types to evdev endpoint, as upstream documentation states that the appropriate event types are automatically enabled when enabling event codes
  - The function enabling event types remains as to document that the function is generally not necessary to use

- Enables nullable reference types in touched classes

- Extensively documented the `Evdev` class

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -27,15 +27,24 @@ csharp_space_after_semicolon_in_for_statement = true
 csharp_space_before_semicolon_in_for_statement = false
 csharp_space_around_declaration_statements = false
 dotnet_sort_system_directives_first = true
+
 # Disposable types
 dotnet_diagnostic.CA1063.severity = warning
 dotnet_diagnostic.CA1816.severity = warning
 dotnet_diagnostic.CA2213.severity = warning
 dotnet_diagnostic.CA2216.severity = warning
+
 # Method can be marked as static
 dotnet_diagnostic.CA1822.severity = warning
+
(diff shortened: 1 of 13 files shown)


Kuuuube:
Needs rebase.

PR #4628, opened 2026-03-02
macOS: Send tablet proximity event immediately on eraser/pen tool switch

juan-martinez-pf:
On macOS, switching between the pen and eraser tool did not immediately notify applications of the tool change. The proximity event (which tells apps like Krita and Photoshop which tool is active) was only sent on the next 200ms idle refresh in `ApplyTabletValues()`, causing a delayed or missed tool switch.

This sends a `kCGEventTabletProximity` event immediately when `SetEraser()` is called with a new value, matching how the Linux `EvdevVirtualTablet` handles tool switches. The 200ms idle proximity refresh in `ApplyTabletValues()` is refactored to use the same shared helper and constants.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/MacOSVirtualMouse.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/MacOSVirtualMouse.cs
@@ -15,6 +15,23 @@ namespace OpenTabletDriver.Desktop.Interop.Input
         private const int DoubleClickMoveTolerance = 8;
         private const int ProximityExpiresDurationInMs = 200;
 
+        // The capabilityMask is essential for Adobe software to recognize tablet pressure sensitivity,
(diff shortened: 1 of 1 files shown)


gonX:
cc @AkiSakurai

AkiSakurai:
I don't have any device with an eraser to test it with, but that makes sense for me.

juan-martinez-pf:
Yes I followed the pattern set in the Linux port. It's a very frustrating experience using the eraser and it randomly does not switch with hover or tap. With this change it switches instantly on hover.

I tested with a Wacom PTH450 & PTH451 and it's a refreshing fix. I can also test with a Wacom Cintiq 24 DTH2420 if y'all like.

InfinityGhost (approved):
Lg2m

gonX (approved):

PR #4629, opened 2026-03-02
Support InitReadTimeout attribute for device detection

juan-martinez-pf:
Some tablets connect through wireless dongles, those dongles stay physically plugged in via USB even when the tablet itself is powered off. Because the dongle's USB descriptors match the tablet configuration, OTD detects the tablet as present, but no data ever arrives, leaving the user with a false positive detection.

This adds support for an `InitReadTimeout` attribute on `DeviceIdentifier.Attributes`. When set, `MatchDevice()` waits up to the specified milliseconds for the first report after opening the endpoint. If no data arrives, the device is rejected. If the timed-out identifier is an auxiliary device, the entire tablet detection is rejected to avoid a partial match.

The signaling is implemented via a `ManualResetEventSlim` in `DeviceReader`, set on the first parsed report.

An upcoming update to the [WirelessKitAddon](https://github.com/juan-martinez-pf/WirelessKitAddon) plugin will build on this change to add automatic re-detection when a wireless tablet powers on after being initially rejected.

### DIFF
--- a/OpenTabletDriver/Devices/DeviceReader.cs
+++ b/OpenTabletDriver/Devices/DeviceReader.cs
@@ -22,6 +22,7 @@ namespace OpenTabletDriver.Devices
         }
 
         private readonly Thread workerThread;
+        private readonly ManualResetEventSlim _firstReport = new ManualResetEventSlim(false);
         private bool initialized, connected;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


gonX (commented):
Is there an example tablet configuration to accommodate this? Or maybe I'm not understanding the purpose of the linked plugin.

juan-martinez-pf:
The Wacom Wireless Accessory Kit (ACK40401) supports these older tablet models:
- Bamboo 3rd Gen (CTH-470, 670)
- Intuos 5, Intuos Pro (1st Gen PTH-451, 651, 851)
- Intuos Pen & Touch (CTH-480, 680, 490, 690)

I was going to replace the configs in the WirelessAddon Plugin mentioned above. Here is an example config with wireless dongle PiD 132: [PTH-450.json](https://github.com/user-attachments/files/25731739/PTH-450.json) 

`DesktopDeviceConfigurationProvider.cs` says "Built-in configurations may be overridden if the Name matches exactly." So I assumed shipping the wireless related configs in the plugin was the correct way to do it.

I only need core to have the `InitReadTimeout` feature, because I can't do that in the plugin.

juan-martinez-pf:
After sleeping on it, maybe I should try a different approach at a higher level. Maybe this change would affect non-Wacom tablets that couldn't defeat the timer.

The root of my issue is not empty data streams, it's that device detection is not triggered by dongle connection events, only by USB connection events.

The plugin already identifies the dongle. Maybe I should start from there.

I'll close the PR sans merge.

PR #4630, opened 2026-03-02
macOS: Fix race condition in DrainPendingPosition

juan-martinez-pf:
I found this bug while doing my other macOS work, and it seemed trivial to fix, so here it is.

`QueuePendingPosition()` and `DrainPendingPosition()` run on different threads. Previously, `_pendingX` was set before `_pendingY`. A concurrent call to `DrainPendingPosition()` could see `_pendingX.HasValue == true` while `_pendingY` was still null, causing a `NullReferenceException` on `_pendingY!.Value`.

This swaps the assignment order so `_pendingY` is set before `_pendingX`, ensuring that when `DrainPendingPosition()` reads `_pendingX` as non-null, `_pendingY` is guaranteed to already be set.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/MacOSVirtualMouse.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/MacOSVirtualMouse.cs
@@ -143,9 +143,9 @@ namespace OpenTabletDriver.Desktop.Interop.Input
 
         private Vector2? DrainPendingPosition()
         {
-            if (_pendingX.HasValue)
+            if (_pendingX.HasValue && _pendingY.HasValue)
             {
-                var vector2 = new Vector2(_pendingX!.Value, _pendingY!.Value);
+                var vector2 = new Vector2(_pendingX.Value, _pendingY.Value);
                 _pendingX = null;
                 _pendingY = null;
                 return vector2;


Kuuuube:
While this probably does fix the issue, would it not be more sane to edit the if statement in `DrainPendingPosition` to check both? I don't think checking that value is particularly expensive.

This kind of fix seems like it could easily accidentally be undone later.

juan-martinez-pf:
> While this probably does fix the issue, would it not be more sane to edit the if statement in `DrainPendingPosition` to check both? I don't think checking that value is particularly expensive.
> 
> This kind of fix seems like it could easily accidentally be undone later.

I think you are correct. Let me attempt to fix it that way.

gonX:
Feel free to get rid of some off the null-silencing operators (`!`) while you're at it, they would've caught this issue initially.

With that being said these operations probably be done on the same thread where possible, or something else altogether, to avoid the race condition from happening in the first place. It can still happen that HasValue turns false before the next value is read atomically. Would using `Interlocked` functions help here?

@AkiSakurai thoughts?

juan-martinez-pf:
I've reset the assignment order, and am now checking both values in DrainPendingPosition(), I think this is a more resilient fix. I also made sure the commit history was clean.

@gonX I'd rather not increase the scope of this PR. It was just a "quick fix". My true goal is to get full macOS support for the WirelessAddonKit.

gonX (approved):
Code changes LGTM, can merge if checks pass

AkiSakurai:
> Feel free to get rid of some off the null-silencing operators (`!`) while you're at it, they would've caught this issue initially.
> 
> With that being said these operations probably be done on the same thread where possible, or something else altogether, to avoid the race condition from happening in the first place. It can still happen that HasValue turns false before the next value is read atomically. Would using `Interlocked` functions help here?
> 
> @AkiSakurai thoughts?

We should probably serialize the pipeline, the logic is too complicated to handle with Interlocked. For now, the current fix  should still be fine if it reduces the chance of a race condition.

PR #4653, opened 2026-03-08
Fix macOS crash when switching profiles with wheel tablets

juan-martinez-pf:
Fixes an ArgumentOutOfRangeException crash on macOS when connect to a tablet with a wheel (e.g., PTH-451 / Intuos5)
- This only affected the UX, not the daemon.
- On macOS, `Pages.Clear()` in `OnProfileChanged` _clears_ tab pages, so IndexOf returns -1 on the next call to `OnTabletChanged`, causing Insert(-1, ...) to throw
- This guards against the -1 index by falling back to Pages.Add()
- Final page order is unaffected since macOS rebuilds it via SetPageVisibility


Kuuuube (approved):
Should be good to go. Does not create any problems on windows or linux as far as I can tell.

PR #4656, opened 2026-03-08
Fix locked cursor handling in relative mode on macOS

kerudion:
This fixes cursor bumping into the edges of the screen when playing games because it's not recentered by the game explicitly, and also fixes the delta being very quantized

I don't know if this also happens on other platforms, but with these changes it works perfectly on macOS

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Relative/MacOSRelativePointer.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Relative/MacOSRelativePointer.cs
@@ -11,6 +11,8 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Relative
 
     public class MacOSRelativePointer : MacOSVirtualMouse, IRelativePointer
     {
+        private Vector2 error;
+
         public void SetPosition(Vector2 delta)
         {
             QueuePendingPosition(delta.X, delta.Y);
@@ -18,9 +20,17 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Relative
 
         protected override void SetPendingPosition(IntPtr mouseEvent, float x, float y)
         {
-            CGEventSetLocation(mouseEvent, GetCursorPosition() + new CGPoint(x, y));
-            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, x);
-            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, y);
+            var pos = GetCursorPosition();
+            if (CGCursorIsVisible())
+                CGEventSetLocation(mouseEvent, pos + new CGPoint(x, y));
+            else
+                CGEventSetLocation(mouseEvent, pos);
+
+            Vector2 delta = new Vector2(x, y) + error;
+            error = new Vector2(delta.X % 1, delta.Y % 1);
+
+            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, Math.Truncate(delta.X));
+            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, Math.Truncate(delta.Y));
         }
 
         protected override void ResetPendingPosition(IntPtr mouseEvent)
--- a/OpenTabletDriver.Native/OSX/OSX.cs
+++ b/OpenTabletDriver.Native/OSX/OSX.cs
@@ -55,6 +55,9 @@ namespace OpenTabletDriver.Native.OSX
         public extern static void CGEventSetLocation(CGEventRef eventRef, CGPoint location);
 
         [DllImport(Quartz)]
+        public extern static bool CGCursorIsVisible();
+
+        [DllImport(Quartz)]
         public extern static CGEventRef CGEventCreateScrollWheelEvent2(CGEventRef eventRef, CGScrollEventUnit units, uint wheelCount, int wheel1, int wheel2, int wheel3);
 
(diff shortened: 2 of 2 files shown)


OpenTabletDriver.Desktop/Interop/Input/Relative/MacOSRelativePointer.cs
```diff
@@ -11,16 +11,26 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Relative
...
+            if (CGCursorIsVisible())
+                CGEventSetLocation(mouseEvent, pos + new CGPoint(x, y));
+            else
+                CGEventSetLocation(mouseEvent, pos);
+
+            Vector2 delta = new Vector2(x, y) + error;
+            error = new Vector2(delta.X % 1, delta.Y % 1);
+
+            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, Math.Truncate(delta.X));
+            CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, Math.Truncate(delta.Y));
```
gonX, line 33: since these functions are taking a double I don't believe you actually need to accumulate the error

OpenTabletDriver.Native/OSX/OSX.cs
```diff
@@ -54,6 +54,9 @@ public extern static CGEventRef CGEventCreateMouseEvent(CGEventSourceRef source,
         [DllImport(Quartz)]
         public extern static void CGEventSetLocation(CGEventRef eventRef, CGPoint location);
 
+        [DllImport(Quartz)]
+        public extern static bool CGCursorIsVisible();
```
gonX, line 62: Interestingly, this function is marked as deprecated with no replacement: https://developer.apple.com/documentation/coregraphics/cgcursorisvisible()

This should be marked as part of the API XML docs for this function as a remark:

```suggestion
        /// <remarks>
        /// Upstream has marked this function as deprecated with no replacement:
        /// <a href="https://developer.apple.com/documentation/coregraphics/cgcursorisvisible()">developer.apple.com</a>
        /// </remarks>
        [DllImport(Quartz)]
        public extern static bool CGCursorIsVisible();
```

gonX (dismissed):
Code looks not completely wrong, but I do have some concerns.

kerudion:
Yeah, Apple loves deprecating APIs with no replacements. Some even work after being removed from the docs, so I think this should be fine.

And even though they take double, looks like they straight up ignore the fractional part. I have tried `Math.Truncate` and it looks the same, so I think it really just ignores it. Also I left truncate in my fix just to be logically safe, and that if it suddenly starts working in some update it does not overcorrect.

In the tests I was moving slow and fast and in different directions. The treshold to move with raw delta appears to be very high, and it also awfully quantizes.

Tests:

Original (this one has cursor highlight to show the bumping, in others with the fix it correctly stays in center)
```csharp
CGEventSetLocation(mouseEvent, GetCursorPosition() + new CGPoint(x, y));
CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, x);
CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, y);
```
https://github.com/user-attachments/assets/3340486f-454b-4da8-b586-fb1cfd346396

Raw delta
```csharp
var pos = GetCursorPosition();
if (CGCursorIsVisible())
  CGEventSetLocation(mouseEvent, pos + new CGPoint(x, y));
else
  CGEventSetLocation(mouseEvent, pos);

CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, x);
CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, y);
```
https://github.com/user-attachments/assets/77336400-e6ae-4d13-8835-2beb36f04648

Truncate delta (looks the same as just raw)
```csharp
var pos = GetCursorPosition();
if (CGCursorIsVisible())
  CGEventSetLocation(mouseEvent, pos + new CGPoint(x, y));
else
  CGEventSetLocation(mouseEvent, pos);

CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, Math.Truncate(x));
CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, Math.Truncate(y));
```
https://github.com/user-attachments/assets/cc7a3451-8795-4046-8599-6ec7c4edcca0

Full fix
```csharp
var pos = GetCursorPosition();
if (CGCursorIsVisible())
  CGEventSetLocation(mouseEvent, pos + new CGPoint(x, y));
else
  CGEventSetLocation(mouseEvent, pos);

Vector2 delta = new Vector2(x, y) + error;
error = new Vector2(delta.X % 1, delta.Y % 1);

CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaX, Math.Truncate(delta.X));
CGEventSetDoubleValueField(mouseEvent, CGEventField.mouseEventDeltaY, Math.Truncate(delta.Y));
```
https://github.com/user-attachments/assets/31ea8f12-d956-4063-ab07-8a6cbea5222a

AkiSakurai:
Would it be better to add an option to toggle this behavior?
Right now it assumes the foreground app is in relative mode whenever the cursor is hidden.

AkiSakurai:
You might need to reset the error when the pen first enters the tablet.

gonX:
Thanks for the thorough evidence that accumulating the fractional part locally is necessary, so that part is fine with me. I'd like for you to address AkiSakurai's concerns, however.

PR #4661, opened 2026-03-10
Add support for rotation parsing and output

Kuuuube:
Implemented for PTK-x70 and PTH-x60. Setting both min and max rotation values in the config for this.

Tested myself with my PTZ-431W, PTK-440, PTK-640, PTK-1240, PTH-451, PTH-460, PTH-660, PTK-470. Other tablets in the same series' as these should match the same specs and this is supported by input-wacom.

### DIFF
--- a/OpenTabletDriver.Plugin/Output/IOutputMode.cs
+++ b/OpenTabletDriver.Plugin/Output/IOutputMode.cs
@@ -38,5 +38,10 @@ namespace OpenTabletDriver.Plugin.Output
         /// Whether to disable tilt
         /// </summary>
         public bool DisableTilt { set; get; }
+
+        /// <summary>
+        /// Whether to disable rotation
+        /// </summary>
+        public bool DisableRotation { set; get; }
     }
 }
--- a/OpenTabletDriver.Plugin/Platform/Pointer/IRotationHandler.cs
+++ b/OpenTabletDriver.Plugin/Platform/Pointer/IRotationHandler.cs
@@ -1,0 +1,7 @@
+namespace OpenTabletDriver.Plugin.Platform.Pointer
+{
+    public interface IRotationHandler
+    {
+        void SetRotation(float percentage);
+    }
+}
--- a/OpenTabletDriver.Plugin/Platform/Pointer/IRotationHandler.cs
+++ b/OpenTabletDriver.Plugin/Platform/Pointer/IRotationHandler.cs
@@ -2,6 +2,10 @@ namespace OpenTabletDriver.Plugin.Platform.Pointer
 {
     public interface IRotationHandler
     {
+        /// <summary>
+        /// Set the rotation of the pen using absolute barrel rotation data.
+        /// </summary>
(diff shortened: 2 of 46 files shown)


OpenTabletDriver.Plugin/Output/IOutputMode.cs
```diff
@@ -38,5 +38,10 @@ public interface IOutputMode : IPipelineElement<IDeviceReport>, IDisposable
         /// Whether to disable tilt
         /// </summary>
         public bool DisableTilt { set; get; }
+
+        /// <summary>
+        /// Whether to disable rotation
+        /// </summary>
+        public bool DisableRotation { set; get; }
```
gonX, line 45: Did we have any issues with plugin output modes (like WinInk) breaking when we added new properties last time? If we did, this PR probably needs a `breaking` tag.

OpenTabletDriver.Plugin/Tablet/IRotationReport.cs
```diff
@@ -0,0 +1,7 @@
+namespace OpenTabletDriver.Plugin.Tablet
+{
+    public interface IRotationReport : IDeviceReport
+    {
+        int Rotation { set; get; }
```
gonX, line 11: Would be preferable to have xmldocs here for the expected range and/or how the range is intended to be looked up

OpenTabletDriver.Plugin/Platform/Pointer/IRotationHandler.cs
```diff
@@ -0,0 +1,7 @@
+namespace OpenTabletDriver.Plugin.Platform.Pointer
+{
+    public interface IRotationHandler
+    {
+        void SetRotation(float percentage);
```
gonX, line 9: add xmldocs

OpenTabletDriver.Plugin/ReportFormatter.cs
```diff
@@ -130,6 +134,10 @@ private static IEnumerable<string> GetStringFormat(ITiltReport tiltReport)
             yield return $"Tilt:[{tiltReport.Tilt.X},{tiltReport.Tilt.Y}]";
         }
 
+        private static IEnumerable<string> GetStringFormat(IRotationReport rotationReport)
+        {
+            yield return $"Rotation:{rotationReport.Rotation}";
+        }
```
gonX, line 140: formatting nitpick:
```suggestion
        }
        
```

gonX (dismissed):
I don't see anything wrong with this other than some nitpicky documentation stuff.

OpenTabletDriver.Plugin/Output/IOutputMode.cs
Kuuuube: The pressure and tilt disable options didnt require any changes to win ink plugin so I don't think so.

Kuuuube:
Should be good now.

gonX (approved):
LGTM

Issue #2318, opened 2022-06-19
Expand `.editorconfig` with new conventions


vedattt:
Our .editorconfig rules should get another look as the rewrite nears completion.

For example, `dotnet format` has [naming rules for private members](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules#example-private-instance-fields-with-underscore) (underscores, etc.)

gonX:
I worked a bit on this but it's quite an undertaking to get correct. I might get some parts in for 0.6.7 but probably not much.

WIP version here (and context is behind this message): https://canary.discord.com/channels/615607687467761684/628651971267788800/1452860384942887034

gonX:
I'm going to move this to v0.6.8 and try to prioritize it as one of the first things after the release of v0.6.7 - I have a branch in the works trying to finalize this up, but it's a lot of changes, so to avoid surprise-breaking something a few days before release this has to get moved.

Issue #1381, opened 2021-08-30
Drag scroll by pressing supporting pen button


xillmera:
add chosable action to mouse properties at binding menu that allow : 
by pressing button and after changing cursor position (not only vertical but also in gorisontal plane) executed shifts on screen in required direction with some ratio|sensativity. Simular functionality can be found in another osce project "Inkscape", where by pressing "space" novigation on layer can be executed.

ziwzih:
yes need something like this

osuka:
one of my favourite functionalities, would love this back - it is in the x driver but no longer supported on wayland/libinput so it would be amazing to have it in opentabletdriver

axrolls-skoged:
yeah please add this. this would be extremely convenient.

xenio:
Is the scroll with button available? I am on MacOs and this is the only function I'm missing to switch from the wacom driver.

jamesbt365:
As this issue isnt closed, no.

SevenChords:
I second this as well, wanted to switch to tablet but still am required to use my mouse for quite a few applications as they don't support middle click + drag scrolling, just scrollwheel.

HollisMeynell:
This functionality is included in TableDriver, but it has not been updated for a long time. Will openTableDriver be supported?
[here](https://github.com/hawku/TabletDriver/blob/fd2185e0778b22839fb6b2f7598f529a8465ebaf/TabletDriverService/InputEmulator.cpp#L425)

Satanepigone:
> Worth noting that this is now indirectly supported via the plugin "[Scroll Bindings](https://github.com/Mrcubix/Scroll-Bindings/)" that you might be able to find in the Plugin Hub.

The "Scroll Bindings" you mentioned is very different from "Drag Scroll".
By Scroll binding, scroll is done with a fixed mount to a fixed direction each time when the button is pressed.
By Drag scroll, scrolling distance and direction is directly proportional to the movement of cursor/pen; like the scrolling on a smartphone screen.

chrleon:
OOh that dragscrolling worked fine @gonX! One minor thing is that it selects the text as I scroll :) 
Is it possible to perform the drag scroll with middle pen button and hover move?

EdwinGray:
> OOh that dragscrolling worked fine [@gonX](https://github.com/gonX)! One minor thing is that it selects the text as I scroll :) Is it possible to perform the drag scroll with middle pen button and hover move?

[...]

PR #4675, opened 2026-03-15
Add PenScrollingBinding

david-polak:
Added a new Binding PenScrollingBinding

The further from the initial position the pen is moved the faster the scrolling. This is the default behaviour for scrolling in wacom windows drivers.

## Configuration options
* Directions - horizontal, vertical and their inversions
* Sensitivity - how fast
* Refresh rate - how often

There's a bit of a push and pull between the sensitivity and refresh rate parameters, but the default configuration seems to work well on my device.

## Showcase

[Screencast_20260315_215656.webm](https://github.com/user-attachments/assets/5e3f9b37-d211-43a1-9d86-bc970fb85319)

### DIFF
--- a/OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
+++ b/OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
@@ -1,0 +1,138 @@
+using OpenTabletDriver.Plugin;
+using OpenTabletDriver.Plugin.Attributes;
+using OpenTabletDriver.Plugin.DependencyInjection;
+using OpenTabletDriver.Plugin.Platform.Pointer;
+using OpenTabletDriver.Plugin.Tablet;
+using OpenTabletDriver.Plugin.Timers;
+using System;
+using System.Numerics;
+
+#nullable enable
+
+namespace OpenTabletDriver.Desktop.Binding
+{
+    [PluginName(PLUGIN_NAME)]
+    public class PenScrollingBinding : IStatePositionBinding
+    {
+        private const string PLUGIN_NAME = "Pen Scrolling Binding";
+
+        private Vector2 _initialPosition = default;
+        private int _scrollAmountHorizontal = 0;
+        private int _scrollAmountVertical = 0;
+
+        [Resolved]
+        public IMouseScrollHandler? Pointer { set; get; }
+
+        private ITimer? _scrollingTimer;
+
+        [Resolved]
+        public ITimer? ScrollingTimer
+        {
+            get => _scrollingTimer;
+            set
+            {
+                if (value is null)
+                    return;
+
+                if (_scrollingTimer != null && !ReferenceEquals(_scrollingTimer, value))
+                {
+                    if (_scrollingTimer.Enabled)
+                        _scrollingTimer.Stop();
+                    _scrollingTimer.Elapsed -= Scroll;
+                    _scrollingTimer.Dispose();
+                }
+
+                _scrollingTimer = value;
+                _scrollingTimer.Elapsed += Scroll;
+            }
+        }
+
+        [BooleanPropertyAttribute("Scroll horizontally", "pen movement scrolls horizontally"), DefaultPropertyValue(true)]
+        public bool HorziontalScrollingEnabled { get; set; }
+
+        [BooleanPropertyAttribute("Scroll vertically", "pen movement scrolls vertically"), DefaultPropertyValue(true)]
+        public bool VerticalScrollingEnabled { get; set; }
+
+        [Property("Sensitivity"), ToolTip("The sensitivity of scrolling with pen movement."), DefaultPropertyValue(50f)]
+        public float Sensitivity { get; set; }
+
+        [SliderProperty("Refresh rate", 1f, 320f, 60f), ToolTip("How often scrolling event gets sent (lower for better performance, higher for smoother scrolling)."), DefaultPropertyValue(60f)]
+        public float RefreshRate { get; set; }
+
+        [BooleanPropertyAttribute("Invert horizontal axis", "invert horizontal scrolling"), DefaultPropertyValue(false)]
+        public bool InvertHorizontalAxis { get; set; }
+
+        [BooleanPropertyAttribute("Invert vertical axis", "invert vertical scrolling (MacOS like)"), DefaultPropertyValue(false)]
+        public bool InvertVerticalAxis { get; set; }
+
+        public void Press(TabletReference tablet, IDeviceReport report)
+        {
+            if (report is not IAbsolutePositionReport)
+            {
+                throw new InvalidOperationException("PenScrollingBinding not supported on this device");
+            }
+
+            ResetScrollingTimer();
+
+            _initialPosition = ((IAbsolutePositionReport)report).Position;
+            _scrollAmountHorizontal = 0;
+            _scrollAmountVertical = 0;
+
+            _scrollingTimer?.Start();
+        }
+
+        protected void ResetScrollingTimer()
+        {
+            if (_scrollingTimer == null)
+            {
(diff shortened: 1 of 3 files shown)


david-polak:
Ah, I see my editor did not pick up the .editorconfig for dotnet, lemme fix the formatting.

Mrcubix:
This should belong in another class instead, inheriting MouseScrollBinding.
Calling it "both" is misleading considering it's using the tool's position instead of a button and the usual parameters.
I'm gonna nitpick on another thing, follow the patterns used for private fields : 

```diff

-   private ITimer timer;
+   private ITimer _timer;
# ...
-    if (this.timer == null)
-        throw new InvalidOperationException($"{nameof(this.Timer)} was not injected by daemon");
+    if (_timer == null)
+         throw new InvalidOperationException($"{nameof(_timer)} was not injected by daemon");
```

david-polak:
> I'm gonna nitpick on another thing, follow the patterns used for private fields

Hmm yeah, I was looking around what the style guide seems to be, newer commits tend to `private . _field;` and older tend to `private . field`. As to `this` I can't find the pattern.

What is the style guide for this project?

> inheriting MouseScrollBinding.

Yeah I was wondering if the original implementation should be preserved, not sure what's the use case for having repeated scroll events on button hold.

There is no functional commonality between the previous implementation and this one so inheritance is out, I can create an entire new binding, any idea what to call it?

gonX:
Style guide is being worked on and is tracked in #2318 


> Hmm yeah, I was looking around what the style guide seems to be, newer commits tend to `private . _field;` and older tend to `private . field`. As to `this` I can't find the pattern.

For private fields please prepend `_` to new classes. Additionally please try to write new code in a `#nullable` context.

And `this.` usage is undecided so far. Currently it's up to the reviewer. Do what you think looks best for your individual code, but I think I prefer without.

For `this.` usage, there's the argument of confusing member field/properties with function arguments/variables but I think the general casing standards solves that; `_camelCase` for private fields/properties, `camelCase` for function args/vars, and `PascalCase` for public fields/properties.

> Yeah I was wondering if the original implementation should be preserved, not sure what's the use case for having repeated scroll events on button hold.

"I hold down my button to keep scrolling" sounds useful. If you don't have that then there would be missing functionality for holding the button, and only clicks would work. Maybe I'm misunderstanding what this PR does?

Mrcubix:
>not sure what's the use case for having repeated scroll events on button hold.

To not injure your finger repeatingly pressing tablet keys, happened once, won't let it happen easily again.

> There is no functional commonality between the previous implementation and this one so inheritance is out, I can create an entire new binding, any idea what to call it?

Not drag scrolling as you're not checking for pressure.

Some kind of name around non-pressure dependent positional scrolling.

david-polak:
Mmm thanks for reviews, moving things around, will squash and force push when I finish with it. I am leaving the `MouseScrollBinding` alone and adding a new binding `PenScrollBinding`...  or `PenScrollingBinding`

As for drag, good idea, I'll add a boolean to toggle it...

Mrcubix:
It would probably satisfy this issue : https://github.com/OpenTabletDriver/OpenTabletDriver/issues/1381

Also remembered the real reason for the interval, is that most tablets don't send aux keys states regularly, only when pressed or released.
(It's not an issue for pen buttons since it's most of the time contained in pen reports).

david-polak:
Alright, how about now? 

[Screencast_20260316_223657.webm](https://github.com/user-attachments/assets/7369a40e-145c-44ee-ade5-e75a5173cf10)

OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
```diff
@@ -0,0 +1,138 @@
...
+        public void Press(TabletReference tablet, IDeviceReport report)
+        {
+            if (report is not IAbsolutePositionReport)
+            {
+                throw new InvalidOperationException("PenScrollingBinding not supported on this device");
+            }
+
+            ResetScrollingTimer();
+
+            _initialPosition = ((IAbsolutePositionReport)report).Position;
```
gonX: I believe you can do this instead:
```suggestion
            if (report is not IAbsolutePositionReport absolutePositionReport)
            {
                throw new InvalidOperationException("PenScrollingBinding not supported on this device");
            }

            ResetScrollingTimer();

            _initialPosition = absolutePositionReport.Position;
```

OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
```diff
@@ -0,0 +1,138 @@
...
+namespace OpenTabletDriver.Desktop.Binding
+{
+    [PluginName(PLUGIN_NAME)]
+    public class PenScrollingBinding : IStatePositionBinding
+    {
+        private const string PLUGIN_NAME = "Pen Scrolling Binding";
+
+        private Vector2 _initialPosition = default;
+        private int _scrollAmountHorizontal = 0;
+        private int _scrollAmountVertical = 0;
```
gonX: styling:
```suggestion
        private Vector2 _initialPosition;
        private int _scrollAmountHorizontal, _scrollAmountVertical;
```
(value-type members are always initialized to their default on construction)

OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            _scrollingTimer?.Start();
+        }
+
+        protected void ResetScrollingTimer()
+        {
+            if (_scrollingTimer == null)
+            {
+                Log.Write(nameof(PenScrollingBinding), "Timer was not injected, can not scroll", LogLevel.Error);
+                return;
```
gonX: If the code doesn't work without, I prefer throwing an `InvalidOperationException`

OpenTabletDriver.Desktop/Binding/PenScrollingBinding.cs
```diff
@@ -0,0 +1,138 @@
...
+            else
+                _scrollAmountVertical = (int)Math.Ceiling((_initialPosition.Y - pos.Y) * (Sensitivity / 100));
+        }
+
+        public void Scroll()
+        {
+            if (Pointer == null)
+            {
+                Log.Write(nameof(PenScrollingBinding), "Pointer was not injected, can not scroll", LogLevel.Error);
+                return;
```
gonX: If the code doesn't work without the `Pointer`, I prefer throwing an `InvalidOperationException` here.

OpenTabletDriver.Plugin/IStatePositionBinding.cs
```diff
@@ -0,0 +1,11 @@
+
+using OpenTabletDriver.Plugin.Tablet;
+using System.Numerics;
+
+namespace OpenTabletDriver.Plugin
+{
+    public interface IStatePositionBinding : IStateBinding
+    {
+        public void SetPosition(Vector2 pos);
```
gonX: styling nitpick (that I haven't been good at following myself)
```suggestion
        void SetPosition(Vector2 pos);
```
(it's implicit from the interface visibility)

gonX (dismissed):
A quick surface-level review, probably more later.

It's all nitpicks as generally your code is very reasonable to read 👍

gonX (commented):
Code looks perfect now. I'll have to test this on the other side of v0.6.7 - we just closed for feature additions for that release about a week ago.

david-polak:
Verified working under latest `0.6.x`. @gonX is it the right time to merge?

david-polak:
Hi @gonX, can we merge?

gonX:
This'll be reviewed again eventually once v0.6.8 releases.

arindavoodian:
It would be great to get this merged in as the reason I stopped using OpenTabletDriver is because it was missing this feature. Thanks!

PR #4680, opened 2026-03-18
Editorconfig small

### LINKED ISSUE #2318: Expand `.editorconfig` with new conventions

gonX:
Clean up codebase with C#'s suggested coding conventions. I'll skip the ones that go against what we prefer and will try to note them here.

**Try to get this in early for v0.6.8**

Can be verified by checking out `.editorconfig` from branch onto target branch and running `dotnet format OpenTabletDriver.sln` and then diffing this PR's contents with your results.

fixes #2318 

## Recommended style rules that deviate from our preference:

- [IDE0045](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0045)/[IDE0046](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0046): Use conditional expression (ternary) for assignment and return
  - Not set as warning because the changes it makes are sometimes less readable

## Style rules that have been chosen not to be enforced (yet):

- [IDE0003/IDE0009](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009): `this.` preferences
  - Enforcing `this.` qualification (or enforcing removal) has arguments for both ways. I think this should be handled on a case-by-case basis.

---

personal gonX notes:

for later rebases:
`git rebase -i origin/0.6.x`, change all to `edit`, into `dotnet format OpenTabletDriver.sln && git add . && git commit --amend --no-edit && git rebase --continue`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -65,6 +65,7 @@ dotnet_style_prefer_inferred_tuple_names = true:warning
 dotnet_style_prefer_inferred_anonymous_type_member_names = true:warning
 dotnet_style_prefer_auto_properties = true:warning
 dotnet_style_prefer_compound_assignment = true:warning
+dotnet_style_prefer_simplified_interpolation = true:warning
 
 [*.{sh,md,csproj,json,yml}]
 indent_size = 2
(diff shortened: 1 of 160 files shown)


gonX:
It looks like there's an SDK change which changes one of the tests that my local version from Arch Linux repositories isn't picking up. I'll continue work one on this once the package has been updated.

PR #4689, opened 2026-03-21
Reintroduce `IPointerProvider<T>`

gonX:
Apparently some plugins were actually inheriting this. This PR reintroduces the type but keeps it uninherited in the output modes and marks the interface as obsolete.

Soft-fixes regression introduced with breaking change in #4582


Kuuuube:
Cubix (as far as we know the only user of this in plugins) has put in a fix for his plugins with https://github.com/OpenTabletDriver/Plugin-Repository/pull/149.

This isnt entirely necessary anymore.

Kuuuube (approved):

gonX:
Going to merge this just to keep API sane. It will be completely removed again in an API bump.

Should the output mode inheritance for some reason really be needed we should be able to simply revert the last commit in this PR and release a bugfix version, though I don't envision that being necessary.

PR #4694, opened 2026-03-21
PR Labeler: Add Evdev glob to linux/gtk label

### LINKED ISSUE #4343: 0.6.x CI: `linux/gtk` label does not account for filenames with `Evdev` in them
#4342 changed Linux binding options but did not get labeled correctly with `linux/gtk`

I think we can safely add `evdev` to the `linux/gtk` PR labeler glob.

juan-martinez-pf:
Fixes #4343

Added **/*Evdev* to the linux/gtk glob list in pr-labeler.yml.

The existing labeler test script passes. Though it only verifies that globs match at least one file, not that all relevant files are covered, so it was passing before too. The issue was just silent mislabeling.


gonX (approved):
LGTM

PR #4695, opened 2026-03-21
[Avalonia] Build: Fix invalid regex for git dirty detection

juan-martinez-pf:
Fixes #4178

- Changed `^*dirty*$` to `^.*dirty.*$` in `eng/bash/lib.sh` on the avalonia branch.
- Syntax check passes and the regex correctly matches dirty suffixes.
- The `0.6.x` branch already has the correct regex.


gonX (approved):
LGTM

PR #4698, opened 2026-03-22
WindowsInput: Log errors when SendInput fails to insert events

### LINKED ISSUE #4531: Windows `SendInput` usage should not trigger CA1806 (unused return value)
### Description

[Roslyn Analyzer CA1806](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1806)

This might help track down issues with users not being able to use virtual keyboards, but if nothing else then it's just good practice to ensure any eventual error codes are bubbled up to the log.

Occurs in the following 2 spots in the code:

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/f59853955c36bf2ba6e0a2404ec6ca795d508689/OpenTabletDriver.Desktop/Interop/Input/Keyboard/WindowsVirtualKeyboard.cs#L33

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/f59853955c36bf2ba6e0a2404ec6ca795d508689/OpenTabletDriver.Desktop/Interop/Input/WindowsVirtualMouse.cs#L97

Implementers should note the return value and remarks section of the [`SendInput`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendinput) documentation.

### Acknowledgements

juan-martinez-pf:
Fixes #4531

## Summary
- Check the return value of `SendInput` in `WindowsVirtualKeyboard` and `WindowsVirtualMouse`
- Log an error when the number of successfully inserted events doesn't match the expected count
- Follows existing logging conventions used by Evdev input classes

## Context
Per the [SendInput docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendinput), the return value indicates how many events were actually inserted. A mismatch could help diagnose issues where virtual keyboards or mice silently fail.

## Concern Regarding User Privilege  
Note that [UIPI](https://learn.microsoft.com/en-us/windows/win32/winmsg/user-interface-privilege-isolation) blocks are not detectable. Windows intentionally reports neither the return value nor `GetLastError` when UIPI is the cause. Since UIPI is the most common reason for `SendInput` to silently fail, it may be worth including a hint in the log message (e.g., "try running as administrator") to guide users toward the fix.

Let me know if you want tme to change the log to this:

`Log.Write("WindowsKeyboard", $"SendInput failed: {result}/{inputs.Length} events inserted. If targeting an elevated application, try running as administrator.", LogLevel.Error);`

## Test plan
- Verified the project builds with 0 warnings and 0 errors (`build.sh macos`)
- Runtime testing requires a Windows environment

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Keyboard/WindowsVirtualKeyboard.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Keyboard/WindowsVirtualKeyboard.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using OpenTabletDriver.Native.Windows;
 using OpenTabletDriver.Native.Windows.Input;
+using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Platform.Keyboard;
 
 namespace OpenTabletDriver.Desktop.Interop.Input.Keyboard
@@ -30,7 +31,9 @@ namespace OpenTabletDriver.Desktop.Interop.Input.Keyboard
             };
 
             var inputs = new INPUT[] { input };
-            SendInput((uint)inputs.Length, inputs, INPUT.Size);
+            var result = SendInput((uint)inputs.Length, inputs, INPUT.Size);
+            if (result != inputs.Length)
+                Log.Write("WindowsKeyboard", $"SendInput failed: {result}/{inputs.Length} events inserted", LogLevel.Error);
         }
 
         public void Press(string key)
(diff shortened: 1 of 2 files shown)


gonX (approved):
This looks correct to me

We don't want to encourage using administrator due to the plugin API existing (which can run arbitrary code) and also the usual weirdness with the GUI when run in administrator.

juan-martinez-pf:
Yes, I didn't think asking for users to run as administrator was the best solution. I converted to draft while I waited for a response as you approved it at the same time.

gonX:
I think this is safe to merge now. Merging.

Issue #2836, opened 2023-07-13
Modifier keys for buttons


jonhoo:
### Description

It'd be great to be able to map buttons to modifier + mouse button combos. For example, map auxiliary button 1 to shift + left mouse button!

### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

jamesbt365:
Duplicate of #1769.

Issue #1769, opened 2021-12-19
Key + Mouse multi-key bindings


gonX:
## Description

Multi-key should support both keyboard and mouse inputs simultaneously. Also see #1768

amanda-santini:
Thanks guys. If this is of any help, I was kind of able to do it via configuration file.
xsetwacom set "HUION Huion Tablet_H320M Pad pad" button 15 "key +Alt button +3"

However, it doesn't work with the stylus tip (hovering over the tablet) for some reason, it only works if I press the tablet key and if I move my actual mouse. Super weird.

PR #4700, opened 2026-03-24
Add Key + Mouse combo binding type

dsarno:
## Summary
- Adds a new `KeyMouseBinding` binding type that combines keyboard key presses with a mouse button click in a single binding
- Adds a `listkeys` console command that lists all valid key names for the current platform
- Implements the feature requested in #1769 and #2836 for the 0.6.x branch

## Motivation
Users working in 3D applications (Blender, Maya, ZBrush, Substance Painter) commonly need modifier+mouse combos like Shift+MMB, Ctrl+MMB, or Alt+MMB bound to tablet buttons for viewport navigation. Currently neither the GUI nor CLI supports this — the binding editor handles key combos and mouse buttons as separate, mutually exclusive binding types.

## Configuration example
```json
{
  "Path": "OpenTabletDriver.Desktop.Binding.KeyMouseBinding",
  "Settings": [
    { "Property": "Keys", "Value": "Shift" },
    { "Property": "Button", "Value": "Middle" }
  ],
  "Enable": true
}
```

## Architectural note
This PR takes the pragmatic approach of adding a concrete binding type, following the same patterns as the existing `MultiKeyBinding` and `MouseBinding` classes. We recognize this introduces a special-case combo class rather than a more general compositional mechanism.

A potentially cleaner alternative would be a general-purpose `CompositeBinding` / `ChordBinding` that wraps and fires multiple child bindings together — this would avoid per-combo-type class proliferation and cover future patterns (key+key, mouse+mouse, modifier+wheel, etc.) without new types. However, that would be a larger architectural change touching serialization, the GUI editor, and the plugin system.

Happy to take guidance from maintainers on the preferred direction — whether this concrete approach is acceptable for 0.6.x, or whether you'd prefer a composite abstraction (which might align better with the v0.7.0 plans in #1769).

## Test plan
- [x] Tested on macOS (ARM64) with XP-Pen Deco 01 V3
- [x] Verified Shift+Middle Mouse binding works correctly in Blender
- [x] Verified `listkeys` outputs all valid key names
- [x] Existing test suite passes (12 pre-existing Linux-specific failures unrelated)
- [ ] Test on Windows
- [ ] Test on Linux


dsarno:
Fixed linting issue

Kuuuube:
This feels more like something that should be a plugin and rather limiting to pull directly into the driver.

dsarno:
> This feels more like something that should be a plugin and rather limiting to pull directly into the driver.

Sure thing, I can rework it as a plug-in. Thanks for your feedback.

dsarno:
Reworked this as a standalone plugin per @Kuuuube's feedback: https://github.com/dsarno/OTD-KeyMouseBinding

Submitted to the Plugin-Repository: https://github.com/OpenTabletDriver/Plugin-Repository/pull/150

PR #4712, opened 2026-03-29
Linux: Add raw hidraw fallback for devices with unparseable HID descriptors

Paberu85:
## Summary
- Adds `LinuxRawHidStream`, a fallback `IDeviceEndpointStream` that uses direct file I/O and ioctl for devices whose HID report descriptors cannot be parsed by HidSharp (e.g. `UnitExponent` encoded as a signed byte instead of a 4-bit nibble)
- Adds `LinuxRawHidDescriptorInfo`, a tolerant HID descriptor parser that extracts only report lengths and report-ID usage, bypassing the metadata items that cause HidSharp to throw
- `HidSharpEndpoint.Open()` falls back to `LinuxRawHidStream` when `TryOpen()` fails on Linux
- `HidSharpEndpoint` report length properties now fall back to the raw descriptor parser, so `Driver.cs` device matching works even when HidSharp returns -1

## Motivation
HidSharpCore's `Unit.DecodeExponent()` throws `ArgumentOutOfRangeException` for devices that encode `UnitExponent` as a signed byte (values like `0xFD`, `0xFE`) rather than the 4-bit nibble the parser expects. This causes `TryOpen()` to fail and report lengths to return -1, making the device undetectable. The proper fix belongs upstream in HidSharpCore; this is a workaround within OTD.

## Test plan
- [x] Builds cleanly on Linux
- [x] 25 new/existing tests pass (`LinuxRawHidDescriptorInfoTest`, `LinuxRawHidStreamTest`, `HidSharpEndpointTest`)
- [x] Verified with Waltop Sirius Battery Free Tablet (172f:0502) — device with malformed descriptor
- [ ] Verify no regression on devices that work normally with HidSharp

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/OpenTabletDriver.Tests/Fakes/FakeHidDevice.cs
+++ b/OpenTabletDriver.Tests/Fakes/FakeHidDevice.cs
@@ -1,3 +1,4 @@
+using System;
 using HidSharp;
 
 namespace OpenTabletDriver.Tests.Fakes
@@ -15,6 +16,11 @@ namespace OpenTabletDriver.Tests.Fakes
         private int _productId;
         private int _vendorId;
         private int _releaseNumberBcd;
+        private byte[] _rawReportDescriptor = Array.Empty<byte>();
+        private Exception? _openException;
+        private Exception? _maxInputReportLengthException;
+        private Exception? _maxOutputReportLengthException;
+        private Exception? _maxFeatureReportLengthException;
 
         public override string DevicePath => _devicePath;
         public override bool CanOpen => _canOpen;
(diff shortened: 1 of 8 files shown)


Kuuuube:
Issues in hidsharp should be fixed in hidsharp not reimplemented like this.

Paberu85:
> Issues in hidsharp should be fixed in hidsharp not reimplemented like this.

The problem is the library hasn’t been updated in two years, and it’s pretty unlikely a proper fix PR will ever get merged into the main branch. That’s why this workaround exists. I’d leave it as a draft instead of closing it.

Kuuuube:
It will get merged if it needs to get merged for OTD.

PR #4714, opened 2026-03-30
Remove note that macos is experimental

Kuuuube:
No need for this anymore. Handled for all languages since it's very obvious where the note is in all of them.


gonX (approved):
LGTM

PR #4716, opened 2026-03-30
Use radio button for hex/binary select in tablet debugger

Kuuuube:
Noticed that it's possible to uncheck both hex and binary. This fixes that.

Eto radio buttons are designed in the most obtuse way I've ever seen radio buttons designed. Theres a root button and all other buttons piggyback off that root button instead of simply all the buttons go into the same collection. There's also no docs stating this is why the constructors are made to take a radiobuttonitem, despite this way of doing things being totally different from how anything else does it.

~~Havent tested on windows yet.~~

### DIFF
--- a/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
+++ b/OpenTabletDriver.UX/Windows/Tablet/TabletDebugger.cs
@@ -265,14 +265,17 @@ namespace OpenTabletDriver.UX.Windows.Tablet
                 Text = "Raw Data Mode",
             };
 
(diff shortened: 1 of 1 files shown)


Kuuuube:
Tested on windows now. Works fine.

gonX (approved):
LGTM. Had to double check the [docs](https://pages.picoe.ca/docs/api/html/T_Eto_Forms_RadioMenuItem.htm), a bit confusing how it's set up but this makes sense when it's in a foreach loop.

PR #4717, opened 2026-03-30
Use `--no-self-contained` in powershell build script

Kuuuube:
This technically is a breaking change since I'm removing the `$selfContained` arg but since using this script for packaging is deprecated it shouldnt matter. It's only used for convenience when testing.

Same deal as #4623


gonX (approved):
Yeah I don't mind losing that arg here. LGTM

PR #4721, opened 2026-04-02
Fix macOS Gatekeeper "app is damaged" error

juan-martinez-pf:
Newer .NET SDK adhoc signs individual Mach-O executables during publish, creating a partially signed bundle that macOS treats as tampered ("damaged") rather than unsigned.

This signs the .app bundle with `rcodesign` during packaging, restoring the "unidentified developer" behavior from 0.6.6.x.

Uses `rcodesign` so the build stays on a Linux runner.

### DIFF
--- a/eng/bash/macos/package.sh
+++ b/eng/bash/macos/package.sh
@@ -18,7 +18,13 @@ cp "${pkg_script_root}/Icon.icns" "${pkg_root}/Contents/Resources/"
 cp "${pkg_script_root}/Info.plist" "${pkg_root}/Contents/"
 
 echo "Signing app bundle..."
-rcodesign sign "${pkg_root}"
+if hash rcodesign 2>/dev/null; then
+  rcodesign sign "${pkg_root}"
+elif hash codesign 2>/dev/null; then
+  codesign --deep --force --sign - "${pkg_root}"
+else
+  echo "Warning: neither rcodesign nor codesign found, skipping signing"
+fi
 
 echo "Creating tarball..."
 create_binary_tarball "${pkg_root}" "${OUTPUT}/${PKG_FILE}"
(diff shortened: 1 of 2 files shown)


gonX:
Looks like the macOS runner needs some changes to be fully supported. Installing a newer BASH version in the runner during runner setup would be preferable

juan-martinez-pf:
> Looks like the macOS runner needs some changes to be fully supported. Installing a newer BASH version in the runner during runner setup would be preferable

Good catch! I added a bash update step.

juan-martinez-pf:
I hate build systems. It always turns into a rabbit hole. I'm adding coreutils as well as updating bash, so that it's compatible with the `readlink -f` command.

juan-martinez-pf:
In effort to keep using a linux runner, I switched the code signing to `rcodesign`. I have updated the branch and PR description to reflect this.

eng/bash/macos/package.sh
```diff
@@ -17,6 +17,9 @@ mkdir -p "${pkg_root}/Contents/Resources"
 cp "${pkg_script_root}/Icon.icns" "${pkg_root}/Contents/Resources/"
 cp "${pkg_script_root}/Info.plist" "${pkg_root}/Contents/"
 
+echo "Signing app bundle..."
+rcodesign sign "${pkg_root}"
+
```
gonX, line 28: This should probably check for presence of either `codesign` or `rcodesign`, starting with `rcodesign`, so that it still builds on native macOS. Pseudocode:

```pseudo
if binary rcodesign exists:
    use rcodesign
else if binary codesign exists:
    use codesign
else:
    warn user that we cannot sign
```

gonX (changes requested):
General idea looks good to me, but I'd like the shell script to also work on native macOS

eng/bash/macos/package.sh
gonX: You can e.g. use `type -P rcodesign >/dev/null` or `hash rcodesign 2>/dev/null` (replace `rcodesign` with your desired binary) to determine if a command is on the path, BTW.

juan-martinez-pf: You know I had this fallback, and then I removed it before committing. 🤣

I'll do the change.

gonX (approved):
LGTM

gonX:
I can't see any way this is going to break so I'm going to merge this without Kuuube's review

PR #4727, opened 2026-04-06
Display units on pen and wheel thresholds

Kuuuube:
`UnitGroup` is in a bit of a weird location in the codebase since it was only intended to be used in the area editor. Moved it to `Generic` (same area `Group` is in) since I think it makes more sense there.

Tested and working on windows and linux.


gonX (approved):
I didn't test it but the code changes make sense to me

PR #4728, opened 2026-04-06
Note rcodesign requirement for macos packaging outside of macos

Kuuuube:


gonX (approved):
Makes sense. LGTM

Introduced in #4721 for context

PR #4730, opened 2026-04-07
Set `SINGLE_FILE` to true for macos for signed builds, disable signing by default

Kuuuube:
Adds an option for signing on macos but run `/p:_EnableMacOSCodeSign=false` by default. Also avoid installing rcodesign by default by defining `macos-signed` as a build target that is required for installing rcodesign.

`SINGLE_FILE` is true by default in the build system but since macos appears to require this for signing and breaks without it, I'd rather have it explicitly. Note that this is not a perfect solution since it increases the macos bundle size by 3x due to packing all dependencies in each executable separately rather than letting them share since macos needs to be self contained.

Confirmed signing and single file makes installation fail in the same way it did previously.
https://discord.com/channels/615607687467761684/628651971267788800/1491195026284347402

Confirmed unsigned also works with these changes.
https://discord.com/channels/615607687467761684/628651971267788800/1491457514712207470

### DIFF
--- a/.github/actions/build-wrapper/action.yml
+++ b/.github/actions/build-wrapper/action.yml
@@ -27,7 +27,7 @@ runs:
           git config --system --add safe.directory /__w/OpenTabletDriver/OpenTabletDriver
 
     - name: "Install rcodesign (macOS builds)"
-      if: inputs.package-name == 'macos'
+      if: inputs.package-name == 'macos-signed'
       shell: bash
       run: cargo install apple-codesign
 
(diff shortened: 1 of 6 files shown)


juan-martinez-pf:
Tested locally on macOS (M1 Max, macOS 26.4).
Built from the PR branch with `rcodesign` available:

- `codesign --verify --deep --strict` passes: valid on disk and satisfies Designated Requirement
- Gatekeeper shows "unidentified developer" warning not "damaged". Same behavior as 0.6.6.x
- Single-file output confirmed: 3 executables in Contents/MacOS/
- Bundle size ~221MB up from ~75MB, about 3x like you mentioned in the PR.

rcodesign + SINGLE_FILE works correctly on macOS.
While I dislike the increase in build size, SINGLE_FILE should give faster loading and is also recommended for macOS distribution.

juan-martinez-pf:
Uh, I didn't do a clean build. Let me know if that should have been tested.

gonX (approved):
Code changes LGTM

PR #4732, opened 2026-04-07
DeviceStringReader: Unset dropdown when bound text fields change

gonX:
Avoids confusion when changing numbers without the dropdown changing - same type of fix as #4718

cc @Kuuuube


Kuuuube (approved):
Looks like witchcraft but alright. Tested and works.

PR #4733, opened 2026-04-07
Actually assign to `counterClockwiseThreshold.StepSize`

Kuuuube:
I'll ignore how cursed this pattern is.

Assuming this is some bad pasting. `counterClockwiseThreshold.StepSize` never got set and the slider would not snap to steps properly.


gonX (approved):
Yeah that's definitely a bad paste. Nice job spotting that - I couldn't figure out why it was inconsistently bugging out (but I never connected the dots that it was only happening on one of them)

PR #4741, opened 2026-04-11
doc: be more precise about the fork

jerome-labidurie:
When creating a fork, by default github fork only the 0.6.x branch
image

This prevent the forked (then cloned) repository to be build
```bash
jjl@guiautec:/tmp/OpenTabletDriver$ ./build.sh linux
fatal: No names found, cannot describe anything.
jjl@guiautec:/tmp/OpenTabletDriver$ git branch 
* 0.6.x
jjl@guiautec:/tmp/OpenTabletDriver$ git describe --tags 
fatal: No names found, cannot describe anything.
```

PR #4742, opened 2026-04-11
Use `--always` on git describe to fallback when no tags are present

Kuuuube:
Currently the build script will fail if run on a git repo with no tags present. This fixes that.

Git describe fallback output looks something like this `ad3e21f1-dirty`, normal output being `v0.6.6.2-879-g3d11d1dc`. This causes the distance from tag check to fail which is expected but doesn't break the build.

Version suffix looks a bit silly but it works: `-ad3e21f1-dirty-dirty`, normal: `+887-gb35caa5b`.

Related: #4741

### DIFF
--- a/eng/bash/lib.sh
+++ b/eng/bash/lib.sh
@@ -50,7 +50,7 @@ if [ -z "$VERSION_SUFFIX" ]; then
   hash sed &>/dev/null && \
   git rev-parse --is-inside-work-tree &>/dev/null
   then
-    GIT_DESCRIBE="$(git describe --long --tags --dirty)"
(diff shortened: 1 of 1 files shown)


gonX:
It seems pretty unlikely that the git repo never contains any tags. `git-pull` documentation describes:
> Most of the tags are fetched automatically as branch heads are
       downloaded, but tags that do not point at objects reachable from
       the branch heads that are being tracked will not be fetched by
       this mechanism

I can't easily test this but I am willing to merge if more users start reporting the same issue.

gonX (approved):
With that being said it's pretty unlikely that this change is problematic. Merging.

(at worst, repos missing tags will have a potentially confusing version but should still work)

PR #4750, opened 2026-04-15
Ensure name and manufacturer are not null

Kuuuube:
Untested fix based on this trace (truncated):
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at OpenTabletDriver.UX.Windows.DeviceStringReader.<>c.<.ctor>b__0_8(SerializedDeviceEndpoint x) in D:\a\OpenTabletDriver\OpenTabletDriver\OpenTabletDriver.UX\Windows\DeviceStringReader.cs:line 122
   at Eto.Wpf.Forms.Controls.WpfTextBindingBlock.Convert(Object value, Type targetType, Object parameter, CultureInfo culture)
```
[message-13.txt](https://github.com/user-attachments/files/26736380/message-13.txt)

Noticed that `x.Manufacturer` likely can be null as well which would cause this to throw too.

```csharp
string name = null;
string manufacturer = null;
name.Contains(""); // obviously throws, the error we see in the above trace
"".Contains(manufacturer); // also throws, should be possible to hit too
```

Could also go for something like this, just kept it as close to the original pattern instead:
```csharp
string name = x.ProductName ?? x.FriendlyName ?? "null";
string title = name;
if (x.Manufacturer != null && !name.Contains(x.Manufacturer))
{
    title = $"{x.Manufacturer} {name}";
}
```

### DIFF
--- a/OpenTabletDriver.UX/Windows/DeviceStringReader.cs
+++ b/OpenTabletDriver.UX/Windows/DeviceStringReader.cs
@@ -118,8 +118,9 @@ namespace OpenTabletDriver.UX.Windows
             this.deviceDropDown.ItemTextBinding = Binding.Delegate<SerializedDeviceEndpoint, string>(x =>
             {
                 // don't include manufacturer if it's already in the product name (e.g. Razer)
-                string name = x.ProductName ?? x.FriendlyName;
-                string title = name.Contains(x.Manufacturer) ? name : $"{x.Manufacturer} {name}";
+                string name = x.ProductName ?? x.FriendlyName ?? "null";
+                string manufacturer = x.Manufacturer ?? "null";
+                string title = name.Contains(manufacturer) ? name : $"{manufacturer} {name}";
 
                 if (title.Length >= 32) // truncate if too long, used in GUI
                     title = title[..29] + "...";


gonX:
Looking a HidSharp source, this seems to theoretically be the correct fix. The manufacturer field can in some cases indeed be nullable, but it seems that the intention is to ensure that null fields instead returns a fallback value as seen in `HidSharpEndpoint` here:

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/eeb2d92362d85c9b9fd66894f4165abb8faa56c4/OpenTabletDriver/Devices/HidSharpBackend/HidSharpEndpoint.cs#L29

However, if we then follow the code called from the above snippet, especially the `SafeGet` code, you'll see that it uses the value from `TryGet` if the provided predicate didn't throw an exception:
https://github.com/OpenTabletDriver/OpenTabletDriver/blob/ce15d024bb7190dd5f08cb01342cf326276240f5/OpenTabletDriver/Extensions.cs#L8-L23

The provided predicate calls `GetManufacturer()` in HidSharp's selected abstraction - for example, `LinuxHidDevice` has the following implementation that will effectively throw an exception if the manufacturer value is null or unavailable:
https://github.com/InfinityGhost/HIDSharpCore/blob/944a386ea6d51480ccd7dde7bb78e78768da8e1d/HidSharp/Platform/Linux/LinuxHidDevice.cs#L120-L124
(no embed as the provided source link is outside this repo)

On the contrary, the `WindowsHidDevice` implementation will freely allow null to be returned:
https://github.com/InfinityGhost/HIDSharpCore/blob/4cc70c201a3fb319de12f3ab4fa90f15bccb8835/HidSharp/Platform/Windows/WinHidDevice.cs#L159-L163

The `RequiresGetInfo` call only returns a boolean value to indicate an error:
https://github.com/InfinityGhost/HIDSharpCore/blob/4cc70c201a3fb319de12f3ab4fa90f15bccb8835/HidSharp/Platform/Windows/WinHidDevice.cs#L87-L99
Sadly, the return value is not used at all as seen in the former link, so there is a chance that `WinHidDevice` will indeed return a null string.

Therefore, a null value (or failure to look up) on Windows will actually very likely indeed return a null value here.

I think it might make sense to also mark the source fields as being nullable, so that null reference static analysis can help us avoid this in the future. #4563 will very likely catch these types of errors if it's implemented correctly (and if nothing else, this backreference can help on deciding the correct nullability type of these fields for #4563).

This was a very long winded way to say LGTM.

gonX (approved):

PR #4752, opened 2026-04-16
Show installed plugins with checkmark to differentiate

Kuuuube:
Only tested on Linux. Shouldn't have any potential of breaking anything since it's just editing a string but worth checking that it looks good elsewhere.

Example of how this looks:

image


gonX (approved):
Tested on Linux. Looks good.

Kuuuube:
How this looks on windows:
image

Not quite as elegant. Might want to change to a different checkmark symbol as also discussed in discord for the debugger checkmark. Not sure.

PR #4753, opened 2026-04-16
Use base variant of heavy checkmark instead of emoji variant in debugger

Kuuuube:
image

https://www.unicode.org/emoji/charts/emoji-variants.html

Comparison in OTD:
image


gonX (approved):
I agree that it looks better. Also code LGTM

PR #4756, opened 2026-04-19
Add Creators field for plugin metadata

Kuuuube:
I don't really like that when I take over maintaining a plugin it essentially strips all credit from the creator unless they have their name in the plugin's name. Or something is added to the description (clunky).

### DIFF
--- a/OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
+++ b/OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
@@ -91,6 +91,11 @@ namespace OpenTabletDriver.UX.Windows.Plugins
                         },
                         new AlignedGroup
                         {
+                            Text = "Creator",
+                            Content = creator = new Label()
+                        },
+                        new AlignedGroup
+                        {
                             Text = "Description",
                             Content = description = new Label
                             {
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
```diff
@@ -140,8 +140,10 @@ public MetadataViewer()
                 }
             };
 
+
+
             name.TextBinding.Bind(MetadataBinding.Child(c => c.Name));
```
gonX: lint please

OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
```diff
@@ -140,8 +140,10 @@ public MetadataViewer()
...
-            owner.TextBinding.Bind(MetadataBinding.Child(c => c.Owner));
+            owner.TextBinding.Bind(MetadataBinding.Child(c => c.Creators != null && c.Creators.Count > 0 ? $"{c.Owner} ({String.Join(" ,", c.Creators)})" : c.Owner));
```
gonX: I'd prefer adding a new field in the GUI.

Also I haven't tested the code directly but does this actually work? I thought you'd need to pass this through some delegate via `.Convert()`, since the `Binding.Child()` function _should_ only be taking in an expression tree as far as I know, which are generally pretty limited, see https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/expression-trees/#limitations. Notably, interpolated strings (`$"hello {var}"`) aren't supported here.

gonX (changes requested):
Idea good, but I'd like this to be a new GUI field instead of piggybacking off the owner field.

OpenTabletDriver.UX/Windows/Plugins/MetadataViewer.cs
Kuuuube: It indeed does work lol.

The reason I'm not sure about adding another field is it's either going to be empty or duplicate of the owner field almost always. But I suppose it's fine.

Kuuuube:
Here's what it looks like as a new field. Fallback to match `owner` if `creators` isn't present.

image


I'm not 100% in love with the term "creators" but not sure what else could go there.

Kuuuube:
Pushing this to the closer milestone. I'd really like to have this in sooner and it shouldn't be able to break anything.

Rebased for nrt.

gonX (approved):
LGTM. We can always adjust the fallback later if it ends up being confusing.

PR #4757, opened 2026-04-19
Fix eto sending back input text as null sometimes but empty string if the user touched it

Kuuuube:
eto moment

This is the `Use alternate source...` thing in the plugin manager.


gonX (approved):
LGTM. Unless I'm undercaffeinated I'm pretty sure the extra null check isn't necessary (`this.inputText?.Length` -> `this.inputText.Length`) but some mass-code cleanup PR will eventually remove that. The code isn't in a hot path so I'm not insisting on fixing that now.

Merging now as this should be fine to put in for v0.6.7.

Kuuuube:
> I'm pretty sure the extra null check isn't necessary

I thought so too and I don't understand why it isn't but it just crashed when I didn't put that there.

PR #4758, opened 2026-04-20
Add support for MTE-450 wheel

Kuuuube:
I own one of these.

Only aux: [tablet-data_1776709746.txt](https://github.com/user-attachments/files/26907255/tablet-data_1776709746.txt)
Aux and pen: [tablet-data_1776709892.txt](https://github.com/user-attachments/files/26907266/tablet-data_1776709892.txt)

This tablet is kindof annoying since it doesn't separate out aux and position/pressure reports if the pen is on the tablet so they need to be parsed in both parser paths.


gonX (approved):
LGTM. Assuming this is perfect since you own the tablet.

PR #4759, opened 2026-04-20
Dont use subtraction in wacom wheel parsing

Kuuuube:
Chasing out the devil from these parsers.
image

Exact same handling for the exact same reported values in the PTH-x60 in `IntuosV2AuxReport` (`0x80 -> 0xC7`).


gonX (approved):
Seems correct to me

PR #4765, opened 2026-04-25
Bump actions versions to node 24 supported versions

Kuuuube:
"Node.js 20 will be removed from the runner on September 16th, 2026."

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/


gonX (approved):
I skimmed the upstream documentation for any breaking changes and didn't see any so this LGTM.

PR Labeler also needs an update. I'll handle that in a separate PR.

PR #4776, opened 2026-04-28
WindowsDisplay: Emit status of DPI Awareness on init

gonX:
As seen in #4562 (3d2dc07)

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
+++ b/OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Numerics;
 using System.Runtime.InteropServices;
--- a/OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
+++ b/OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Numerics;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
```diff
@@ -17,10 +17,16 @@ public WindowsDisplay()
...
-                Log.Debug("Display", "DPI Awareness enabled");
+                int dpiResult = SetProcessDpiAwareness(2);
+                if (dpiResult == 0) // S_OK
+                    Log.Debug("Display", "DPI Awareness enabled");
+                else
+                    Log.Write("Display", $"Could not enable DPI awareness (HRESULT: 0x{dpiResult:X8})");
+            }
+            catch
+            {
+                Log.Write("Display", "Silently catching exception while setting DPI awareness");
```
Kuuuube, line 28: Is it silent if we're logging it?

Kuuuube (approved):

Kuuuube:
Also idk how much verification you want but I can verify it logs failures just fine. Tested on windows vm:
```
[Display:Info]	Could not enable DPI awareness (HRESULT: 0x80070005)
```

OpenTabletDriver.Desktop/Interop/Display/WindowsDisplay.cs
gonX: Oopsie daisy

gonX:
> Also idk how much verification you want but I can verify it logs failures just fine. Tested on windows vm:
> 
> ```
> [Display:Info]	Could not enable DPI awareness (HRESULT: 0x80070005)
> ```

At the very least this does exactly as it does before, except it logs what happens exactly.

It throwing an access denied (which is what that `HRESULT` means) is however surprising to me

gonX:
Forgot to add to above, since the PR only changes logging behavior (and not actual behavior) we can go ahead with merging this, and debug the access denied result above separately.

PR #4781, opened 2026-04-29
Add wheel bindings to console GetBindings

Kuuuube:
Ends up looking like this:
```
Wheel Bindings:
Wheel 1 Button Bindings: [None]
Wheel 1 Clockwise Rotation: []@15°
Wheel 1 Counter-Clockwise Rotation: []@15°
```
```
Wheel Bindings:
Wheel 1 Button Bindings: ['Key Binding: { Key: C }']
Wheel 1 Clockwise Rotation: ['Key Binding: { Key: A }']@5°
Wheel 1 Counter-Clockwise Rotation: ['Key Binding: { Key: B }']@5°
```
```
Wheel Bindings:
Wheel 1 Button Bindings: ['Key Binding: { Key: C }']
Wheel 1 Clockwise Rotation: ['Key Binding: { Key: A }']@5°
Wheel 1 Counter-Clockwise Rotation: ['Key Binding: { Key: B }']@5°
Wheel 2 Button Bindings: ['Key Binding: { Key: F }']
Wheel 2 Clockwise Rotation: ['Key Binding: { Key: D }']@5°
Wheel 2 Counter-Clockwise Rotation: ['Key Binding: { Key: E }']@5°

```
```
Wheel Bindings: None
```

Not sure if this is the best way to display this.

### DIFF
--- a/OpenTabletDriver.Console/Extensions.cs
+++ b/OpenTabletDriver.Console/Extensions.cs
@@ -66,7 +66,11 @@ namespace OpenTabletDriver.Console
             var wheelIndex = 0;
             foreach (var wheelBinding in wheelBindings)
             {
-                yield return $"Wheel {wheelIndex + 1} Button Bindings: [{string.Join(", ", wheelBinding.WheelButtons.Format())}]\nWheel {wheelIndex + 1} Clockwise Rotation: [{wheelBinding.ClockwiseRotation.Format()}]@{wheelBinding.ClockwiseActivationThreshold}°\nWheel {wheelIndex + 1} Counter-Clockwise Rotation: [{wheelBinding.CounterClockwiseRotation.Format()}]@{wheelBinding.CounterClockwiseActivationThreshold}°";
+                yield return $"""
+                            Wheel {wheelIndex + 1} Button Bindings: [{string.Join(", ", wheelBinding.WheelButtons.Format())}]
+                            Wheel {wheelIndex + 1} Clockwise Rotation: [{wheelBinding.ClockwiseRotation.Format()}]@{wheelBinding.ClockwiseActivationThreshold}°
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Console/Extensions.cs
```diff
@@ -58,5 +59,16 @@ public static IEnumerable<string> Format(this IEnumerable<PluginSettingStore> st
...
             }
         }
+
+        public static IEnumerable<string> Format(this IEnumerable<WheelBindingSettings> wheelBindings)
+        {
+            if (!wheelBindings.Any()) { yield return "None"; }
+            var wheelIndex = 0;
+            foreach (var wheelBinding in wheelBindings)
+            {
+                yield return $"Wheel {wheelIndex + 1} Button Bindings: [{string.Join(", ", wheelBinding.WheelButtons.Format())}]\nWheel {wheelIndex + 1} Clockwise Rotation: [{wheelBinding.ClockwiseRotation.Format()}]@{wheelBinding.ClockwiseActivationThreshold}°\nWheel {wheelIndex + 1} Counter-Clockwise Rotation: [{wheelBinding.CounterClockwiseRotation.Format()}]@{wheelBinding.CounterClockwiseActivationThreshold}°";
```
gonX: This line is a bit long. You should either replace the newlines with extra `yield return` statements, or you can split it into a multiline string (`"""`), or do something else that makes this easier to read.

gonX (changes requested):
Generally looks fine but one line is kinda unwieldy.

OpenTabletDriver.Console/Extensions.cs
Kuuuube: Switched this to a multiline string.

gonX (approved):
LGTM

PR #4783, opened 2026-04-30
Add support for aux keys of Artisul M0610 Pro Tablet

null-calx:
Hi, this pull request tries to add support for 7 auxilary keys present on the Artisul M0610 Pro tablet.

Points I would like to state before the PR gets merged:
- The 8th aux key, the wheel itself, and the key inside the wheel, don't send any event, which I verified on TabletDebugger and by reading from device /dev file.
- I didn't try to add support of tilt.
- I verified the 7 aux keys are getting binded correctly and actually working. But, whenever I was on TabletDebugger screen, and pressed any aux key, I would get the following error, whose root cause I couldn't figure out:
  ```
  System.IndexOutOfRangeException: Index was outside the bounds of the array.
    at OpenTabletDriver.Desktop.ViewModels.Utility.Statistic.SaveButtons(Boolean[] buttons, Int32 expectedButtons) in /home/yoimiya/git/OpenTabletDriver/OpenTabletDriver.Desktop/ViewModels/Utility/Statistics.cs:line 201
    at OpenTabletDriver.Desktop.ViewModels.TabletDebuggerViewModel.set_ReportData(DebugReportData value) in /home/yoimiya/git/OpenTabletDriver/OpenTabletDriver.Desktop/ViewModels/TabletDebuggerViewModel.cs:line 125
    at GLib.Timeout.TimeoutProxy.Handler()
  ```

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Artisul/M0610 Pro.json
+++ b/OpenTabletDriver.Configurations/Configurations/Artisul/M0610 Pro.json
@@ -32,7 +32,7 @@
       "VendorID": 9580,
       "ProductID": 111,
       "InputReportLength": 12,
-      "ReportParser": "OpenTabletDriver.Plugin.Tablet.TabletReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicV2ReportParser",
       "DeviceStrings": {
         "201": "OEM16_T205_\\d{6}$"
       },
@@ -44,7 +44,7 @@
       "VendorID": 9580,
       "ProductID": 110,
       "InputReportLength": 12,
-      "ReportParser": "OpenTabletDriver.Plugin.Tablet.TabletReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicV2ReportParser",
       "DeviceStrings": {
         "201": "OEM16_T205_\\d{6}$"
       },
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -255,7 +255,7 @@
 | Artisul A1201                      |  Missing Features | Touch bar is not yet supported.
 | Artisul AP604 (Pencil Small)       |  Missing Features | Aux buttons and eraser detection are not yet supported. Windows: Requires Zadig's WinUSB to be installed on interface 1.
 | Artisul D16 Pro                    |  Missing Features | Wheel is not yet supported.
-| Artisul M0610 Pro                  |  Missing Features | Tablet buttons, tilt, and wheel are not yet supported.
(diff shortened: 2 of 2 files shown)


Kuuuube:
Looks like you should be using `UCLogicReportParser` not creating a new parser.

null-calx:
@Kuuuube I thought of doing that, but that parser has around 20 auxiliary keys, while my tablet only has 7. Should I still go ahead and replace with that parser?

Kuuuube:
The parser supports **up to** 20 aux keys, that does not mean you need to define 20 aux keys in the config or have a tablet with 20 aux keys. Only define the amount that actually function/exist in the config and it should work as expected.

Also check if `UCLogicTiltReportParser` makes tilt work.

null-calx:
got it, will try and let you know

null-calx:
@Kuuuube hi, thanks for the suggestion, now the error in TabletDebugger has disappeared and i can see the tilt values on there too. I ended up using `UCLogicV2ReportParser` parser in the end tho.

Also, I have one more query. From lsusb's output, my tablet's VendorID = 9580, and ProductID = 111. But, the next entry in [M0610 Pro.json](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/d9c3b7deedff370bfc45d079b2387a8fd880fce4/OpenTabletDriver.Configurations/Configurations/Artisul/M0610%20Pro.json#L45) has same VendorID, and ProductID = 110. Should I also update that entry?

Kuuuube:
> Should I also update that entry?

If you mean updating the parser, sure. Should be safe enough to assume it reports the same way.

Also for verification, record some data in debugger of you pressing all the aux buttons and moving the pen around a bit. In debugger, `File > Data Recording`. Close debugger after and the recording file will be in `~/.config/OpenTabletDriver`.

null-calx:
@Kuuuube updated, and here is the file [tablet-data_1777523081.txt](https://github.com/user-attachments/files/27229926/tablet-data_1777523081.txt)

gonX (changes requested):
> @Kuuuube updated, and here is the file [tablet-data_1777523081.txt](https://github.com/user-attachments/files/27229926/tablet-data_1777523081.txt)

It looks like you missed an aux button in this recording.

null-calx:
@gonX the tablet only has 7 keys though, either i should update the button count to 7 [here](https://github.com/null-calx/OpenTabletDriver/blob/d9c3b7deedff370bfc45d079b2387a8fd880fce4/OpenTabletDriver.Configurations/Configurations/Artisul/M0610%20Pro.json#L15) which will also affect other devices, or something else. I am not sure which though, can you please point out?

null-calx:
@gonX @Kuuuube any pointers?

PR #4788, opened 2026-05-01
Add support for another way of wheel reporting from Huion Kamvas 13 (Gen 3)

Kuuuube:
Yes, we already have wheel support for this tablet. Yes, this is detected as the exact same tablet we already have support for. Yes, it reports the wheel differently while not even having any difference in identifiers.

...

Wheel data alone: [tablet-data_1777616712.txt](https://github.com/user-attachments/files/27283082/tablet-data_1777616712.txt)
Moving the pen around: [tablet-data_1777617525.txt](https://github.com/user-attachments/files/27283087/tablet-data_1777617525.txt)
First wheel moving while pressing aux buttons (note that wheel and aux are seen in the same reports): [tablet-data_1777617562.txt](https://github.com/user-attachments/files/27283092/tablet-data_1777617562.txt)
Same as above but the second wheel: [tablet-data_1777617618.txt](https://github.com/user-attachments/files/27283099/tablet-data_1777617618.txt)

Verification: https://discord.com/channels/615607687467761684/723399565780189234/1500386953202372699

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Huion/GianoReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Huion/GianoReportParser.cs
@@ -11,6 +11,8 @@ namespace OpenTabletDriver.Configurations.Parsers.Huion
         {
             if (data[1] == 0xF1)
                 return new KamvasRelWheelReport(data);
+            if (data[1] == 0xF0)
+                return new UCLogicAuxRelWheelReport(data);
             if (data[1].IsBitSet(5) && data[1].IsBitSet(6))
                 return new UCLogicAuxReport(data);
             else
--- a/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicAuxRelWheelReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicAuxRelWheelReport.cs
@@ -1,0 +1,35 @@
+using OpenTabletDriver.Plugin.Tablet;
+using OpenTabletDriver.Plugin.Tablet.Wheel;
+
+namespace OpenTabletDriver.Configurations.Parsers.UCLogic
+{
+    public struct UCLogicAuxRelWheelReport : IAuxReport, IRelativeWheelReport
+    {
+        public UCLogicAuxRelWheelReport(byte[] report)
+        {
+            Raw = report;
+
+            AuxButtons =
+            [
+                report[4].IsBitSet(0),
+                report[4].IsBitSet(1),
+                report[4].IsBitSet(2),
+                report[4].IsBitSet(3),
+                report[4].IsBitSet(4),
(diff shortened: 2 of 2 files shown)


gonX (commented):
This seems potentially problematic for the 47 other tablets using the same parser. `0xF0` matches the `if` clause below (bit `5` and `6` is on) and the new code path cannot support all the aux buttons.

Kuuuube:
It should be E0 or E1 not F0 usually iirc.

Kuuuube:
Additional verification on this for this tablet: https://discord.com/channels/615607687467761684/723399565780189234/1501790005649997824

```
normal:
09 E0 00 00 00 00 00

wheel 1 cw:
09 F0 00 00 00 01 00
wheel 1 ccw:
09 F0 00 00 00 FF 00

wheel 2 cw:
09 F0 00 00 00 00 01
wheel 2 ccw:
09 F0 00 00 00 00 FF
```

gonX (changes requested):
I would prefer to see an auxiliary button recording of one of the following tablets using the same parser that has more than 8 buttons:

- Gaomon GM156HD
- Gaomon PD1560
- Gaomon PD1561
- Gaomon PD156 Pro
- Huion GT-156HD V2
- Huion GT-221
- Huion GT-221 Pro
- Huion Kamvas 16 (2021)
- Huion Kamvas 16
- Huion Kamvas Pro 20
- Huion Kamvas Pro 22 (2019)
- Huion Kamvas Pro 24
- Huion RDS-160
- Huion WH1409
- Huion WH1409 V2

> List generated using `sh`+`jq` one-liner:
> 
> ```
> for file in **/*.json; do jq -r 'select(.Specifications.AuxiliaryButtons.ButtonCount != null and .DigitizerIdentifiers[].ReportParser == "OpenTabletDriver.Configurations.Parsers.Huion.GianoReportParser") | { "Name": .Name, "AuxButtons": .Specifications.AuxiliaryButtons.ButtonCount } | select(.AuxButtons > 8) | "- " + .Name ' "$file"; done
> ```
> (you have to be in the `OTD.Configurations/Configurations` subdir):

It does not need to be a tablet recording from this branch, but it would be preferable.

Kuuuube:
#4837 also supporting this type of data being reported but on the kamvas 16 gen 3. Will need to pull this parsing into the `KamvasOffsetReportParser` from #4840 as well.

Kuuuube:
> I would prefer to see an auxiliary button recording of one of the following tablets using the same parser that has more than 8 buttons

Gave the initial support verification for each of these a check and not a single one has a recording of the buttons. They always work without even needing to do anything lol.

There are however a few other tablets with less buttons that match the E0 behavior for aux as expected. Can't be 100% certain but this is probably fine.

gonX (approved):
Based on the above comments I'm going to approve this.

PR #4789, opened 2026-05-02
Require Deco 01 V2 last identifier regex match to be from start

Kuuuube:
Back at it again with the Deco 01 V2... There was a bug in my testing script for this. Some variants of the Deco 01 V2 (Variant 2) could get double detected.

Python has created a glorious footgun in their regex handler by making `re.match` ONLY match at the start of a string (effectively forcing a `^` at the start of any regex). Whereas `re.search` is the one you actually want to use if you want your regex to work like you wrote it. Only a true genius could comprehend the reasoning for making a whole extra method just to avoid typing one extra character in a regex string. (They also have `re.fullmatch` for when you're two lazy to type both `^` and `$` in your string...)

Anyways, this one regex in the config ended up missing a `^` at the start due to this.


gonX (approved):
Assuming this works but we really need to start including original device strings in tests 🙈

PR #4791, opened 2026-05-02
Remove variation selector 16 from down arrow

Kuuuube:
Windows doesn't have the fonts to render it or wpf is doing something weird... it renders as a box. And we arent even using it.

image


gonX (approved):
Very confusing diff but I assume it works

PR #4792, opened 2026-05-02
Use symbols from the Miscellaneous Symbols and Arrows block instead of the Arrows and Dingbats blocks

Kuuuube:
It looks alright on Linux as-is but on Windows it does not (ignore the boxes, those would be fixed with #4791).
image

Windows:
image

Linux:
image


gonX:
I don't like the updown arrows. I don't think they're clear enough for the end user that this means that the button is "known good"

A different way of expressing success (like a tri-state checkbox) could be considered instead.

Kuuuube:
Closing as #4794 took care of the bad looking checkmark.

PR #4793, opened 2026-05-02
Fix windows tablet debugger additional statistics from failing due to STAThread error

Kuuuube:
The error:
image

"STAThread refers to 'Single-Threaded Apartments' which refers to the threading model used by the current (main) thread."
"The single-threaded application model requires that no single object 'live in' more than one STA thread at a time"


gonX (approved):
LGTM. Thanks for finding the fix.

PR #4807, opened 2026-05-05
Gaomon M6: Add wheel support

StaySound4:
# Changes

- Add wheel support for the Gaomon M6.
- Add a UCLogic V2 wheel parser for `0xf0` reports where the wheel position is reported in `data[5]`.
- Configure the M6 as a 12-step absolute wheel without changing the shared `UCLogicV2ReportParser` used by other tablets.
- Update `TABLETS.md` so the M6 only lists the touch bar as unsupported.
- Add parser and binding tests covering wheel report parsing, idle wheel reports, clockwise rotation, and counter-clockwise rotation.

This intentionally only covers wheel rotation. Touch bar support and VMulti/Windows Ink mouse-scroll compatibility are outside the scope of this PR.

Validation:
- `git diff --check origin/0.6.x..HEAD`
- Static M6 config assertions: one wheel, `AbsoluteWheelMax = 11`, `ButtonCount = 0`, and both M6 digitizer identifiers use `UCLogicV2WheelReportParser`.
- `dotnet test` could not be run locally because this machine has no .NET SDK installed and dotnet reports the upstream `global.json` SDK version as invalid in this environment.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/M6.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/M6.json
@@ -13,14 +13,20 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 13
-    }
+    },
+    "Wheels": [
+      {
+        "AbsoluteWheelMax": 11,
+        "ButtonCount": 0
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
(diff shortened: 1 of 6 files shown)


StaySound4:
I tested this on real Gaomon M6 hardware. The touch wheel now works.

PR #4817, opened 2026-05-09
macOS: Add "About" to tray menu

AkiSakurai:
Since the app menu is no longer shown after adding LSUIElement, we’ve added the "About" item back to the tray menu so it remains accessible.

### DIFF
--- a/OpenTabletDriver.UX/TrayIcon.cs
+++ b/OpenTabletDriver.UX/TrayIcon.cs
@@ -88,6 +88,12 @@ namespace OpenTabletDriver.UX
             };
             close.Click += (sender, e) => window.Close();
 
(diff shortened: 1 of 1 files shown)


gonX (approved):
Makes sense to me. I don't have a tray icon on Linux, so a Windows users will need to test also.

Kuuuube:
Doesn't work on windows. The about window "opens" but it is invisible. Might need to also show the full gui window first to get it to display properly.

Could also limit this to macos only but I'm not quite sure what this is solving. Isn't it a much bigger issue that the app menu isn't available than just the about window being inaccessible? What about all the other things on the menu like tablet debugger, plugin manager, or anything else?

PR #4819, opened 2026-05-10
macOS: restore menu bar and bring window to foreground

AkiSakurai:
Use NSApplicationActivationPolicy to dynamically hide the Dock icon when the MainForm loses focus or is minimized. Call ActivateIgnoringOtherApps to properly bring the application to the foreground when it gains focus or is de-minimized.

Supersedes #4816 #4817


Kuuuube (approved):
Looks good. Works.

PR #4821, opened 2026-05-10
Add XP-Pen Deco Pro M wheel and touchpad support

DarkFoxV:
Adds support for the XP-Pen Deco Pro M wheel and touch wheel input mappings.

The tablet exposes two wheel-like inputs:

* an outer physical wheel using the existing mapping
* an inner touch wheel using different bits (`0x04` / `0x08`) 

This change adds support for the Deco Pro M touch wheel bit mapping while preserving compatibility with existing devices.

## Tested on

* XP-Pen Deco Pro Medium

## Notes

The inner control is technically a touch wheel/touchpad, but it behaves as a wheel-like rotational input in practice.

## Input mapping captures
wheel1_counterclockwise
wheel1
wheel2_counterclockwise
wheel2

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenAuxReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenAuxReport.cs
@@ -33,12 +33,28 @@ namespace OpenTabletDriver.Configurations.Parsers.XP_Pen
                 report[auxIndex + 2].IsBitSet(3),
             ];
 
-            // 0x01 for 1st wheel clockwise, 0x02 for counterclockwise, verified on XP Pen Artist 13.3 Pro V2
-            // 0x10 for 2nd wheel clockwise, 0x20 for counterclockwise, verified on XP Pen Artist 22R Pro
+            // 0x01 for 1st wheel clockwise, 0x02 for counterclockwise,
+            // verified on XP-Pen Artist 13.3 Pro V2 and Deco Pro M
+
+            // 0x10 for 2nd wheel clockwise, 0x20 for counterclockwise,
+            // verified on XP-Pen Artist 22R Pro
+
+            // 0x04 for touchpad clockwise, 0x08 for counterclockwise,
+            // verified on XP-Pen Deco Pro M
+
             AnalogDeltas =
             [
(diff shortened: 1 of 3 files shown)


TABLETS.md
gonX, line 1: Not sure why this is changing so much but you also need to ensure it's ordered by Status column first (in its existing order) and _then_ name. This should diff as removing and adding 1 line (it currently removes and adds 345 lines)

OpenTabletDriver.Configurations/Parsers/XP_Pen/XP_PenAuxReport.cs
```diff
@@ -33,12 +33,28 @@ public XP_PenAuxReport(byte[] report, int auxIndex = 2, int wheelIndex = 7)
...
+                    ? 1
+                    : report[wheelIndex].IsBitSet(1)
+                        ? -1
+                        : 0,
+
+                report[wheelIndex].IsBitSet(4) || report[wheelIndex].IsBitSet(2)
+                    ? 1
+                    : report[wheelIndex].IsBitSet(5) || report[wheelIndex].IsBitSet(3)
+                        ? -1
+                        : 0,
```
gonX, line 57: I don't love how this reads

Also this may not work as well as we'd like since it looks like this is using a 3rd set of bits compared to just the first 2. What happens if a later tablet that could use this parser reports its 2 wheels on the new set of bits you're checking?

I feel like the wheel specifications needs to change a bit to know which index to reach each wheel for, so that the parser can support reporting all 3 wheels independently, but tablets using this parser would then (somehow) define in their wheel specifications which index in the `AnalogDeltas`/absolute-wheel-equivalent that the wheel is associated with. While you are welcome to try implementing this yourself in a new PR, it feels like something that a team member would need to do to ensure that not too much is breaking.

@Kuuuube thoughts? Maybe I'm overengineering this and this just needs to be moved into its own parser.

gonX (dismissed):
While this looks like it'll work fine for anyone needing this, there are some technicalities I don't like. Also merge conflicts would need to be resolved

TABLETS.md
DarkFoxV: Fixed. I updated TABLETS.md and kept the existing Status order, then sorted by name. The diff is now only removing and adding 1 line instead of changing the whole file.

PR #4824, opened 2026-05-13
Fixing Veikk A15 V2 Tilting | Implement ITiltReport in VeikkTabletReport struct

huybin1234b:


Kuuuube:
?

https://github.com/OpenTabletDriver/OpenTabletDriver/blob/0.6.x/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltTabletReport.cs

PR #4823, opened 2026-05-11
Fix Gaomon M10K Pro: Add Touch Ring support and unbind auxiliary buttons

itzikgebner:
This PR fixes the Touch Ring and the center button behavior for the Gaomon M10K Pro.

Technical details:
1. Switched ReportParser from UCLogicTiltReportParser to HuionTiltReportParser. The previous parser was incorrectly interpreting the absolute wheel coordinates (0x01 - 0x0D) as auxiliary button presses.
2. Added Wheels array with AbsoluteWheelMax: 13. This enables smooth, directional scrolling/brush resizing.
3. Increased ButtonCount to 11. The physical center button sends "04" in the second auxiliary byte (Button 11). Previously, it was mirroring Button 3 due to the 10-button limit.

Tested on Gentoo Linux with OpenTabletDriver v0.6.7. Everything works as intended.

PR #4825, opened 2026-05-13
UCLogic wheel support for Gaomon M10K / M10K Pro, artist mode scroll fix

### LINKED ISSUE #5024: Gaomon M10K Pro touch wheel support
### Description

# **Product Name**
Gaomon M10K Pro

# **Feature Request**
Gaomon M10K Pro has a touch / scroll wheel ring that can be used for actions such as zooming, changing brush size and scrolling etc...

The tablet itself is already detected and works perfectly, but I can't configure or even use the touch wheel, it just do the copy / paste actions on Firealpaca

I would like to request support for the M10K Pro's touch wheel so that its input can be exposed as a configurable wheel/scroll input and mapped to the wanted actions

### Acknowledgements

BarrettLowe:
## What this does

The Gaomon M10K and M10K Pro have a touch ring. The 0xf0 report byte carrying wheel data was being parsed and thrown away. This wires it up, and also fixes scroll bindings doing nothing in artist mode on Linux.

### UCLogic wheel support

New `UCLogicWheelReport` struct handles 0xf0 reports across all three UCLogic parsers. Position comes from data[5]; 0 means not touching, otherwise decremented to zero-index it. The M10K and M10K Pro configs now declare the wheel (AbsoluteWheelMax: 11, AngleOfZeroReading: 180), and both are marked Supported in TABLETS.md.

### Artist mode scroll fix

In Linux artist mode the output pointer is a uinput tablet device — it doesn't implement IMouseScrollHandler, so scroll bindings were registered against nothing. The daemon now falls back to the relative pointer (virtual mouse) for scroll when the output mode's pointer doesn't provide it.

## Testing

Tested on a Gaomon M10K Pro. Touch ring works in artist mode, and scroll bindings work in both absolute and artist output modes.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K Pro.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K Pro.json
@@ -13,7 +13,14 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 10
-    }
+    },
+    "Wheels": [
+      {
+        "AbsoluteWheelMax": 11,
+        "AngleOfZeroReading": 180.0,
+        "ButtonCount": 0
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K Pro.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K Pro.json
@@ -12,11 +12,11 @@
       "ButtonCount": 2
     },
     "AuxiliaryButtons": {
-      "ButtonCount": 10
+      "ButtonCount": 11
     },
     "Wheels": [
       {
-        "AbsoluteWheelMax": 11,
+        "AbsoluteWheelMax": 12,
         "AngleOfZeroReading": 180.0,
         "ButtonCount": 0
       }
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/M10K.json
@@ -13,7 +13,14 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 11
-    }
+    },
+    "Wheels": [
+      {
+        "AbsoluteWheelMax": 11,
+        "AngleOfZeroReading": 180.0,
+        "ButtonCount": 0
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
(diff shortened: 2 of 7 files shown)


Kuuuube:
Record some tablet data for verification. The artist mode change should be split into a separate PR.

File > record in debugger then close debugger and the file will be in otd's appdata folder. On linux this is most likely `~/.config/OpenTabletDriver`.

BarrettLowe:
Flagging some context for reviewers before this gets looked at alongside #4823.

I only have a Gaomon M10K (2018), so the M10K changes here are hardware-verified. The M10K Pro config is a best-guess port on the assumption the hardware behaves the same way — I can't confirm the button count, wheel range, or zero angle on that device.

On the M10K, `data[5]` reported `0x00` when the ring was untouched, `0x01` at the 6 o'clock position, and incremented clockwise up to `0x0C` — 12 discrete positions. Subtracting 1 to zero-index gives a range of 0–11, hence `AbsoluteWheelMax: 11` and `AngleOfZeroReading: 180.0`.

#4823 covers the M10K Pro only and uses `AbsoluteWheelMax: 13` and `HuionTiltReportParser` instead of the UCLogic parser. If whoever wrote that has a Pro in hand, they likely have better data for that device than I do. Happy to drop or split out the M10K Pro config here if that's cleaner.

BarrettLowe:
Tablet recording from Gaomon M10K (2018): https://gist.github.com/BarrettLowe/cb27d48afd700a79b0d327ca1262d5dd

Covers full wheel rotation (positions 0–11), idle state, pen input, and all aux buttons. The 0xf0 reports are correctly parsed as `UCLogicWheelReport` with `data[5]` carrying the position.

Also split the daemon scroll fix out to #4826 as requested.

itzikgebner:
Hey @BarrettLowe! Thanks for the explanation about the zero-indexing — that makes perfect sense! `AbsoluteWheelMax: 12` is absolutely correct for the **Pro** given the `0x0D` raw max.

I pulled your branch (#4825) and compiled it locally on my Linux setup with the actual **M10K Pro** connected. I can confirm:
1. The wheel works perfectly and smoothly with the native _UCLogic_ parser. No phantom button presses at all.
2. The center button is correctly exposed and fully functional as _Auxiliary Button 11_.

Awesome work on fixing the `0xf0` routing natively! I will go ahead and close my PR (#4823) in favor of yours. Thanks for collaborating!

BarrettLowe:
> Tablet recording from Gaomon M10K (2018): https://gist.github.com/BarrettLowe/cb27d48afd700a79b0d327ca1262d5dd
> 
> Covers full wheel rotation (positions 0–11), idle state, pen input, and all aux buttons. The 0xf0 reports are correctly parsed as `UCLogicWheelReport` with `data[5]` carrying the position.
> 
> Also split the daemon scroll fix out to #4826 as requested.

Hey @Kuuuube does this show what you were wanting?

BarrettLowe:
@Kuuuube ping on this when you get a chance. Thanks!

PR #4826, opened 2026-05-13
Daemon: Fix scroll bindings in artist mode on Linux

BarrettLowe:
In artist mode, the output pointer is a virtual tablet device that doesn't implement IMouseScrollHandler, so scroll bindings were silently dropped.

Now falls back to DesktopInterop.RelativePointer for scroll when the output mode's pointer doesn't provide it.


gonX:
Similar PR that predates this and had almost exactly the same logic before I rebased it just now: #4804 

Adding this feature will be tracked there as it was already actively a work in progress. Thanks for your PR though!

PR #4779, opened 2026-04-29
Enable Nullable Reference Types (NRT)

gonX:
Please put extra scrutiny on null-silencing operators (`!`). The only allowed ones should be:
- Anything involving UI bindings as null values relating to bindings are seemingly properly handled by Eto
- Anywhere where the only sensible replacement is adding a null-coalescing operator to throw a different type of exception
  - As in, the code is shorter this way but reads less clean. This can of course be discussed to have more explicit handling in the code (as sometimes a custom exception may clearer indicate what went wrong)

Additionally, some macOS desktop/native operations have not been marked as nullable yet as it is impossible for me to test.

Cherry-picked commit 055a869 from #4563 and ironed out the remaining warnings. It depended on some other changes that took care of some other nullability considerations which I've tried reintroducing here in the following commits.

PR #4831, opened 2026-05-17
Remove intentional throw from unknown path on `outputModeSelector.SelectedItemBinding.Convert`

Kuuuube:
It's okay if `PluginSettingStore` is null here.

I would like to do `c => c?.OutputMode` but that's not allowed. `An expression tree lambda may not contain a null propagating operator`

Regression from #4779

### DIFF
--- a/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
+++ b/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
@@ -70,7 +70,7 @@ namespace OpenTabletDriver.Desktop.Reflection
             return obj;
         }
 
-        public static PluginSettingStore? FromPath(string path)
+        public static PluginSettingStore? FromPath(string? path)
         {
             var pathType = AppInfo.PluginManager.PluginTypes.FirstOrDefault(t => t.FullName == path);
             return pathType != null ? new PluginSettingStore(pathType) : null;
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
```diff
@@ -70,7 +70,7 @@ public PluginSettingStore(string path, ObservableCollection<PluginSetting> setti
             return obj;
         }
 
-        public static PluginSettingStore? FromPath(string path)
+        public static PluginSettingStore? FromPath(string? path)
         {
```
gonX, line 74: I think it would be preferable to have this short-circuit when `path` is null:

```suggestion
        {
            if (path == null) return null;

```

`Type.FullName` can theoretically be null in some edge cases, and I think there could be a surprise vector here without a short-circuit.

I'll add a follow-up PR since this PR as-is still keeps the exact same behavior as before, so there is no reason for me to request changes here.

gonX (approved):
Good change - I can't remember the original reasoning behind me adding this intentional throw but now there's at least a `git blame` record for needing this.

gonX:
Also, for context to readers in the future (since the PR doesn't fully explain it), there as some rare cases, especially (only?) on Windows, where this is unexpectedly being observed as being null, and massaging it as this PR does fix the issue seamlessly.

PR #4834, opened 2026-05-21
Display a Notification when daemon & UX version mismatch

Mrcubix:
Would somewhat limit the amount of people coming in the discord because a required function is missing on the still running daemon after an update.

### DIFF
--- a/OpenTabletDriver.Desktop/AppInfo.cs
+++ b/OpenTabletDriver.Desktop/AppInfo.cs
@@ -83,7 +83,7 @@ namespace OpenTabletDriver.Desktop
             }
             get;
         }
-        
+
         public string? CommandLineConfigurationDirectory
         {
             set
--- a/OpenTabletDriver.Desktop/AppInfo.cs
+++ b/OpenTabletDriver.Desktop/AppInfo.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Reflection;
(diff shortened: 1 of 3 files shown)


gonX (commented):
I'll have to test this manually but code looks good

Kuuuube (changes requested):
`Version` should probably be a string.

While I do like the idea of checking if the version is higher or lower to give troubleshooting advice, versions can have some arbitrary stuff in there. Especially when packaging for Linux or using dev builds.

For example, `0.6.7+78-ga39bea21`. Rather than a clean `0.6.7`. If this happens to be the case it entirely breaks.

PR to your PR here https://github.com/Mrcubix/OpenTabletDriver-Except-I-Lied/pull/2 for what this could look like. Using `String.Compare` here which is kinda jank but beyond implementing this multiple ways or having fallbacks, this is fairly clean. Versions like `6.10.0` vs `6.9.0` mess up `String.Compare` and it will report `6.10.0` as lower.

Wouldn't be against it if you do want to implement something to parse the version better either. Could do something like splitting by `.` and try int parsing until it throws then using that number. Or chopping off the version at the first non number or dot then throwing that into a `Version` to be handled the same as you have now.

Could also combine the `specific` tips so it tells the user both instead of checking if the version is higher or lower.

Mrcubix:
I do not like the way it was done.
The commit should be separate to the version in the first place.
Version checks should then still work as expected.
We can then check if commits matches, which then wouldn't and we can provide feedback on that.

Kuuuube (approved):
Looks good.

PR #4836, opened 2026-05-23
Allow configs to match `HID_REPORTS` device attribute

Kuuuube:
Got a case where this will be necessary. Might update this PR to include dependant config later.

Example of what this looks like:
```json
"DigitizerIdentifiers": [
  {
    "VendorID": 1386,
    "ProductID": 1013,
    "InputReportLength": 192,
    "ReportParser": "OpenTabletDriver.Configurations.Parsers.Wacom.IntuosV3.IntuosV3ReportParser",
    "FeatureInitReport": [
      "AgI="
    ],
    "Attributes": {
      "HidReports": "FE:0001:0080, 06:000D:0001, 1E:FF0D:0001, 11:FF0D:0001, 13:FF0D:0001, AC:FF0D:0001"
    }
  }
]
```

I don't totally love that this is string matching into something that should probably be a collection. But `HID_REPORTS` does appear to always be consistently ordered. And device attributes currently can only be a string. Could also split it apart for this check but that also seems a little jank. Can deal with this being a problem when it arises.

### DIFF
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -228,7 +228,8 @@ namespace OpenTabletDriver
             }
 
             if (attributes.TryGetValue("HidReports", out var hidReports)
-                && (!device.DeviceAttributes.TryGetValue("HID_REPORTS", out var usbHidReports) || !Regex.IsMatch(usbHidReports, hidReports)))
+                && device.DeviceAttributes.TryGetValue("HID_REPORTS", out var usbHidReports)
+                && !Regex.IsMatch(usbHidReports, hidReports))
             {
                 return false; // HidReports specified and no HID Reports match.
             }
--- a/OpenTabletDriver/Driver.cs
+++ b/OpenTabletDriver/Driver.cs
@@ -227,6 +227,12 @@ namespace OpenTabletDriver
                 }
             }
 
+            if (attributes.TryGetValue("HidReports", out var hidReports)
+                && (!device.DeviceAttributes.TryGetValue("HID_REPORTS", out var usbHidReports) || !Regex.IsMatch(usbHidReports, hidReports)))
+            {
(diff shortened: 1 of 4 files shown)


OpenTabletDriver/Driver.cs
```diff
@@ -227,6 +227,12 @@ private static bool DeviceMatchesAttribute(IDeviceEndpoint device, Dictionary<st
                 }
             }
 
+            if (attributes.TryGetValue("HidReports", out var hidReports)
+                && (!device.DeviceAttributes.TryGetValue("HID_REPORTS", out var usbHidReports) || !Regex.IsMatch(usbHidReports, hidReports)))
```
gonX: To match the styling above:
```suggestion
                && attributes.TryGetValue("HID_REPORTS", out var usbHidReports)
                && !Regex.IsMatch(usbHidReports, hidReports))
```

Using `attributes` also allows a global `HidReports` to be specified in the tablet config, as they're merged earlier in this function.

gonX (changes requested):
Good start to this idea, can't really think of anything this is gonna really miss. We can still match partially since we're using a regex.

Only a small styling and logic change needed to make it on par with the other code in the function.

OpenTabletDriver/Driver.cs
Kuuuube: The `device.DeviceAttributes.TryGetValue("HID_REPORTS", out var usbHidReports)` is only for the device endpoint itself. `attributes.TryGetValue("HidReports", out var hidReports)` is already used for fetching the config attribute value.

Fixed the styling.

Kuuuube:
Added config tests for this value.

OpenTabletDriver/Driver.cs
gonX: Whoops. Thanks.

gonX (approved):
LGTM

PR #4841, opened 2026-05-24
Skip known dangerous strings in string dump by default

Kuuuube:
Sometimes users run a string dump on their tablets without knowing they probably shouldn't be randomly doing that. This should allow for minimizing the risk.


gonX (approved):
LGTM. I think string requests (specific index) should also warn users, but that can come later.

Kuuuube:
Probably yeah. But the chance a user explicitly requests a dangerous string is pretty low.

PR #4842, opened 2026-05-24
Fix reversed App.Driver.IsConnected check in string reader

Kuuuube:
Regression caused by #4779. Completely bricks the string reader.


gonX (approved):
LGTM, good catch.

PR #4843, opened 2026-05-25
Fix full area resize breaking aspect ratio lock

Kuuuube:
Regression caused by #4435

Same deal as the logic above it running under `if (sender == tabletWidth || sender == tabletAreaEditor)` except flipped for height.

Without this when using the full area resize the height aspect ratio locker for may (if the ratio is >1) try to keep the height full and expand the width out of bounds. Then the area constrainer runs and sees that the width is out of bounds and snaps it into bounds. This leaves the user with the tablet's area set to the entire full area but ignoring their aspect ratio setting.

Why this wasn't broken before (v0.6.6.2), I'm not sure and I haven't looked into.

Also not entirely sure why this logic is locked behind `tabletAreaEditor.FullAreaCommandExecuting` since this allows the user to type in values manually that break the aspect ratio. But this matches the previous behavior.

### DIFF
--- a/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
+++ b/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
@@ -191,7 +191,17 @@ namespace OpenTabletDriver.UX.Controls.Output
                 }
                 else if (sender == tabletHeight)
                 {
-                    tabletWidth.DataValue = displayWidth.DataValue / displayHeight.DataValue * tabletHeight.DataValue;
+                    var fullWidth = tabletAreaEditor.FullAreaBounds!.Value.Width;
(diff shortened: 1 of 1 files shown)


gonX:
I think I prefer the approach in #4844 as it'd be less code to maintain. This might work now, but if we ever change the logic later, there's a chance we'll have a new weird edge case again.

Kuuuube:
Part of the reason I put together this change is that we already do this exact same thing for width. If you look a bit up in the code at `if (sender == tabletWidth || sender == tabletAreaEditor)`.

I assume #4844 is fine (havent tested) but it feels like we're adding a race condition to fix a race condition.

gonX:
Might be worth taking a look at #4845 as it's a more encompassing fix to the core issue and extends the aforementioned PR

Issue #2766, opened 2023-05-30
Add support for HS611 w/ latest FW


adam-kulagowski:
### Manufacturer

Huion

### Product Name

HS611

### Device Width

258.4mm

### Device Height

161.5mm

### Pressure Levels

8192

### Lines per Inch (LPI)

5080

### Additional Information

Latest FW on HS611 changes USB product ID of upgraded tablet. This makes the device unusable in Linux unless one want to use closed source Huion linux drivers.  New USB ID is:

`256c:006f HUION Huion Tablet_HS611`

Attached new configuration, placed in `/usr/share/OpenTabletDriver/Configurations/HS611-fw.json` makes the tablet fully visible and operational again for OpenTabletDriver:

[HS611-new-fw.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11597282/HS611-new-fw.txt)

I've created this configuration based on original HS611 config from github. The only things I've changed are:
`ProductID` and `DeviceStrings/201`

I'm attaching also device strings dump:
[hs611-dump.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11597315/hs611-dump.txt)

### Diagnostic Information

[hs611-diag.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11597321/hs611-diag.txt)

### Acknowledgements

- [X] I have searched the existing issues and this new issue is not a duplicate of any.
- [X] I have written a concise and meaningful title.

adam-kulagowski:
I'm attaching updated configuration with tilt working. 

[HS611-fw-tilt.txt](https://github.com/OpenTabletDriver/OpenTabletDriver/files/11598827/HS611-fw-tilt.txt)

X9VoiD:
Resolved by #2801

PR #4867, opened 2026-06-03
Lock Huion HS611 interface to 0

Kuuuube:
Diag seen here detecting on interface 1, user reporting tablet does not work: [Diagnostics-067 Huion HS611.json](https://github.com/user-attachments/files/28561872/Diagnostics-067.Huion.HS611.json)
Verification: https://discord.com/channels/615607687467761684/723399565780189234/1511776476125728869

Prior issue on `111` pid variant (#2766) shows both of these conflicting identifiers too but interface `0` shows up first so it detects correctly.

I can't find any info about the very start of this config. It think it was pulled in from hawku driver with no verification on our end. It probably works fine on windows without this change but needs it on linux.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/HS611.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/HS611.json
@@ -42,6 +42,7 @@
     }
   ],
   "Attributes": {
-    "libinputoverride": "1"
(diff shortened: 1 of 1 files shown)


gonX (approved):
LGTM, verification looks good.

PR #4896, opened 2026-06-23
Fixed runtime error when clicking ... for tip binding

Davidj361:
Fixes #4895

PR #4897, opened 2026-06-23
Allow PluginSettingStore Path to be null and handle it as such

### LINKED ISSUE #4895: Clicking `...` for Tip Binding Gives Error
### Steps to Reproduce

Clicking this red box button give this runtime error popup

Image

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at OpenTabletDriver.Desktop.Reflection.PluginSettingStore..ctor(Type type, Boolean enable) in F:\code\OpenTabletDriver\OpenTabletDriver.Desktop\Reflection\PluginSettingStore.cs:line 18
   at OpenTabletDriver.UX.Windows.Bindings.AdvancedBindingEditorDialog.<>c.<.ctor>b__0_0(TypeInfo t) in F:\code\OpenTabletDriver\OpenTabletDriver.UX\Windows\Bindings\AdvancedBindingEditorDialog.cs:line 73
   at Eto.Forms.BindableBinding`2.<>c__DisplayClass7_0`1.<Convert>b__0(T c)
   at Eto.Forms.ObjectBinding`2.get_DataValue()
   at Eto.Forms.ObjectBinding`2.set_DataValue(TValue value)
   at Eto.Forms.DualBinding`1.SetDestination()
   at Eto.Forms.DualBinding`1.SetBinding(Boolean setInitialValue)
   at Eto.Forms.DualBinding`1..ctor(DirectBinding`1 source, DirectBinding`1 destination, DualBindingMode mode)
   at Eto.Forms.ObjectBinding`2.Bind(DirectBinding`1 sourceBinding, DualBindingMode mode)
   at Eto.Forms.BindableBinding`2.Bind(DirectBinding`1 sourceBinding, DualBindingMode mode)
   at OpenTabletDriver.UX.Windows.Bindings.AdvancedBindingEditorDialog..ctor(PluginSettingStore currentBinding) in F:\code\OpenTabletDriver\OpenTabletDriver.UX\Windows\Bindings\AdvancedBindingEditorDialog.cs:line 73
   at OpenTabletDriver.UX.Controls.BindingDisplay.<.ctor>b__0_2(Object sender, EventArgs e) in F:\code\OpenTabletDriver\OpenTabletDriver.UX\Controls\BindingDisplay.cs:line 46
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__124_0(Object state)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
```

### Expected Behavior

No error

### Observed Behavior

No runtime error, popup for what bindings for pen tip.

### Tablet Device

Huion Kamvas Pro 16

### Diagnostic Information

[Diagnostics-067+83-g6c330431-dirty Huion Kamvas Pro 16.json](https://github.com/user-attachments/files/29233029/Diagnostics-067%2B83-g6c330431-dirty.Huion.Kamvas.Pro.16.json)

### Acknowledgements

- [x] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [x] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.

Kuuuube:
Issue that snuck in with NRT. Restoring this to the original behavior instead of throwing.

Added `if (Path == null) return null;` is new and likely not possible to hit this case but silencing the warning.

Fixes #4895

### DIFF
--- a/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
+++ b/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
@@ -3,6 +3,7 @@ using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
 using Newtonsoft.Json;
+using OpenTabletDriver.Plugin;
 using OpenTabletDriver.Plugin.Attributes;
 using OpenTabletDriver.Plugin.DependencyInjection;
 using OpenTabletDriver.Plugin.Tablet;
@@ -48,7 +49,11 @@ namespace OpenTabletDriver.Desktop.Reflection
 
         public T? Construct<T>(TabletReference? tabletReference = null, bool trigger = true) where T : class
         {
-            if (Path == null) return null;
+            if (Path == null)
+            {
+                Log.Write($"Construct<T>", $"{nameof(Path)} is null, returning null", LogLevel.Debug);
+                return null;
+            }
 
             var obj = AppInfo.PluginManager.ConstructObject<T>(Path);
             ApplySettings(obj);
--- a/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
+++ b/OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
@@ -15,8 +15,8 @@ namespace OpenTabletDriver.Desktop.Reflection
 
         public PluginSettingStore(Type type, bool enable = true)
         {
-            Path = type.FullName ?? throw new InvalidOperationException($"Could not look up full name for type {type}");
-            Settings = GetSettingsForType(type);
+            Path = type?.FullName;
+            Settings = type != null ? GetSettingsForType(type) : new ObservableCollection<PluginSetting>();
             Enable = enable;
         }
 
@@ -24,8 +24,7 @@ namespace OpenTabletDriver.Desktop.Reflection
         {
             var sourceType = source.GetType();
 
-            Path = sourceType.FullName ??
-                   throw new InvalidOperationException($"Could not look up {nameof(Path)}'s full name via type '{sourceType}'");
+            Path = sourceType.FullName;
 
             Settings = GetSettingsForType(sourceType, source);
             Enable = enable;
@@ -38,10 +37,10 @@ namespace OpenTabletDriver.Desktop.Reflection
             Settings = settings;
         }
 
-        public string Path { set; get; }
+        public string? Path { set; get; }
 
         [JsonIgnore]
-        public string? Name => AppInfo.PluginManager.GetFriendlyName(Path);
+        public string? Name => Path != null ? AppInfo.PluginManager.GetFriendlyName(Path) : null;
 
         public ObservableCollection<PluginSetting> Settings { set; get; }
 
@@ -49,6 +48,8 @@ namespace OpenTabletDriver.Desktop.Reflection
 
         public T? Construct<T>(TabletReference? tabletReference = null, bool trigger = true) where T : class
         {
+            if (Path == null) return null;
+
             var obj = AppInfo.PluginManager.ConstructObject<T>(Path);
             ApplySettings(obj);
             if (trigger)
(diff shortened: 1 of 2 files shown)


OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
```diff
@@ -44,17 +43,19 @@ public PluginSettingStore(string path, ObservableCollection<PluginSetting> setti
...
+        public string? Name => Path != null ? AppInfo.PluginManager.GetFriendlyName(Path) : null;
 
         public ObservableCollection<PluginSetting> Settings { set; get; }
 
         public bool Enable { set; get; }
 
         public T? Construct<T>(TabletReference? tabletReference = null, bool trigger = true) where T : class
         {
+            if (Path == null) return null;
+
```
gonX: I don't like these silent returns while building the driver pipeline. It makes diagnosis of potential issues more difficult. We should either log it or throw an exception:

```suggestion
            if (Path == null)
            {
                Log.Write($"Construct<T>", $"{nameof(Path)} is null, silently returning null", LogLevel.Debug);
                return null;
            }
```
(you might need to double check that the above code segment actually compiles)
or
```suggestion
            if (Path == null) throw new InvalidOperationException($"Cannot construct from a null {nameof(Path)}");
```

OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
```diff
@@ -26,8 +26,7 @@ public PluginSettingStore(object? source, bool enable = true)
             {
                 var sourceType = source.GetType();
 
-                Path = sourceType.FullName ??
-                       throw new InvalidOperationException($"Could not look up full name for type {sourceType}");
+                Path = sourceType.FullName;
                 Settings = GetSettingsForType(sourceType, source);
```
gonX: I don't think this change is actually necessary to fix the source issue, as this code path is only used when getting defaults for a tablet profile (which is used in daemon and in some edge cases on UX init), in benchmarks, or in the CLI.

gonX (changes requested):
I personally think these types of issues should be fixed on the caller end instead, e.g. in the style of #4896, but this can of course be discussed.

OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
Kuuuube: Added in a log.

OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
Kuuuube: You're correct. But I removed it since there's also not really a reason to do the nullcheck anymore since `Path` is allowed to be null. I can revert this change if you want.

OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
```diff
@@ -38,17 +38,23 @@ public PluginSettingStore(string path, ObservableCollection<PluginSetting> setti
             Settings = settings;
         }
 
-        public string Path { set; get; }
+        public string? Path { set; get; }
```
gonX, line 41:
```suggestion
        // TODO: make non-nullable or similar fix, since it never makes sense to have a null/empty path? -gonX
        public string? Path { set; get; }
```

gonX (dismissed):
Since it restores behavior, add this comment then we can merge.

OpenTabletDriver.Desktop/Reflection/PluginSettingStore.cs
Kuuuube: Added.

gonX (approved):
LGTM

PR #4899, opened 2026-06-23
Fix output mode selector not working

Kuuuube:
I have no clue why this doesn't work with the existing code.

Doesn't work:
```
c => c != null ? c.OutputMode : null
```
```
c => c != null ? (PluginSettingStore?)c.OutputMode : null
```

Works:
```
c => c.OutputMode
```
```
c => (PluginSettingStore?)c!.OutputMode)
```

Compiler is pulling some bs on us. The solution in this PR is the only way I could get it to not have warnings and work.


gonX (approved):
LGTM. Note to self that null checks do not work in expression trees.

PR #4901, opened 2026-06-23
CONTRIBUTING: Restrict usage of null-suppressing operators

gonX:
Too many surprise behaviors have cropped up after trusting our GUI framework to automagically handle nulls, so here's an attempt at reeling this back for future code.

The wording can probably be improved a bit so feel free to suggest changes.


Kuuuube (approved):
Looks good. Can be edited more in another PR if needed.

PR #4902, opened 2026-06-23
CLI: Document and move `EnsureDaemonReady()` into partial relating to IPC instead

gonX:
Cleanup PR that adds xmldocs to 1 function


Kuuuube (approved):
Looks good. Works.

PR #4903, opened 2026-06-24
package.sh: Support linux-*arm64 too

JamiKettunen:
Works now on ARM64 Linux hosts natively running the script without having to define `NET_RUNTIME` in environment.


gonX (approved):
Seems fine. Thanks.

Kuuuube (approved):
Seems fine. Doesnt break anything with the actions builds.

PR #1147, opened 2021-06-06
Add Artisul M0610 Pro Config

Kuuuube:
Verification for Artisul M0610 Pro Config:
[discord](https://discord.com/channels/615607687467761684/615611007951306863/851127217258102785)
[discord](https://discord.com/channels/615607687467761684/615611007951306863/851124595942883388)
[discord](https://discord.com/channels/615607687467761684/615611007951306863/851115933904273453)

Verification for UC-Logic 1060N String (needed to avoid conflict with Artisul and any other future UC-Logic tablets):
[discord](https://discord.com/channels/615607687467761684/615611007951306863/828744043248091146)

Artisul M0610 Pro is missing tablet button support and tilt support. Current parsers and possibly current init methods don't work for its aux endpoint.

PR #4905, opened 2026-06-25
Add support for buttons and tilt on artisul m0610 variants that support it

Kuuuube:
First variant in this config was added in #1147. User helping get this supported shows some data here that only has 10 bytes. This is not enough for tilt so I've excluded tilt from this variant in the config.
bruuuuuuuuuuuuuh

Newer variant has 12 bytes which is expected for a tablet that supports tilt.
Data from newer variants (some pen movement at the top then buttons pressed in order): [tablet-data_1782406657.txt](https://github.com/user-attachments/files/29350788/tablet-data_1782406657.txt)
Strings from newer variant: [string-dump_9580-111.txt](https://github.com/user-attachments/files/29350830/string-dump_9580-111.txt)
Verification: https://discord.com/channels/615607687467761684/1519649118509858927/1519756547314417856

This tablet also has a wheel. Not adding support for it in this PR.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Artisul/M0610 Pro.json
+++ b/OpenTabletDriver.Configurations/Configurations/Artisul/M0610 Pro.json
@@ -32,7 +32,7 @@
       "VendorID": 9580,
       "ProductID": 111,
       "InputReportLength": 12,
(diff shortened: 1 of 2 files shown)


gonX (approved):
Makes sense. LGTM.

PR #4917, opened 2026-06-29
Add FilterState async plugin method

Kuuuube:
Followup to #4916 with a full solution. Preserves existing async filter behavior and stops timer polls being consumed by `DeviceReport` and `ITouchReport` without eating them.

Putting this on the 0.6.8 milestone but that could be changed. This does not break any existing plugins.

For example, if a plugin wants all reports they can override like this:
```csharp
protected override bool FilterState(IDeviceReport value) => true;
```

### DIFF
--- a/OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
+++ b/OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
@@ -2,7 +2,6 @@ using System;
 using OpenTabletDriver.Plugin.Attributes;
 using OpenTabletDriver.Plugin.DependencyInjection;
 using OpenTabletDriver.Plugin.Tablet;
-using OpenTabletDriver.Plugin.Tablet.Touch;
 using OpenTabletDriver.Plugin.Timers;
 using OpenTabletDriver.Plugin.Timing;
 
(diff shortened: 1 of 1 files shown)


OpenTabletDriver.Plugin/Output/AsyncPositionedPipelineElement.cs
```diff
@@ -67,9 +67,12 @@ public float Frequency
 
         public void Consume(T? value)
         {
-            // Block DeviceReport and ITouchReport from being consumed for now
-            if (value is DeviceReport or ITouchReport)
+            if (!FilterState(value))
+            {
+                if (value != null)
+                    Emit?.Invoke(value);
```
gonX, line 73: I may be misunderstanding the issue to begin with, but why are these lines necessary now? This seems to change the behavior in a way that I am not sure is intended

Kuuuube: It is intended and very important. If the async filter decides it doesnt want to handle a type of reports, theyre passed onto the next element in the pipeline.

The filtering here is not to allow async filters to block the pipeline but purely to allow them to save their timer polls from getting used up by junk they dont want.

If it isnt emitted then the pipeline stops there and the report doesnt go any further.

It is also important that state is not set which is why i dont call OnEmit, setting state means that UpdateState will be fed that state and defeats the point since now we have junk in the state again.

gonX: Ah, that makes sense now. If the plugin isn't interested in the specific state/report type, it'll get passed on to the next element in the pipeline instead. PR LGTM then.

gonX (approved):
LGTM. Good change.

PR #4923, opened 2026-07-03
add wheel support for innovator 16

JohnWeak1:
this adds wheel support for both wheels on the innovator 16, RelativeWheelSteps may be inaccurate but it works fine


Kuuuube:
This needs to be updated in tablets.md.

gonX:
JSON file needs linting as the format is not true JSON - it's also why the shell test using `jq` fails
(see the unit test results for the needed diff)

JohnWeak1:
Should be fixed now

gonX (approved):
LGTM, [Discord Context](https://canary.discord.com/channels/615607687467761684/668631467752161283/1522568677264986203)

PR #4928, opened 2026-07-06
Cleanup debugger recording and additional data accessibility/discoverability

Kuuuube:
Users never know where the recordings are stored so having a button to open the directory is needed. Switching to a separate directory for the recordings so they dont clog up the appdata directory. The open recordings directory button is also shown to the user when they start the recording so there shouldnt be any issue with them finding it.

Having a separate tab for recording to have that label there also helps with discoverability. And I think it's a bit cleaner to not put that in the file tab.

Also enabling the additional statistics by default should help for making configs since users will be able to more easily find the listing of maxes. This is probably more useful for most users and if a user does need to closely inspect the visualizer, they can still turn off the additional statistics.

Linux:

image
image

Windows:
image

The spacing and width of additional statistics is kinda goofy on windows here but... eto. It works at least.

### DIFF
--- a/OpenTabletDriver.Desktop/AppInfo.cs
+++ b/OpenTabletDriver.Desktop/AppInfo.cs
@@ -19,7 +19,8 @@ namespace OpenTabletDriver.Desktop
             temporaryDirectory,
             cacheDirectory,
             backupDirectory,
-            trashDirectory;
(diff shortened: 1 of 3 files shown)


gonX (approved):
LGTM. Good changes

PR #4931, opened 2026-07-07
Fix double-applied virtual-screen offset in "Set to display"

### LINKED ISSUE #4832: incorrect display area auto positioning on multi-monitor setup
https://github.com/user-attachments/assets/ad6728ec-cd11-4edc-87e9-c4c58d65bd7f

## issue description

using `set to display` will produce a display area not positioned correctly inside the monitor bounds.

i think the issue appears on multi-monitor setups, or possibly when using a portrait or flipped screen.

## observed behavior

- display area positioning appears incorrect
- monitor bounds do not match the expected display area position
- the display area is offset instead of properly aligned inside the monitor geometry

## expected behavior

the display area should be correctly aligned and centered within the monitor bounds, without offset or incorrect positioning.

## environment

- windows
- issue reproduced consistently
- video recording attached

Reonowo:
Hello, this has bugged me for a little while, so:

"Set to display" could target the wrong monitor when displays were left of or above the primary. 
This was due to a regression in [ff0c1e31](https://github.com/OpenTabletDriver/OpenTabletDriver/commit/ff0c1e31) where virtual-screen offset normalization was applied twice on Windows/X11.

I have removed the redundant offset.

Also, alignment menu actions may have a related negative-origin issue and are left for follow-up.

Closes #4832.

### DIFF
--- a/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
+++ b/OpenTabletDriver.UX/Controls/Output/AbsoluteModeEditor.cs
@@ -295,17 +295,10 @@ namespace OpenTabletDriver.UX.Controls.Output
 
                 var subMenu = base.ContextMenu.Items.GetSubmenu("Set to display");
 
-                var displays = DesktopInterop.VirtualScreen?.Displays.ToArray()
+                var virtualScreen = DesktopInterop.VirtualScreen
                     ?? throw new InvalidOperationException("Could not get VirtualScreen");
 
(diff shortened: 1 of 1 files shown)


Kuuuube:
This seems unnecessarily complex considering #4940 appears to work. Even if fetching the `xOffset` and `yOffset` from the monitors is more reliable than using the virtualScreen's position, we don't also need a refactor of this thrown in. This PR could've been two lines changed.

gonX:
Is the tree private? I cannot check this out locally to test.

PR #4932, opened 2026-07-08
Add warning to device string reader window about arbitrary use

Kuuuube:
Locking debugger size for width to make the text wrap.

image

Open to suggestions if this text seems too scary. Just throwing in a label like this seems to be the most sane way to handle this in eto. I'd like something fancy like a bar maybe and a different text color but I don't want to define a custom color and then someone with a ridiculous theme cant see the text.


gonX (approved):
LGTM. Good change

PR #4934, opened 2026-07-08
Cache device strings during detect

Kuuuube:
We probably shouldn't be blasting devices with hundreds of string requests.

Importantly caching the value here even if it fails to fetch so we don't end up requesting an inaccessible index over and over which can really slow down detection since it may have to wait for the timeout every single time.


gonX (approved):
I think this looks reasonable

PR #4937, opened 2026-07-10
Add linux support for Bosto BT-12HD

Kuuuube:
Diag: [2Diagnostics-067.json](https://github.com/user-attachments/files/29869854/2Diagnostics-067.json)

Diag shows it failing to detect from the tablet refusing a string init but once removed it did detect fine. It's possible the windows config doesnt need those init but windows may not care if the tablet refuses the string requests, unsure.

Verification: https://discord.com/channels/615607687467761684/723399565780189234/1524918004192645231


gonX (approved):
Seems harmless

PR #4938, opened 2026-07-10
Add menu item for opening presets directory

Kuuuube:
If a user wants to delete a preset or rename a preset theyre kinda just on their own to figure it out which is not great. We should at least have a way to easily open the directory to help out there.

Also small change to the order of menu items here:
image

Save, Refresh, Open > Refresh, Save, Open. A bit nicer imo and also roughly matches what we do in plugin manager (Install, Refresh, Alt source, Open). The first thing a user will need to do is save the preset, not refresh their presets.


gonX (approved):
LGTM

PR #4075, opened 2025-09-14
[0.6.x] MacOS: Add experimental ARM64 build and release support

AkiSakurai:
This commit introduces comprehensive support for building, testing, and releasing OpenTabletDriver natively for macOS on the ARM64 architecture (Apple Silicon).
  Key changes include:
  CI/CD:
  
  Added a macospublish-arm64 job to dotnet.yml to build and upload ARM64 artifacts.
  Added a macos_arm64 job to release.yml to create and upload ARM64 release packages.
  
  Build System:
  
  The eng/macos/package.sh script now auto-detects the architecture (arm64 or x64) and builds for the corresponding runtime.
  
  The script performs ad-hoc codesigning on the application bundle for both ARM64 and x64 binaries.
  Codesigning is required for ARM64 binaries to run on Apple Silicon but is not required for x64 binaries; however, codesigning is applied to both for consistency.
  
  An earlier version of Eto had a bug where the automatic bundle was not generated when an output folder was specified.
  This bug has been fixed in Eto version 2.10.2, which now generates the bundle automatically.
  Since we manually create the application bundle, this commit disables Eto's automatic .app creation during publishing in the build script to prevent nested bundle creation.
  
  Project Configuration:
  
  Upgraded Eto.Platform.Mac64 to version 2.10.2.
  This version includes critical fixes for UI rendering, stability, universal binary compatibility on ARM64, and resolves the issue with extraneous bundle creation.
  
  Updated OpenTabletDriver.UX.MacOS.csproj to include resources like the app icon and Info.plist for Eto's automatic bundle generation.
  
  Added UseCurrentRuntimeIdentifier to the project configuration to allow Rider to build and execute using the host's architecture.

PR #4943, opened 2026-07-15
Add baseline macos arm64 building

Kuuuube:
Pulling in the csproj change from Aki (#4075) that allows the gui to work on macos arm64. And some simple packaging docs.

Verification: https://discord.com/channels/615607687467761684/723399607522033664/1527082880960823509

Excluding CI changes at the moment.


gonX (approved):
Seems like an upgrade

PR #4944, opened 2026-07-16
Move pressure rewriting to early filter

### LINKED ISSUE #4921: AsyncPositionedPipelineElement repeatedly generates clicks while pen is down
### Steps to Reproduce

Install a plugin that uses AsyncPositionedPipelineElement
(for example https://github.com/make-42/otd-frenet-serret-frame-predictor or https://github.com/X9VoiD/VoiDPlugins/wiki/MeLInterp)

Press your pen down.

### Expected Behavior

Only one click sent whenever the pen goes above the threshold.

### Observed Behavior

Repeated clicks sent.

### Tablet Device

Wacom CTL-472

### Diagnostic Information

(i was testing with a modified version of the plugin to isolate the issue)

The diagnostics are kinda lare (7.2 MB), be warned.

[Diagnostics-067 Wacom CTL-472.json](https://github.com/user-attachments/files/29614025/Diagnostics-067.Wacom.CTL-472.json)

### Acknowledgements

- [x] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [x] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.

gonX:
The previous behavior seems to have been written in a way that expected the function parameters being passed around to behave like by-value, given that the tablet reports are structs, but as we're accessing the struct through interface types, we're actually getting a reference instead, so report clones don't exactly behave as it seems was intended.

This can be seen in OpenTabletDriver/OpenTabletDriver#4921 - basically, any async plugin that wasn't setting the reports `Pressure` on `Update()` _and_ threshold bindings were in use, especially on higher thresholds, users would experience the very annoying behavior of having their bindings repeatedly toggle, as the report to be `Consume()`d by the async filters would correctly be cloned, but the pressure value would be continuously be affected by the pressure rewrite as the report passed through the binding handler multiple times, leading to pressure values above the threshold (but not maxed) eventually having their pressure values rewritten to 0 as the now-incorrect pressure value was being compared with, leading to bindings being released even when they shouldn't have.

This commit then replaces this entire logic and moves it as early in the filter chain as possible, so that instead of the pressure being rewritten _after_ any plugins might have saved the reference to the report, the pressure is instead rewritten long before any plugins have a chance of seeing the report.

Fixes #4921

### DIFF
--- a/OpenTabletDriver.Daemon/DriverDaemon.cs
+++ b/OpenTabletDriver.Daemon/DriverDaemon.cs
@@ -417,13 +417,20 @@ namespace OpenTabletDriver.Daemon
         {
             string group = dev.Properties.Name;
 
+            var pressureRewriteFilter = new PressureRewriteFilter
+            {
+                TipPressureThreshold = profile.BindingSettings.TipActivationThreshold,
+                EraserPressureThreshold = profile.BindingSettings.EraserActivationThreshold,
+                MaxPenPressure = dev.Properties.Specifications.Pen.MaxPressure,
+            };
+
             var elements = (from store in profile.Filters
                             where store is { Enable: true }
                             let filter = store!.Construct<IPositionedPipelineElement<IDeviceReport>>(outputMode.Tablet)
                             where filter != null
                             select filter!).ToArray();
 
-            outputMode.Elements = elements.Append(bindingHandler).ToList();
+            outputMode.Elements = elements.Prepend(pressureRewriteFilter).Append(bindingHandler).ToList();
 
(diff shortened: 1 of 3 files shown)


Kuuuube (approved):
Looks good, works.

Noting that this is kindof breaking for Slimy Scylla, Tablet Debounce, and Hover Distance Limiter. Will need some tweaks there. Not aware of any other plugins that expect to see the raw pressure.

PR #4947, opened 2026-07-18
Recheck configuration directory on detect button press

Kuuuube:
Making users restart the daemon is inconvenient here. Particularly an issue for Linux users.

Also did some cleanup. Some bits were all over the place.

The daemon setting `AppDataDirectory` and `ConfigurationDirectory` directly on startup if there were args without any way of knowing later that the daemon did this. There's `GetDefaultConfigurationDirectory` but then `ConfigurationDirectory` could get set by the getter of `AppInfo Current` instead.

Now `GetDefaultConfigurationDirectory` covers everything for getting the configuration directory.


gonX (approved):
Yeah this makes sense

PR #4955, opened 2026-07-19
Add support for detecting renamed Digimend tablet driver

### LINKED ISSUE #4954: [Linux] Foreign Driver Detection does not find new Digimend driver
Module name `hid_digimend_uclogic`

[Found in the wild](https://canary.discord.com/channels/615607687467761684/723399565780189234/1528421920435994826) via `digimend-dkms` on PikaOS, which is Debian based.

gonX:
On at least Debian, the driver is now named `hid_digimend_uclogic`

Fixes #4954


Kuuuube (approved):
This is good but it's worth noting that `LinuxModuleName` is used as a regex string. So the existing `hid_uclogic` could be easily extended instead if desirable.

PR #4959, opened 2026-07-20
(minor) Remove superfluous nullable preprocessor directives

gonX:
The projects already have nullable enabled, so these do nothing.

I can't find a .NET analyzer setting for this, only ReSharper, and it is already defaulted to a warning: https://www.jetbrains.com/help/rider/RedundantNullableDirective.html - so no accompanying `.editorconfig` change here.


gonX:
Oops, accidentally committed a commit to the wrong branch and pushed. Will force-push it out in a second.

Kuuuube (approved):

PR #4963, opened 2026-07-27
Add support for huion KD200

Kuuuube:
Verification: https://discord.com/channels/615607687467761684/1179851852523765960/1531167497766834307

Diag (undetected): [Diagnostics-067 Huion KD200.json](https://github.com/user-attachments/files/30400065/Diagnostics-067.Huion.KD200.json)

String 201 (from a different user but matches this user's tablet too):
image

Data is muddy since user had the pen on the tablet while getting the data.
Aux buttons and wheel data: [tablet-data_1785127454.txt](https://github.com/user-attachments/files/30400049/tablet-data_1785127454.txt)
Wheel data: [tablet-data_1785127660.txt](https://github.com/user-attachments/files/30400056/tablet-data_1785127660.txt)

The config creation thread for this tablet has some other users with a variant of this tablet that has pid 100 rather than 109. Not doing anything preemptive there.

Although this tablet has a note in tablets.md I'm not putting it in quirks since this is simply the intended behavior of the tablet. Note added since it's possible users could be confused about it. The tablet has half of a keyboard tacked onto the side of it and most of the keyboard simply acts as a regular keyboard device separate from the tablet. Keys K1-K5 and the profile switch key are the 6 keys that act as customizeable aux buttons.

[Huion says](https://www.huion.com/products/inspiroy-keydial-kd200) the tablet has "23 Standard keys, 5 Programmable Press Keys+Dial" but for some reason they dont count the profile switching button which reports as an aux key. So for OTD's purposes we have 6 aux buttons.

image

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Huion/KD200.json
+++ b/OpenTabletDriver.Configurations/Configurations/Huion/KD200.json
@@ -1,0 +1,41 @@
+{
+  "Name": "Huion KD200",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 228.6,
+      "Height": 142.875,
+      "MaxX": 45720,
+      "MaxY": 28575
+    },
+    "Pen": {
+      "MaxPressure": 8191,
+      "ButtonCount": 2
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 6
+    },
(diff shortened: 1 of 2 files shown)


gonX (approved):
LGTM

PR #4968, opened 2026-07-30
XP-Pen Deco Pro Medium: add support for the second wheel within the trackpad

Raekellie:
This is basically a copy of what is there for the XP-Pen Deco Pro Small, as it's effectively the same thing but bigger. Works just fine with my device.

While doing this I realised there are a few other tablets with wheels but no explicit support for them in the config file, such as the "Deco Pro MW Gen2"; is there a particular reason for this, given the parsing code is already there? This was just an example, by the way, I only own the Deco Pro M.

PS.: are there plans/is it within OTD's scope to add trackpad support? Or, if it already exists, what steps do I need to take in regards to the configuration file?

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco Pro Medium.json
+++ b/OpenTabletDriver.Configurations/Configurations/XP-Pen/Deco Pro Medium.json
@@ -13,7 +13,13 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 8
-    }
+    },
+    "Wheels": [
+      {
+        "RelativeWheelSteps": 24,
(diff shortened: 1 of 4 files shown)


Mrcubix:
[tablet-data_1785444797.txt](https://github.com/user-attachments/files/30562878/tablet-data_1785444797.txt)

Data recording from the author

Raekellie:
Thanks to @Mrcubix , who kindly coded this in, there is now support to use the tablet's second wheel, which is part of the trackpad.

Using its alternate mode, plain trackpad, seems to require some deeper changes to OTD to allow the endpoint to be handled as a regular mouse.

PR #4970, opened 2026-08-01
Configurations: Add ViewSonic ID1330

rinnyanneko:


Mrcubix:
You'll need to provide a diagnostic file `Help > Export diagnostics...`.
If possible, also provide a data recording from the tablet debugger `File > Data recording` :

- Move your pen to the bottom right of the tablet (if not physically flipped),
- Use the various buttons.

I observe a lack of initialization items, such as a string index, or base64 encoded feature data, used for initialization
(from PNP to a venfor interface).

Does this device only have a vendor interface and doesn't actually have any PNP support ?

rinnyanneko:
The device does have PNP pen support. Interface 0 provides the standard Digitizer/Pen report (`0x10`) and the vendor express-key report (`0xAC`), and both start reporting immediately, so no initialization is needed. Interface 1 only has an empty 7-byte HID collection.

[ViewSonic-ID1330-diagnostics.json](https://github.com/user-attachments/files/30626253/ViewSonic-ID1330-diagnostics.json)
[ViewSonic-ID1330-tablet-data.txt](https://github.com/user-attachments/files/30626341/ViewSonic-ID1330-tablet-data.txt)

Mrcubix:
As it is, this tablet most likely won't work on windows without WinUSB installed on Interface 0. 

Kinda odd, considering woodpads takes `AhABAAA=` as a feature init to switch to vendor mode. But dozsn't seem like there is another interface here (either hidden, or the device has different handling on linux)

rinnyanneko:
alright, it should work on Windows now

Mrcubix:
Do you have a diag on hand for Windows ?

rinnyanneko:
[Diagnostics-067+188-gcd599b1e-dirty ViewSonic ID1330.json](https://github.com/user-attachments/files/30667428/Diagnostics-067%2B188-gcd599b1e-dirty.ViewSonic.ID1330.json)
[tablet-data_1785771056.txt](https://github.com/user-attachments/files/30667429/tablet-data_1785771056.txt)

Mrcubix:
That vendor interface does seems to provide data that seem more in line with what we can see in the woodpad parser.
Actually it's the same structure.

Mrcubix:
Reading the specs sheet, i'm noticing that the vendor mode specs are mo accurate than what is in the config.

The digitizer width is most likely not exactly 300mm & in fact, the specsheet says the display is 294mm x 165mm.
I expect the digitizer to be the same size, if not slightly smaller.

The height would then most likely be what you have found : 29376 (MaxX) / 100 (lines per mm) = 293.76mm
As for the height, it's most likely 16523(MaxX) / 100 (lines per mm) = 165.23mm.

Like most display tablet, this one will require some calibration plugin (and afaik, there are 2 of those that exists currently)

Mrcubix:
If the manufacturer drivers are installed, then you might have to uninstall them first, then replug, before being able to see the vendor endpoint (14 0 5).

It's possible that one of the kernel modules like `wacom`, `hid_uclogic`, `hid_digimend_uclogic`, or just the digimend package itself `digimend-dkms` are installed and hiding endpoint for this device.

If that work, then apply the things i suggested to the config, and you'll prob end up using the woodpad parser instead.

rinnyanneko:
emm... the menufacturer only provide driver for Windows and MacOS (I already uninstalled, of course), do I need to uninstall anything on my CachyOS (a distro based on Arch Linux) before testing?
https://www.viewsonic.com/global/support/article?title=Is%20there%20Linux%20drivers%20for%20my%20monitor?&articleId=33000222292
image

btw, the buttons seems different with woodpad
image
image

Mrcubix:
The kernel modules i sent needs to be rmmod and the digimend package uninstalled.
Let me know if anything got removed.

If one get removed, replug your tablet, and check the diagnostic file for the missing 14 input, 5 feature length endpoint.

rinnyanneko:
seems both HID interfaces are bound to hid-generic, mods mentioned above are not loaded
```fish
~
❯ for d in /sys/bus/hid/devices/*0543:E659*
      readlink -f "$d/driver"
  end
/sys/bus/hid/drivers/hid-generic
/sys/bus/hid/drivers/hid-generic

~
❯ for m in wacom hid_uclogic hid_digimend_uclogic
      if test -d /sys/module/$m
          echo "$m: loaded"
      else
          echo "$m: not loaded"
      end
  end
wacom: not loaded
hid_uclogic: not loaded
hid_digimend_uclogic: not loaded
```

PR #4985, opened 2026-08-10
Add mouse scroll support to the evdev virtual tablet (Artist Mode)

mugaaaaa:
The Linux Artist Mode output only creates a virtual pen tablet, so mouse scroll bindings (touch ring/wheel) could not emit scroll events on Wayland: the binding's `[Resolved]` `IMouseScrollHandler` was null and the daemon logged `IMouseScrollHandler unavailable`.

This creates a separate evdev virtual mouse for scroll events and makes the virtual tablet implement `IMouseScrollHandler`, so wheel bindings work in Artist Mode alongside the pen tablet.

The scroll device enables the same REL codes (`REL_WHEEL`, `REL_WHEEL_HI_RES`, `REL_HWHEEL`, `REL_HWHEEL_HI_RES`) as the existing `EvdevAbsolutePointer`.

### Testing
Verified with a Gaomon M6 touch ring on Wayland (niri): scrolling works in Artist Mode, and the pen remains in tablet mode.

### DIFF
--- a/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
+++ b/OpenTabletDriver.Desktop/Interop/Input/Absolute/EvdevVirtualTablet.cs
@@ -8,13 +8,15 @@ using OpenTabletDriver.Plugin.Platform.Pointer;
 
 namespace OpenTabletDriver.Desktop.Interop.Input.Absolute
 {
-    public class EvdevVirtualTablet : IPenActionHandler, IAbsolutePointer, IPressureHandler, ITiltHandler, IEraserHandler, IHoverDistanceHandler, ISynchronousPointer, IDisposable
+    public class EvdevVirtualTablet : IPenActionHandler, IAbsolutePointer, IPressureHandler, ITiltHandler, IEraserHandler, IHoverDistanceHandler, IMouseScrollHandler, ISynchronousPointer, IDisposable
     {
         private const int RESOLUTION = 1000; // subpixels per screen pixel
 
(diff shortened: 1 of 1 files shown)


Mrcubix:
This PR has multiple issues.

We cannot know that the DE will actually scroll in the place the pen cursor is at.
Depending on their implementation, they might only scroll at the mouse cursor, which will end up somewhere else inevitably.

So in summary, it will work in some DE, not others.

Another issue, is re-implementing stuff from other pointers.
This results in not only duplicate code, but you might as well put every other functionalities in one pointer, and that doesn't make sense.

Another more blatant issue, is that you're instanciating yet another evdev device, just to scroll, when we still have an active absolute mouse pointer you could use.


The thing i proposed not that long ago for the kind of issue, was to have fallback pointers.
Aka some pointers advertise capabilities.

For any roles not filled within an output mode, we take a pointer with said capability as a fallback.
This means we won't lack some functionalities while using specific output modes.

As a side note, the now obselete Scroll Binding Plugin Bindings, already provides the functionality you're looking for.
(The plugin ended up being integrated into OTD, which is why is it marked as obselete)

We could do something simillar in the integrated Binding, but specify that the pointer is a fallback one.
This however means, we need a way for Bindings to receive pen coordinates on the regular.
Otherwise you will end up scrolling the wrong app on any platforms that isn't some specific Linux DEs

gonX:
See #4804 - this is already being worked on. But as mentioned by Cubix, this type of solution (which the PR I linked also does), does not offer predictable behavior yet.

PR #4991, opened 2026-08-16
Parblo Intangbo S: wheel support

Un1nstallexe:
This PR adds wheel support for Parblo Intangbo S. 

Before, this tablet used a generic XP Pen parser and a config that did not describe wheel properties. This led to:
- No wheel settings in the UI;
- Wheel manipulation being confused with button presses by OTD;
- Wheel button not being recognized and confused with some other buttons as well.

### Reporting structure

I will omit the report types which are related to the digitizer input, because the generic XP reporters handle that part. That is, I will only describe the case when the tablet notifies about button press / wheel rotation. 

The way the report is structured is as follows:
When `report[1].IsBitSet(4)`, that is, the tablet reports about buttons/wheel, we get 10 bytes of leftover data. the 3rd byte is used for all the buttons, excluding the wheel button. In other words, 6 bits in 8th byte are used for auxillary buttons, a bit per button, and 2 bits left unused. 

When any button is released, *all* bytes starting from the 3rd are zeroed. When multiple buttons are pressed, the tablet would only set the bit that corresponds to the last pressed button. Those 2 quirks make it impossible to handle any multiple presses situation. 

The wheel report contains multiple set bits. First, the 3rd byte is set to 8, i.e. `report[2].isBitSet(3) == true`. Note that this bit also corresponds to the 4th aux button on the tablet. Second, the 4th byte is set to 1 or 0000 0001 (clockwise turn), 2  or 0000 0010 (counter-clockwise turn) or 3 or 0000 0011 (wheel button). Also note that button release leads to same result as with aux buttons - all bytes are zeros starting from the 3rd.

#### Some examples of raw tablet data

| What I do | What tablet says |
| ------ | ------ |
| Nothing pressed | `02 F0 00 00 00 00 00 00 00 00` |
| Button 1 pressed | `02 F0 01 00 00 00 00 00 00 00` |
| Button 3 pressed | `02 F0 04 00 00 00 00 00 00 00` |
| Button 3 pressed before button 1 is released | `02 F0 04 00 00 00 00 00 00 00` |
| Button 4 pressed | `02 F0 08 00 00 00 00 00 00 00` |
| Button 6 pressed | `02 F0 20 00 00 00 00 00 00 00` |
|Any button released| `02 F0 00 00 00 00 00 00 00 00` |
| Wheel - turn clockwise |  `02 F0 08 01 00 00 00 00 00 00` |
| Wheel - turn counterclockwise |  `02 F0 08 02 00 00 00 00 00 00` |
| Wheel button pressed |  `02 F0 08 03 00 00 00 00 00 00` |
### Implementation details

I moved the decision on whether the report is about wheel & wheel button or about aux buttons to the reporter, away from parser, for a reason. As mentioned before, the button release report data is identical regardless of what button has been released. So to zero out all the buttons, i.e. both the aux buttons and the center wheel button, when zeros are received, I added every aux feature to one report, couldn't come up with more elegant solution :). If anyone has a better idea, please let me know, I'd be happy to improve the code here. It's not that I have much experience with C# or USB or OTD tbh.

I ran `dotnet format` and `dotnet test` before pushing so those things should be OK, although `dotnet format` found a file out of scope of my changes that has some formatting issues, some other config for another tablet. I removed the changes to that file.

The docs suggest that I might want to follow some sort of convention regarding commit messages and references commits page of that github repo, although I didn't really find that there is any unified style for that. So I tried to get the average idea, hope it's OK. 

### Testing

This works on NixOS 26.11. I didn't test it anywhere else.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Parblo/Intangbo S.json
+++ b/OpenTabletDriver.Configurations/Configurations/Parblo/Intangbo S.json
@@ -13,7 +13,13 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 6
-    }
+    },
+    "Wheels": [
+      {
+        "RelativeWheelSteps": 24,
+        "ButtonCount": 1
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
@@ -31,7 +37,7 @@
       "ProductID": 40980,
       "InputReportLength": 1101,
       "OutputReportLength": 1101,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.XP_Pen.XP_PenReportParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Parblo.ParbloIntangboSParser",
       "OutputInitReport": [
         "ArAE"
       ]
--- a/OpenTabletDriver.Configurations/Configurations/Parblo/Intangbo S.json
+++ b/OpenTabletDriver.Configurations/Configurations/Parblo/Intangbo S.json
@@ -37,7 +37,7 @@
       "ProductID": 40980,
       "InputReportLength": 1101,
       "OutputReportLength": 1101,
-      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Parblo.ParbloIntangboSParser",
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Parblo.ParbloIntangboSReportParser",
       "OutputInitReport": [
         "ArAE"
       ]
--- a/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSAuxReport.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSAuxReport.cs
@@ -1,0 +1,60 @@
+using OpenTabletDriver.Plugin.Tablet;
+using OpenTabletDriver.Plugin.Tablet.Wheel;
+
+namespace OpenTabletDriver.Configurations.Parsers.Parblo
+{
+    public struct ParbloIntangboSAuxReport : IRelativeWheelReport, IWheelButtonReport, IAuxReport
+    {
+        public ParbloIntangboSAuxReport(byte[] report)
+        {
+            Raw = report;
+            int wheelDataIndex = 3;
+            int auxIndex = 2;
+            if (report[auxIndex] != 0 && report[wheelDataIndex] != 0)
+            {
+                AnalogDeltas =
+                [
+                    report[wheelDataIndex] switch
+                    {
+                        1 => 1,
+                        2 => -1,
+                        _ => 0,
+                    },
+                ];
+                WheelButtons =
+                [
+                    [
+                        report[wheelDataIndex] switch
+                        {
+                            3 => true,
+                            _ => false,
+                        },
+                    ],
+                ];
+                AuxButtons = [false, false, false, false, false, false];
+            }
+            else
+            {
+                AnalogDeltas = [0];
+                WheelButtons =
+                [
+                    [false],
+                ];
+                AuxButtons =
+                [
+                    report[auxIndex].IsBitSet(0),
+                    report[auxIndex].IsBitSet(1),
+                    report[auxIndex].IsBitSet(2),
+                    report[auxIndex].IsBitSet(3),
+                    report[auxIndex].IsBitSet(4),
+                    report[auxIndex].IsBitSet(5),
+                ];
+            }
+        }
+
+        public byte[] Raw { set; get; }
+        public int[] AnalogDeltas { get; set; }
+        public bool[][] WheelButtons { get; set; }
+        public bool[] AuxButtons { get; set; }
+    }
+}
--- a/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSReportParser.cs
@@ -1,0 +1,26 @@
+using System.Diagnostics.CodeAnalysis;
+using OpenTabletDriver.Configurations.Parsers.XP_Pen;
+using OpenTabletDriver.Plugin.Tablet;
+
+namespace OpenTabletDriver.Configurations.Parsers.Parblo
+{
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
+    public class ParbloIntangboSParser : IReportParser<IDeviceReport>
+    {
+        public IDeviceReport? Parse(byte[]? report)
+        {
+            if (report == null || report.Length == 0)
+                return null;
+            if (report[1] == 0xC0)
+                return new OutOfRangeReport(report);
+            if (report[1].IsBitSet(4))
+                return new ParbloIntangboSAuxReport(report);
+            if (report.Length >= 12)
+                return new XP_PenTabletOverflowReport(report);
+            else if (report.Length >= 10)
+                return new XP_PenTabletReport(report);
+            else
+                return new TabletReport(report);
+        }
+    }
+}
--- a/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Parblo/ParbloIntangboSReportParser.cs
@@ -5,7 +5,7 @@ using OpenTabletDriver.Plugin.Tablet;
 namespace OpenTabletDriver.Configurations.Parsers.Parblo
 {
     [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
-    public class ParbloIntangboSParser : IReportParser<IDeviceReport>
+    public class ParbloIntangboSReportParser : IReportParser<IDeviceReport>
     {
         public IDeviceReport? Parse(byte[]? report)
         {
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -75,6 +75,7 @@
 | Parblo A609                        |     Supported     |
 | Parblo A610 Pro (Variant 2)        |     Supported     |
 | Parblo A640 V2                     |     Supported     |
+| Parblo Intangbo S                  |     Supported     |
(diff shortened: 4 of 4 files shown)


Mrcubix:
A lot of this could just be replaced with just a copy of the normal xp-pen report parser, just by providing a different wheel index as arg when instanciating the aux report.

Un1nstallexe:
> A lot of this could just be replaced with just a copy of the normal xp-pen report parser, just by providing a different wheel index as arg.

Maybe I just didn't get how it works but I think it needed wheels and buttons to use separate parts of report. I'll recheck that.

Un1nstallexe:
Well, no, correct me if I'm dumb, but I see no way how this would work and also what indexes should I put there?
```csharp
public XP_PenAuxReport(byte[] report, int auxIndex = 2, int wheelIndex = 7) 
        {
            Raw = report;

            AuxButtons =
            [
                report[auxIndex].IsBitSet(0),
                report[auxIndex].IsBitSet(1),
                report[auxIndex].IsBitSet(2),
                report[auxIndex].IsBitSet(3), // <- this will be set no matter what kind of report that is, whether it's wheel rotation or 4th button. 
                report[auxIndex].IsBitSet(4), // this reporter assumes wheel and buttons have different indexes in report data
                report[auxIndex].IsBitSet(5), 
                report[auxIndex].IsBitSet(6), 
                report[auxIndex].IsBitSet(7),
                report[auxIndex + 1].IsBitSet(0),
                report[auxIndex + 1].IsBitSet(1),
                report[auxIndex + 1].IsBitSet(2),
                report[auxIndex + 1].IsBitSet(3),
                report[auxIndex + 1].IsBitSet(4),
                report[auxIndex + 1].IsBitSet(5),
                report[auxIndex + 1].IsBitSet(6),
                report[auxIndex + 1].IsBitSet(7),
                report[auxIndex + 2].IsBitSet(0),
                report[auxIndex + 2].IsBitSet(1),
                report[auxIndex + 2].IsBitSet(2),
                report[auxIndex + 2].IsBitSet(3),
            ];

            // 0x01 for 1st wheel clockwise, 0x02 for counterclockwise, verified on XP Pen Artist 13.3 Pro V2
            // 0x10 for 2nd wheel clockwise, 0x20 for counterclockwise, verified on XP Pen Artist 22R Pro
            AnalogDeltas =
            [
                report[wheelIndex].IsBitSet(0) ? 1 : report[wheelIndex].IsBitSet(1) ? -1 : 0,
                report[wheelIndex].IsBitSet(4) ? 1 : report[wheelIndex].IsBitSet(5) ? -1 : 0,
            ];
        }
```

Un1nstallexe:
I'm referring to that part just in case.
> The wheel report contains multiple set bits. First, the 3rd byte is set to 8, i.e. report[2].isBitSet(3) == true. Note that this bit also corresponds to the 4th aux button on the tablet. Second, the 4th byte is set to 1 or 0000 0001 (clockwise turn), 2 or 0000 0010 (counter-clockwise turn) or 3 or 0000 0011 (wheel button). Also note that button release leads to same result as with aux buttons - all bytes are zeros starting from the 3rd.

Mrcubix:
Can you provide an actual data recording from the tablet debugger > File  > Data Recording ?
Once done, disable the option and check ~/config/OpenTabletDriver for the tablet-data file.txt .

Un1nstallexe:
OK, here's the recording, OTD v0.6.7 from nixpkgs.

If you wanted recording from OTD version in this PR please let me know.

Actions: 
- Press & release buttons 1 to 6,  
- rotate wheel counterclockwise, 
- rotate wheel clockwise, 
- press & release wheel button.

```txt
Device:Parblo Intangbo S, { 02 F0 01 00 00 00 00 00 00 00 }, Delta:5724.782ms, AuxButtons:[True False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:115.293ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 02 00 00 00 00 00 00 00 }, Delta:528.624ms, AuxButtons:[False True False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:110.189ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 04 00 00 00 00 00 00 00 }, Delta:518.648ms, AuxButtons:[False False True False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:107.936ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 08 00 00 00 00 00 00 00 }, Delta:530.120ms, AuxButtons:[False False False True False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:94.996ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 10 00 00 00 00 00 00 00 }, Delta:548.903ms, AuxButtons:[False False False False True False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:116.843ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 20 00 00 00 00 00 00 00 }, Delta:493.303ms, AuxButtons:[False False False False False True False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:111.849ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 08 01 00 00 00 00 00 00 }, Delta:2073.733ms, AuxButtons:[False False False True False False False False True False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 08 02 00 00 00 00 00 00 }, Delta:1503.204ms, AuxButtons:[False False False True False False False False False True False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 08 03 00 00 00 00 00 00 }, Delta:1313.357ms, AuxButtons:[False False False True False False False False True True False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Device:Parblo Intangbo S, { 02 F0 00 00 00 00 00 00 00 00 }, Delta:118.670ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
Additional Statistics:
  Report Rate:
    Minimum Value: 95.00 ms
    Maximum Value: 5724.78 ms
  Tablets Parsed:
    Parblo Intangbo S: 16
  Report Parser:
    XP_Pen.XP_PenAuxReport: 16
  Aux Buttons:
    0: Press Down
    1: PASS
    2: PASS
    3: PASS
    4: PASS
    5: PASS
    Status: 2✔ 3✔ 4✔ 5✔ 6✔ 1↓️️
```

Un1nstallexe:
This was mentioned in the initial comment to that PR, but I think the reason why I needed a new reporter isn't highlighted well enough.

Taking the recording I provided (or the table in the very beginning),  consider the following 2 outputs:
```
02 F0 08 00 00 00 00 00 00 00  - button 4 press
02 F0 08 01 00 00 00 00 00 00 - clockwise wheel turn
       ^
```
 The byte that indicates the 4th button press (the one below the wheel) is also active when there's a wheel turn.

So, when we know that's a wheel we should ignore 4th button input. The issue with the XP Pen Aux Report is that it **will** read the 4th button as active, no matter what offset is set for the wheel. The generic XP Aux Report basically assumes that buttons and wheels use different bits and can be read simultaneously and independently. But in that case the state of the buttons depends on the state of the wheel. The reporter in this PR on the other hand makes sure there is no collision between this.

Another approach would be using one XP Pen Aux Reporter that would only report the buttons, and another that would report the wheel. With appropriate offsets, one reporter would always return 0's in buttons + wheel data, and another would always report button states + 0's in wheel deltas. The issue is, this won't work with the wheel button, because its report looks like this:
```
02 F0 08 03 00 00 00 00 00 00
          ^
this byte reads as 0000 0011
```
That is, when the wheel button is pressed, 2 bits are set to 1, which eliminates any possibility to use `XP_PenAuxReport` as its logic is based on the system where the state of one button corresponds to a single bit, i.e. 
```csharp
report[auxIndex + foo].IsBitSet(bar);
```

And a small note from semantic perspective, `XP_PenAuxReport`  doesn't implement `IWheelButtonReport`. That would mean that, if someone finds a way for it to work with this tablet properly, the wheel button will not be detected as... a wheel button. It would be a 7th aux button. It is nothing significant, I know, but anyways.

PR #4993, opened 2026-08-21
Set AppInfo.Version property to backing AppInfo.assemblyVersion static field

Mrcubix:
Instead of only feeding the version when building `AppInfo.Current`, we can feed it directly to the property.
Also add some comments to warn developers about using the `assemblyVersion` static property.


gonX (approved):
LGTM

PR #5002, opened 2026-08-28
Add new Gaomon S830 variant

Kuuuube:
Diag (undetected): [Diagnostics-067-58.json](https://github.com/user-attachments/files/31575428/Diagnostics-067-58.json)
String 201: `GM001_T204_220627`
Confirmation it doesnt do weird bs with strings: https://discord.com/channels/615607687467761684/615611007951306863/1543000181488885760
Data recording: [tablet-data_1787950627.txt](https://github.com/user-attachments/files/31575418/tablet-data_1787950627.txt)

This tablet does not have the same weird thing where it would roll its strings after they're requested so we can be normal and use string 201 here.

### DIFF
--- a/OpenTabletDriver.Configurations/Configurations/Gaomon/S830.json
+++ b/OpenTabletDriver.Configurations/Configurations/Gaomon/S830.json
@@ -27,6 +27,18 @@
       "InitializationStrings": [
         200
       ]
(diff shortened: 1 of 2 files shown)


gonX (approved):
LGTM

PR #5005, opened 2026-08-30
Add support for veikk vk850

### LINKED ISSUE #5009: Add support for VEIKK VK850
### Manufacturer

VEIKK

### Product Name

VK850

### Device Width

_No response_

### Device Height

_No response_

### Pressure Levels

_No response_

### Lines per Inch (LPI)

_No response_

### Additional Information

pleaase support this

### Diagnostic Information

-

### Acknowledgements

hadis-crystalis:
Added support for the VEIKK VK850.

Verified with my own tablet.

Maxes at the bottom right corner:
image

Diagnostics:
[Diagnostics-067 VEIKK VK850.json](https://github.com/user-attachments/files/31613829/Diagnostics-067.VEIKK.VK850.json)

### DIFF
--- a/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
@@ -12,6 +12,7 @@ namespace OpenTabletDriver.Configurations.Parsers.Veikk
             {
                 0x41 when report[2] is 0xC0 => new OutOfRangeReport(report),
                 0x41 => new VeikkTiltTabletReport(report),
+                0x42 when report[2] is 0x03 => new VeikkRelativeWheelReport(report),
                 0x42 => new VeikkAuxReport(report),
                 0x43 => new DeviceReport(report),// touchpad report not supported yet
                 _ => new DeviceReport(report),
--- a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
+++ b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
@@ -13,7 +13,13 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 10
-    }
+    },
+    "Wheels": [
+      {
+        "RelativeWheelSteps": 24,
+        "ButtonCount": 0
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
--- a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
+++ b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
@@ -8,7 +8,7 @@
       "MaxY": 25400
     },
     "Pen": {
-      "MaxPressure": 16384,
+      "MaxPressure": 16383,
       "ButtonCount": 2
     },
     "AuxiliaryButtons": {
--- a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
+++ b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
@@ -1,0 +1,35 @@
+{
+  "Name": "VEIKK VK850",
+  "Specifications": {
+    "Digitizer": {
+      "Width": 203.2,
+      "Height": 127.0,
+      "MaxX": 40640,
+      "MaxY": 25400
+    },
+    "Pen": {
+      "MaxPressure": 16384,
+      "ButtonCount": 2
+    },
+    "AuxiliaryButtons": {
+      "ButtonCount": 10
+    }
+  },
+  "DigitizerIdentifiers": [
+    {
+      "VendorID": 12267,
+      "ProductID": 32,
+      "InputReportLength": 13,
+      "OutputReportLength": 9,
+      "ReportParser": "OpenTabletDriver.Configurations.Parsers.Veikk.VeikkTiltReportParser",
+      "OutputInitReport": [
+        "CQEE",
+        "CQIC",
+        "CQMC"
+      ]
+    }
+  ],
+  "Attributes": {
+    "libinputoverride": "1"
+  }
+}
--- a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
+++ b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
@@ -3,7 +3,7 @@
   "Specifications": {
     "Digitizer": {
       "Width": 203.2,
-      "Height": 127.0,
+      "Height": 127,
       "MaxX": 40640,
       "MaxY": 25400
     },
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -317,6 +317,7 @@
 | VEIKK Voila (VO1060)               |  Missing Features | Wheel is not yet supported.
 | VEIKK VK1060PRO                    |  Missing Features | Wheels not yet supported.
 | VEIKK VK2200PRO                    |  Missing Features | Wheels not yet supported, tablet buttons may be mapped incorrectly.
+| VEIKK VK850                        |  Missing Features | Wheel is not yet supported.
 | Wacom CTE-450                      |  Missing Features | Wheel is not yet supported.
 | Wacom CTE-650                      |  Missing Features | Wheel is not yet supported.
 | Wacom CTH-300                      |  Missing Features | Touch is not yet supported.
--- a/TABLETS.md
+++ b/TABLETS.md
@@ -97,6 +97,7 @@
 | VEIKK VK430                        |     Supported     |
 | VEIKK VK430 V2                     |     Supported     |
 | VEIKK VK640                        |     Supported     |
+| VEIKK VK850                        |     Supported     |
 | ViewSonic Woodpad PF0730           |     Supported     |
 | ViewSonic Woodpad PF1030           |     Supported     |
 | Wacom CTC-4110WL                   |     Supported     |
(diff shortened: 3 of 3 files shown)


Mrcubix:
Max Pressure is most likely 16383 ? not 16384.

Also, could you provide a data recording from the tablet debugger, purely for the wheel ?

We mostly need to confirm : 

- It's relative, like most other tablets from veikk,
- How many steps,
- Which value is report[2] ( 3rd byte ) when using the wheel,
- Value of report[4] & report[3] when using the wheel.

hadis-crystalis:
My apologies, I miscalculated the max pressure, it should be fixed now.

The 3rd byte is 0x03, 4th byte flickers to some value when i rotate the value and the 5th byte 0x01 when i rotate the wheel counterclockwise and 0x02 for clockwise rotation.

Mrcubix:
Ok last question, what's the value of the second byte when using the wheel ?

You might be able to just use `VeikkRelativeWheelReport` depending on `report[1]` & `report[2] == 0x03`.

hadis-crystalis:
The second byte is 0x42, same also when I use the aux buttons.

Mrcubix:
Can you try applying this ? 

Check if the wheel & aux works properly after that.

```patch
diff --git a/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs b/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
index d6eeaa78..04662902 100644
--- a/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
+++ b/OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
@@ -12,6 +12,7 @@ namespace OpenTabletDriver.Configurations.Parsers.Veikk
             {
                 0x41 when report[2] is 0xC0 => new OutOfRangeReport(report),
                 0x41 => new VeikkTiltTabletReport(report),
+                0x42 when report[2] is 0x03 => new VeikkRelativeWheelReport(report),
                 0x42 => new VeikkAuxReport(report),
                 0x43 => new DeviceReport(report),// touchpad report not supported yet
                 _ => new DeviceReport(report),
```

Also this

````patch
diff --git a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
index b73c35b1..68bc68e9 100644
--- a/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
+++ b/OpenTabletDriver.Configurations/Configurations/VEIKK/VK850.json
@@ -13,7 +13,13 @@
     },
     "AuxiliaryButtons": {
       "ButtonCount": 10
-    }
+    },
+    "Wheels": [
+      {
+        "RelativeWheelSteps": 12,
+        "ButtonCount": 0
+      }
+    ]
   },
   "DigitizerIdentifiers": [
     {
```

Mrcubix:
I'm unsure about the amount of steps, would be best if you could provide a data recording when spinning the wheel 360 degress in all directions.

- Tablet >Tablet Debugger
- File > (Enable) Data Recoyding.
- Spin the wheel 360 degress in both directions
- File > (Disable) Data Recoyding.
- Find the `tablet-data.txt` file in `~/.config/OpenTabletDriver`

hadis-crystalis:
Here is the data recording:
[tablet-data_1788103634.txt](https://github.com/user-attachments/files/31616038/tablet-data_1788103634.txt)

I cannot confirm they are precisely 360° but I estimate that 360° has ~60 recorded reports.

Mrcubix:
```json
"RelativeWheelSteps": 24,
```

instead of 12 should do the trick.

Mrcubix:
You can check if it's correct by setting a binding in the UX, set threshold to 360 and spin 360 degrees.

image

hadis-crystalis:
It is indeed correct, thank you for your contribution. I will add it into my pull request as soon as possible.

Mrcubix:
Just to make sure, aux still works properly right ?

hadis-crystalis:
Yes it does.

Mrcubix:
Don't forget to edit TABLETS.md to reflect that wheel is supported for this one.

hadis-crystalis:
Thank you for reminding me, I have done that now.

hadis-crystalis:
Albeit something weird I notice is that it doesn't work for wheel thresholds other than 15°.

Mrcubix:
Can you do another data recording now that the parser has wheel reports implemented ?

hadis-crystalis:
Here:
[tablet-data_1788106768.txt](https://github.com/user-attachments/files/31617627/tablet-data_1788106768.txt)

Mrcubix:
Can you drop one last diagnostic file After setting a value higher than 15 degrees ?

360 / 24 = 15.

Any values in increment of 15 should result in spinning extra steps to trigger the Binding.

hadis-crystalis:
Here:
[Diagnostics-067-6c71537a-dirty-dirty VEIKK VK850.json](https://github.com/user-attachments/files/31617712/Diagnostics-067-6c71537a-dirty-dirty.VEIKK.VK850.json)

Mrcubix:
So with a threshold of 30 degress, bindings don't get triggered ?

hadis-crystalis:
Yeah, they don't get triggered when the threshold is higher than 15°.

Mrcubix:
If you restart the daemon after setting the binding higher, does it work properly ? or does it still not work ?

Seems like this might be an issue with the relative wheel implementation, since absolute wheels seems to work as expected.

hadis-crystalis:
It still doesn't work after restarting the daemon.

Mrcubix:
Most definitely is an issue with multi-relative wheel implementation if accurate.

But then the issue isn't specific to this tablet, but all relative tablets.

Mrcubix:
You could try getting more logs in console or diag by applying this temporarily & spinning the wheel: 

```patch
diff --git a/OpenTabletDriver.Desktop/Binding/DeltaThresholdBindingState.cs b/OpenTabletDriver.Desktop/Binding/DeltaThresholdBindingState.cs
index 7c44bf33..27492aab 100644
--- a/OpenTabletDriver.Desktop/Binding/DeltaThresholdBindingState.cs
+++ b/OpenTabletDriver.Desktop/Binding/DeltaThresholdBindingState.cs
@@ -1,3 +1,4 @@
 using OpenTabletDriver.Plugin.Tablet;
 
 namespace OpenTabletDriver.Desktop.Binding
@@ -18,10 +19,14 @@ namespace OpenTabletDriver.Desktop.Binding
         {
             _accumulatedDelta += delta;
 
+            Log.Write("Device", $"Accumulated delta: {_accumulatedDelta}");
+
             while (IsNegativeThreshold ? _accumulatedDelta <= -ActivationThreshold : _accumulatedDelta >= ActivationThreshold)
             {
                 _accumulatedDelta -= IsNegativeThreshold ? -ActivationThreshold : ActivationThreshold;
 
+                Log.Write("Device", $"Threshold reached: {_accumulatedDelta}");
+
                 base.Invoke(tablet, report, true);
                 base.Invoke(tablet, report, false);
             }
```

hadis-crystalis:
Here's the debug recording with your patches:
[tablet-data_1788171888.txt](https://github.com/user-attachments/files/31640582/tablet-data_1788171888.txt)

Mrcubix:
The last patch i sent only throw logs in the console or diagnostic file.
I would need a diagnostic file after using the wheel bindings set to something higher than 15 instead.

hadis-crystalis:
My mistake, here's the new diagnostic file then:
[Diagnostics-067-b5911da1-dirty-dirty VEIKK VK850.json](https://github.com/user-attachments/files/31640753/Diagnostics-067-b5911da1-dirty-dirty.VEIKK.VK850.json)

Also in the logs when I rotate the wheel counterclockwise, the logs say:
```
[Device:Info]   Accumulated delta: -15
```
While clockwise, the logs say:
```
[Device:Info]   Accumulated delta: 15
```

Unlike when the bind threshold is set the 15 where there would be this after each line where the accumulated delta is ±15, a non 15 threshold doesn't have that.
```
[Device:Info]   Accumulated delta: 0
```

Mrcubix:
Couple of weird things :

- The threshold being reached is never logged.
- The accumulated delta never goes above 15 or under -15.

Mrcubix:
Ok this issue is for all tablets with relative wheels, the delta is being reset after each call to `HandleWheelData` which it's never supposed to do for obvious reasons.

Mrcubix:
You can remove the logging patch and use the wheel as is for now until relative wheels are fixed in another PR.

Mrcubix:
You could apply the fix from #5006  in your local build to see if that fixes the issue on your side.
( **Do not** apply the fix to this PR )

hadis-crystalis:
It does indeed fix the issues. Thank you for your contribution and your assistance debugging the wheel.

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
gonX, line 1: This change affects 5 other tablets:

`/opentabletdriver/OpenTabletDriver.Configurations/Configurations $ grep -R VeikkTiltReportParser | awk -F: '{ print $1 }' | sed 's!/! !;s/.json//' | uniq`
- VEIKK A15 Pro
- VEIKK VK1060
- VEIKK VK640
- VEIKK VK1200
- VEIKK VK2200PRO

/edit: sorry, meant to put this directly on the changed line instead of the file

gonX (commented):
Code LGTM but because this may unintentionally affect other tablets, approving will need extra care. @Kuuuube thoughts?

OpenTabletDriver.Configurations/Parsers/Veikk/VeikkTiltReportParser.cs
Mrcubix: Can you run the workflows so we can generate some build we can send for testing ?

Kuuuube: Done.

Issue #4942, opened 2026-07-15
"IndexOutOfRangeException" Bug


Ihnian:
### Steps to Reproduce

so after some time of usage otd randomly stop detecting tablet and show this message in console "[IndexOutOfRangeException:Error]	Index was outside the bounds of the array.
   at OpenTabletDriver.Configurations.Parsers.UCLogic.UCLogicTiltReportParser.Parse(Byte[] data) in /__w/OpenTabletDriver/OpenTabletDriver/build-RedHat/opentabletdriver/BUILD/opentabletdriver-0.6.7-build/opentabletdriver-0.6.7/OpenTabletDriver.Configurations/Parsers/UCLogic/UCLogicTiltReportParser.cs:line 11
   at OpenTabletDriver.Devices.DeviceReader`1.Main() in /__w/OpenTabletDriver/OpenTabletDriver/build-RedHat/opentabletdriver/BUILD/opentabletdriver-0.6.7-build/opentabletdriver-0.6.7/OpenTabletDriver/Devices/DeviceReader.cs:line 110" i dont know how to show this its just happen and tablet stop working its like i disable deamon 

### Expected Behavior

stop randomly stoping detecting tablet 

### Observed Behavior

it most often happend when i used buttons with binds and its most often ctrl z all things but now its just happen even 30 seconds after enabling otd deamon

### Tablet Device

huion inspiroy 2s

### Diagnostic Information

[Diagnostics-067.json](https://github.com/user-attachments/files/30065709/Diagnostics-067.json)

### Acknowledgements

- [x] I have installed OpenTabletDriver by following its [official installation instructions](https://opentabletdriver.net/).
- [x] I have checked the [OpenTabletDriver Wiki](https://opentabletdriver.net/Wiki) and respective FAQ pages, and my issue was not covered or actually fixed.
- [x] I have searched the existing issues and this new issue is not a duplicate of any.
- [x] I have written a concise and meaningful title.
- [x] I am on the latest version of OpenTabletDriver.

PR #5008, opened 2026-08-31
Core: Log bytes of throwing reports

gonX:
This should help debug instances of tablets sending sporadic weird reports that the parser throws an exception on.

Issues like #4942 are easier to debug with this.


Kuuuube (approved):
Looks good.

PR #5019, opened 2026-09-06
UX: Open load settings dialog in settings directory

TheAussie:
Changes the initial directory for File -> Load settings... from the user's Documents folder to the directory containing OpenTabletDriver's active settings.json.\n\nThis uses AppInfo.Current.SettingsFile rather than hardcoding a platform-specific configuration path, so custom app-data locations continue to work.\n\nTested on Linux with the GTK frontend and with a locally built package. The dialog now opens directly in the directory containing the active settings.json.

### DIFF
--- a/OpenTabletDriver.UX/MainForm.cs
+++ b/OpenTabletDriver.UX/MainForm.cs
@@ -511,7 +511,7 @@ namespace OpenTabletDriver.UX
         {
             var fileDialog = Extensions.OpenFileDialog(
                 "Load OpenTabletDriver settings...",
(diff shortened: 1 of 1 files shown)


gonX:
I'm not sure if I agree with how this PR is solving the aforementioned issue. Are you trying to reload the current settings file?

TheAussie:
> I'm not sure if I agree with how this PR is solving the aforementioned issue. Are you trying to reload the current settings file?

Essentially yeah, instead of clicking Load setting and having it open in the users document folder or home folder, it will open where the settings.json is located.

I don't know about you, but i got really annoyed having to navigate to the settings file all the time.

PR #5026, opened 2026-09-13
Add wheel support for UGEE M908

EinTyto:
### Summary

Adds wheel support for the UGEE M908 (VID: 28BD, PID: 2902) by introducing a dedicated report parser and updating the tablet configuration.

### Device Details

- **Tablet**: UGEE M908
- **VID/PID**: 28BD:2902
- **Wheel type**: Incremental rotary encoder, 24 steps per revolution
- **Wheel report format**: `02 F0 00 00 00 00 00 XX 00 00 00 00`
  - `XX = 0x01`: clockwise rotation
  - `XX = 0x02`: counterclockwise rotation
  - `XX = 0x00`: neutral / reset report

### Implementation

- Added `XP_PenM908WheelReport` implementing `IRelativeWheelReport`.
- Added `XP_PenM908ReportParser` that detects wheel reports before delegating to the existing XP-Pen parser logic.
- Updated `UGEE M908.json`:
  - Added `Wheels` declaration with `RelativeWheelSteps: 24` and `ButtonCount: 0`.
  - Pointed `ReportParser` to `OpenTabletDriver.Configurations.Parsers.XP_Pen.XP_PenM908ReportParser`.
- Updated `TABLETS.md` to mark the UGEE M908 as **Supported** (previously Missing Features).

### Testing

- Tested on Fedora 44 with a locally built OpenTabletDriver (0.6.x branch).
- Wheel rotation triggers bindings correctly in both directions.
- Auxiliary buttons remain functional while scrolling; the wheel reset report does not incorrectly release held buttons.
- Pen pressure, tilt, and all auxiliary buttons work as expected.

### Notes

The M908 sends two reports per wheel step: a rotation report followed by a reset report. The reset report (`02 F0 00 00 00 00 00 00 00 00 00 00`) is byte-identical to an "all auxiliary buttons released" report. The new parser tracks the previous report to avoid incorrectly releasing held buttons during wheel scrolling.

A note on naming: Since UGEE and XP-Pen are different product lines under the same company, I named the parser files `XP_PenM908`. Strictly speaking, they should probably be named `UgeeM908`. If the maintainers feel this naming is inappropriate, or if there are any other issues with the code, please feel free to let me know and I'll adjust accordingly.


Mrcubix:
Can you provide a data recording of the wheel being used in both directions ?
One can be obtainable from the Tablet Debugger's File Menu.

- Enable the option,
- Do the actions,
- Disable the option,
- Find the file generated in ~/.config/OpenTabletDriver and upload that here.

Mrcubix:
Also i'm about sure you can just re-use the existing XP_PenAuxReport, by providing an index for the wheel.

EinTyto:
> Can you provide a data recording of the wheel being used in both directions ? One can be obtainable from the Tablet Debugger's File Menu.
> 
>     * Enable the option,
> 
>     * Do the actions,
> 
>     * Disable the option,
> 
>     * Find the file generated in ~/.config/OpenTabletDriver and upload that here.

Sure, I've attached the data recording file(older version of Fedora repo). I just rotated the wheel one step clockwise, then one step counterclockwise.
[tablet-data_1789303677.txt](https://github.com/user-attachments/files/32160843/tablet-data_1789303677.txt)


> Also i'm about sure you can just re-use the existing XP_PenAuxReport, by providing an index for the wheel.

Regarding reusing XP_PenAuxReport, that was actually my initial thought as well. However, after scrolling the wheel one step, the device always sends a follow-up report 02 F0 00 00 00 00 00 00 00 00 00 00, which causes all currently pressed buttons to be released. That's why I wrote a dedicated parser. If we reuse XP_PenAuxReport, it may cause other buttons to be seen as released while scrolling the wheel. If that behavior is acceptable, then reusing XP_PenAuxReport is indeed a cleaner approach, and I can switch to that immediately if you'd like.

Mrcubix:
The behavior look simillar to the one found in the newer Veikk 850, after a delta, the tablet immediately sends another packet w/ 0 delta.
This one on the other hand, provides no way of knowing if the wheel is still used.
The issue with what you have currently, is that you have no way of guaranteeing that the next report will be the reset report.

There could be a pen report, an aux report or some other misc report that get thrown.
This means the code you currently have won't work for every cases, and there are still going to be some issue.

This one is gonna need a firmware update soon.

Now the question is, should we describe this tablet w/ Has Quirks, and this bug as the description ?

Mrcubix:
Here is a potential issue with the current code : What happens to the aux keys when a wheel report happen ?
As it is right now, it seems to kill any of their update.

EinTyto:
You are right, I will consider these cases. There will be some problems when some other reports show up between the two wheel reports. Cause some unexpected issue.
Honestly, before submitting this PR, I only tested this case: scroll the wheel while keep another auxbutton pressed. And it seems okay in my test.
Will update my code in several days after I have tested all these cases, with every case and its debug report uploaded in comment. Your comments are really helpful.

EinTyto:
I have tested some other cases. In some case,  the two wheel reports were seperated by another pen report. 
You can find it in [this record](https://github.com/user-attachments/files/32202691/tablet-data_1789359006.txt) at line 97.
```
Device:UGEE M908, { 02 F0 00 00 00 00 00 02 00 00 00 00 }, Delta:4.036ms, Wheel 1 Delta:-1, ReportType:XP_PenM908WheelReport
Device:UGEE M908, { 02 A0 68 67 DF 3E 00 00 11 00 00 00 }, Delta:2.026ms, Position:[26472,16095], Pressure:0, PenButtons:[False False], Eraser:False, Tilt:[17,0], ReportType:XP_PenTabletOverflowReport
Device:UGEE M908, { 02 F0 00 00 00 00 00 00 00 00 00 00 }, Delta:4.023ms, AuxButtons:[False False False False False False False False False False False False False False False False False False False False], Wheel 1 Delta:0, Wheel 2 Delta:0, ReportType:XP_PenAuxReport
```
Obviously, a pen report shows up between the wheel report pair. It resets all aux keys, forcing them to "release".

So I repaired the logic:
ParsingLogic

I'm testing this new logic and it seem okay with the case above.
I'm pretty sure this will solve the issue you have mentioned.

New parse file is done. Will push it later.

Mrcubix:
This already looks more sane to me.

One thing i wonder, is, if you could just store the aux keys state until the reset report happens.
Though i'm not sure if it will be worth testing.

Something around the lines of : 

1. You receive an aux report,
2. You save the aux state,
3. When a reset report occur, you replace the aux state with your savestate.

But anyway, if this code works on your side, i guess you may still get some feeback from the maintainers, and from other users that own this tablet.

Mrcubix:
I do also wonder if you could just use the normal Xp_PenReportParser & have a filter type plugin handle this logic to fix this quirk.

We have discussed the eventual need for plugins that fix such quirks anyway.

EinTyto:
I made a briefer version of the parsing logic, hope this will help to understand.
图片

It's a good idea using a filter plugin. I may look into that later.

In addition, I found that Ugee UE16/UE16 pro have the same buttons & wheel layout to M908.
图片图片


And it's also marked Missing Features | Wheel is not yet supported.
So I'm wondering if this parser could be reused by some other Ugee tablets/screens?
Need someone to test this screen.

One more thing, I just tested the official Ugee tablet driver v4.3.6 on Windows10.
When I hold a auxbutton (in my case, it is set to Ctrl, and wheel is set to "[" and "]" ) then scroll the wheel, the wheel will instantly reset all the auxbutton to release state. So I can not rotate the viewport in Krita.
If I want to rotate the viewport in Krita, I have to bind Ctrl to my pen button. Then hold pen button and rotate the wheel.
Which means pen button donot conflict with wheel.
The official driver team did not test/handle this auxbutton&wheel conflict at all !

EinTyto:
> I do also wonder if you could just use the normal Xp_PenReportParser & have a filter type plugin handle this logic to fix this quirk.

@Mrcubix 
I tried using the normal Xp_PenReportParser and it fuctions okay, just like the official driver.
They all have the same issue: **wheel reset report resets all auxbuttons**.
So it's okay we mark this tablet `Has Quirks | Wheel rotate resets auxiliary buttons`, and just add the M908.json.
This will keep the cleanest parser folder.

I will update the fork and pr later, but keep the custom parser version in my machine.
Anyway, thank you for your help. ^_^

PR #5032, opened 2026-09-16
MouseScrollBinding: Add bool property for inverting direction

### LINKED ISSUE #4876: Scroll direction in Mouse Scroll Bindings is unintuitive
The sign of `Amount` dictates which direction we're scrolling in, ie. a negative amount will scroll in the opposite direction. This is not intuitive for at least 2 users (myself being 1 of them) so I think we should look at alternative approaches to defining scroll direction.

Ideas to fix this could be to use a checkbox "Reverse direction", or a 2-state drop-down with up/down similar to how we have vertical/horizontal. Reverse direction is probably the better option of the 2, as it's likely more easier to program in (due to the fact that the dropdown method would have to be dynamic to accommodate scroll direction).

gonX:
Screenshot:
image

It's a bit difficult to make this look not weird given the current way boolean properties work in our GUI abstraction. Ideally I'd have no title (the left hand) and the right side would be "Invert Scroll Direction"

Fixes #4876


Kuuuube:
```
[BooleanProperty("Invert Scroll Direction", "")]
```
image

```
[BooleanProperty("", "Invert Scroll Direction")]
```
image

gonX:
Yeah I don't like the look of either of those. The 2nd makes the most sense but it looks naked without a title.

Kuuuube (approved):
Alright.

gonX:
We can always improve it later

PR #5033, opened 2026-09-16
CLI: Fix broken command aliases (`otd log` etc)

gonX:
Also makes them reappear in the `--help` menu

`git blame`'d to c4cd531 - marking as regression.


Kuuuube (approved):
I trust you.

gonX:
Yeah I'm not sure what's going on here either. Maybe a dependency update changed the behavior so that it was necessary at one point.

A Discord message from 2021 (which is prior to the linked offending commit) implies that `otd load` worked at some point: https://canary.discord.com/channels/615607687467761684/723399565780189234/925549107546849381

PR #5038, opened 2026-09-16
MacOS: Add Universal (Apple Silicon + Intel) app builds

xsyetopz:
# Changes

- Build and publish signed Universal macOS packages for Intel and Apple Silicon.
- Run Linux-specific HID tests only on Linux to avoid failing said tests on macOS & Windows.

Tested on macOS Golden Gate 27.0 (26A428)

EDIT: AI-assisted tooling was used in process of this contribution.


Kuuuube:
Is this AI (we dont strictly reject generated/assisted PRs but you must disclose it) or did you refactor parts of the build system just because you felt like it? A lot of this seems unnecessary here.

I also wonder what the goal of this PR is. We already have both intel and arm support. I'd rather not complicate the build system purely to save users 2 seconds reading the architecture of the file theyre downloading.

xsyetopz:
> Is this AI (we dont strictly reject generated/assisted PRs but you must disclose it) or did you refactor parts of the build system just because you felt like it? A lot of this seems unnecessary here.
> 
> I also wonder what the goal of this PR is. We already have both intel and arm support. I'd rather not complicate the build system purely to save users 2 seconds reading the architecture of the file theyre downloading.

Yes, it is. I had to because my macOS on 26.6.2, and then macOS 27.0 kept telling me I need Rosetta to run the application, AND that I should preferrably *install* it just to run OTD. It's been like that since late-December of 2025.

EDIT: I didn't know there was a rule to disclose it. I looked at the CONTRIBUTING.md but didn't find any guidelines for AI-assisted contributions. I'll keep this in mind for any future areas I go into that have no explicit guidelines.

EDIT2: Universal builds are supposed to be single app binary that run on both architectures, instead of creating two separate app binaries, so if it was published on the OTD website download section, Apple Silicon users wouldn't have to (like me) run it through Rosetta.

Kuuuube:
> Universal builds are supposed to be single app binary that run on both architectures, instead of creating two separate app binaries, so if it was published on the OTD website download section, Apple Silicon users wouldn't have to (like me) run it through Rosetta.

Indeed, but I question whether it's an actual issue to have two separate binaries. As for the rosetta issues in general, we plan on having arm builds posted for users on apple silicon sometime soon. It's worth noting that apple isnt discontinuing rosetta support until late next year.

xsyetopz:
> > Universal builds are supposed to be single app binary that run on both architectures, instead of creating two separate app binaries, so if it was published on the OTD website download section, Apple Silicon users wouldn't have to (like me) run it through Rosetta.
> 
> Indeed, but I question whether it's an actual issue to have two separate binaries. As for the rosetta issues in general, we plan on having arm builds posted for users on apple silicon sometime soon. It's worth noting that apple isnt discontinuing rosetta support until late next year.

[Universal builds](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) are supposed to be single downloadable file that works on both Intel & Apple Silicon MacOS users, so you're not exactly creating 2 separate binaries. Users on Apple Silicon ideally shouldn't depend on a translation layer to use OTD, because it's yet another additional download for no meaningful reason when the ability to create Universal binaries is right there.

I'd say it isn't much of a hassle to have a single, unified downloadable binary that doesn't require Rosetta for users on Apple Silicon.

gonX:
> [Universal builds](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) are supposed to be single downloadable file that works on both Intel & Apple Silicon MacOS users, so you're not exactly creating 2 separate binaries. Users on Apple Silicon ideally shouldn't depend on a translation layer to use OTD, because it's yet another additional download for no meaningful reason when the ability to create Universal binaries is right there.
> 
> I'd say it isn't much of a hassle to have a single, unified downloadable binary that doesn't require Rosetta for users on Apple Silicon.

I think what Kuuube is really asking is whether the 2 individual binaries we'll have for the upcoming v0.6.8 release (arm64 and intel) aren't adequate enough, and whether it even makes sense to bundle the binaries together

For macOS, we use single-file, and most importantly, self-contained builds (meaning the entire .NET runtime is bundled with the application), so the executable size is quite substantial.
Combining these into a single build (as your PR seemingly does) would inflate the download size for both users with no apparent gain for me. For the pending `arm64` deployment PR (#5039), the macOS downloads are already 75MB:
image
([link](https://github.com/OpenTabletDriver/OpenTabletDriver/actions/runs/35171180393))

If you enable Workflows on your fork you can see the resulting artifacts. I'd expect them to be over 100MB which I don't find acceptable. And again, the resulting convenience is from my perspective questionable.

Again I'd suggest checking up on #5039 and seeing whether that fixes it for you. arm64 builds has always been planned, but the 2 current maintainers (Kuuube and I) don't have direct access to a compatible arm64 device to test functionality on, so we've left it as a developer-only option until this PR came around.

AkiSakurai:
.NET only offers a two-year support cycle on macOS. There is a chance that Microsoft could drop universal build support from the toolchain.

xsyetopz:
> [...] whether the 2 individual binaries we'll have for the upcoming v0.6.8 release (arm64 and intel) aren't adequate enough, and whether it even makes sense to bundle the binaries together

Oh. Sorry, I haven't caught up to pre-releases.

> If you enable Workflows on your fork you can see the resulting artifacts. I'd expect them to be over 100MB which I don't find acceptable. And again, the resulting convenience is from my perspective questionable.

Wouldn't it then make sense to use a Workflow that creates a small, native binary through Native AOT, if that's even feasible here, which in theory should only bundle a subset of the .NET runtime, rather than the full thing?

> [...] but the 2 current maintainers (Kuuube and I) don't have direct access to a compatible arm64 device to test functionality on, so we've left it as a developer-only option until this PR came around.

I'm right here. I've got a compatible arm64 device. You could just tell me how I could test out your things the way you'd need.

Kuuuube:
> You could just tell me how I could test out your things the way you'd need.

Give it a go if you'd like: https://github.com/OpenTabletDriver/OpenTabletDriver/actions/runs/35261178326

Bottom of the page, grab the one ending in `_osx-arm64.tar.gz `. Nothing specific to test just whether it loads up fine and works in general.

xsyetopz:
> > You could just tell me how I could test out your things the way you'd need.
> 
> Give it a go if you'd like: https://github.com/OpenTabletDriver/OpenTabletDriver/actions/runs/35261178326
> 
> Bottom of the page, grab the one ending in `_osx-x64.tar.gz `. Nothing specific to test just whether it loads up fine and works in general.

image

gonX:
Damn, you actually asked about native AOT. I had almost 2 full paragraphs written out before I decided to not put it in my message.

The gist of it is that we can't do trimming and hence can't do native AOT.

xsyetopz:
> Damn, you actually asked about native AOT. I had almost 2 full paragraphs written out before I decided to not put it in my message.
> 
> The gist of it is that we can't do trimming and hence can't do native AOT.

So the options are...?

PR #5039, opened 2026-09-17
Add arm64 builds

Kuuuube:
This adds arm64 builds to CI for windows, macos, linux Simple, and linux BinaryTarBall.

MacOS builds tested and working, others untested (linux most likely works since we have users occasionally building manually for arm without issue).


.github/workflows/build-matrix.yml
```diff
@@ -32,11 +32,22 @@ jobs:
...
             runtime-prefix: linux-
           - build: macos
+            runtime-suffix: x64
             os: ubuntu-latest
             runtime-prefix: osx-
+          - build: macos
+            runtime-suffix: arm64
+            os: ubuntu-latest
+            runtime-prefix: osx-
+            signed: true
```
gonX, line 42: I feel like these new lines should be handled in the build script. If you're building for arm64 we should always sign there?

.github/workflows/build-matrix.yml
```diff
@@ -32,11 +32,22 @@ jobs:
...
           - build: windows
             os: windows-latest
             runtime-prefix: win-
+        exclude:
+          - build: Debian
+            runtime-suffix: arm64
+          - build: RedHat
+            runtime-suffix: arm64
```
gonX, line 50: I assume this is just because the build scripts are hardcoded to use x64? Did you try fixing that or was that more effort than worth?

.github/actions/build-wrapper/action.yml
```diff
@@ -27,9 +30,15 @@ runs:
...
+      if: inputs.runtime == 'osx-arm64' && inputs.signed == 'true'
       shell: bash
-      run: cargo install apple-codesign
+      env:
+        RCODESIGN_VERSION: 0.29.0
+      # Download binaries from github to avoid ~5 minute rust build times
+      # Alternative: `cargo install apple-codesign`
+      run: |
+        curl -L "https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign%2F${RCODESIGN_VERSION}/apple-codesign-${RCODESIGN_VERSION}-x86_64-unknown-linux-musl.tar.gz" | tar -xz -C /tmp
+        sudo install /tmp/apple-codesign-${RCODESIGN_VERSION}-x86_64-unknown-linux-musl/rcodesign /usr/local/bin/rcodesign
```
gonX: Fixing slow `cargo install` can theoretically be solved with the `cache` GitHub action if you're up for that task (most difficult part is engineering a good cache key), otherwise we'll just keep this as-is.

.github/actions/build-wrapper/action.yml
```diff
@@ -27,9 +30,15 @@ runs:
           git config --system --add safe.directory /__w/OpenTabletDriver/OpenTabletDriver
 
     - name: "Install rcodesign (macOS builds)"
-      if: inputs.package-name == 'macos-signed'
+      if: inputs.runtime == 'osx-arm64' && inputs.signed == 'true'
```
gonX, line 34: Is this uninteresting for osx-x64? If not I feel like this should just check for presence of `osx` instead

gonX (commented):
LGTM, just some questions

.github/actions/build-wrapper/action.yml
Kuuuube: There's no reason to adhoc sign x64. The only reason we're doing this for arm64 is because we're forced.

.github/workflows/build-matrix.yml
Kuuuube: There are a few bits that are hardcoded. I'm sure it's doable but I don't know exactly how debian and redhat's packaging works for arm.

.github/workflows/build-matrix.yml
Kuuuube: It is possible to run unsigned arm64 on macos but it's some developer stuff not regular app install iirc. At the moment it doesn't really matter besides giving a bit more flexibility if we want to change things later. If we ever needed signing for windows or linux builds in some way this also makes that trivial to include.

.github/actions/build-wrapper/action.yml
Kuuuube: I've added cache support. It hits cargo's API to check the latest version of rcodesign then uses that for the cache key. Ending up with something like this: `osx-arm64-rcodesign-0.29.0`.

gonX (approved):
LGTM

PR #5040, opened 2026-09-17
Fix warning on console format PluginSettingStore prefix

Kuuuube:
Annoying warning.

image

Shouldn't break anything.

PR #5042, opened 2026-09-17
CLI: Fix potential null assignment on non-nullable type

gonX:
It's currently throwing a warning on every build. "Regressed" with #4897 though it is debatable how often this field is actually null in normal use, so fairly low priority all things considered.


gonX:
Oh, #5040 :see_no_evil:

PR #5043, opened 2026-09-18
Add init sender tool

Kuuuube:
A fun and exciting new way for users to brick their tablets!

It can be quite useful for testing to be able to send arbitrary inits without having to screw around with the config constantly. There are also some restrictions on how OTD can send inits to devices in configs that make it unsuitable for things such as pulling XP-Pen tablets out of DFU mode whereas this initsender tool can do it no problem.

image


Kuuuube:
This should be good to go now. Tested and working on windows and linux.

I'm putting it on the 0.6.8 milestone but if it's too much to review, it can be dropped off.

